Issue #20, opened 2020-04-09
App freezes after taking a photo


isaiahtaylor:
I have not found this to be re-creatable 100% of the time on all devices. One of our iPhones it's hit or miss, another one it happens every time without fail.

Steps to recreate:

* press the + button for attachment
* press "photo from camera"
* take a photo

App is frozen. Force restart needed.

iPhone 11 Max is the one that always freezes. iPhone 7 Plus intermittent.

Issue #18, opened 2020-04-09
Keyboard opens on top of camera view


isaiahtaylor:
To recreate:

* tap in the text area of the message send box so that the keyboard comes up
* with the keyboard still up, tap the + button for an attachment
* select "photo from camera" or "video from camera"

The keyboard will come up over the camera view. There's no way to escape this without force quitting the app.

Maybe the + button needs to unfocus the text field?

imtoori:
Pretty sure It's related to this issue https://github.com/flutter/flutter/issues/42417
Trying to find a workaround

PR #24, opened 2020-04-10
Hotfix/attachments

imtoori:
- Fix video aspect ratio
- Add property to decide whether to enable video fullscreen
- Add property to hide the attachment button
- Do not show send button if an attachment is still uploading
- Unfocus and disable the TextField before opening the camera (workaround for https://github.com/flutter/flutter/issues/42417)
- Add gesture (vertical drag down) to close the keyboard
- Add keyboard type parameters (set it to `TextInputType.text` to show the submit button that will even close the keyboard)

The property showVideoFullScreen was added mainly because of this issue brianegan/chewie#261


imtoori:
It should close #20 
There was a bug opening the camera twice nad with the keyboard #18 bug it was really confusing

isaiahtaylor:
Thank you for such a rapid response on these items!

PR #79, opened 2020-09-09
Pass in SendMessageResponse to onMessageSent callback, instead of mos…

kriztoph:
…tly empty Message object.

When a message is sent, there's a callback called `onMessageSent`; right now it passes in the `Message` object before it's sent, so if we do something like modify the message in a presend hook, we have no way to get the message response and check for something like an error type. This change passes in the `Message` object that is returned from the API. 

# Submit a pull request

## CLA

### DIFF
--- a/lib/src/message_input.dart
+++ b/lib/src/message_input.dart
@@ -909,9 +909,9 @@ class MessageInputState extends State<MessageInput> {
           );
     }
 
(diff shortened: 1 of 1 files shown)


imtoori:
Hi @kriztoph 

thanks a lot for the pr! It will be available in the next release of the SDK 👍

PR #167, opened 2020-12-07
Make public autofocus field of the TextField of message_input

TramPamPam:
# Submit a pull request

## CLA

### DIFF
--- a/lib/src/message_input.dart
+++ b/lib/src/message_input.dart
@@ -95,6 +95,7 @@ class MessageInput extends StatefulWidget {
     this.attachmentThumbnailBuilders,
     this.inputTextStyle,
     this.attachmentIconColor,
+    this.autofocus,
   }) : super(key: key);
 
   /// Message to edit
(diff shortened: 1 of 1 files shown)


TramPamPam:
@imtoori @xsahil03x @tbarbugli 

Could someone review this pull request, please?

lib/src/message_input.dart
```diff
@@ -258,7 +262,7 @@ class MessageInputState extends State<MessageInput> {
             });
           },
           style: widget.inputTextStyle ?? Theme.of(context).textTheme.bodyText2,
-          autofocus: false,
+          autofocus: widget.autofocus ?? false,
```
xsahil03x, line 265: Can you please set the default value in the constructor itself. We do not really need to handle the null case as the textField itself throws an assertion error.

xsahil03x (changes requested):
@TramPamPam Just left one comment otherwise LGTM.

lib/src/message_input.dart
imtoori: I'll merge it and add this change

imtoori (approved):

lib/src/message_input.dart
TramPamPam: Sure thing. Changed in the last commit, please check.

PR #179, opened 2020-12-17
Feature/channel list view

imtoori:
sometimes users sort by custom property 
in that case it's good to not sort the channels on message.new events setting `lockChannelsOrder: true` and to listen to that event on your own
with this pr we expose the channels setter so that users can update it


xsahil03x (approved):
LGTM

xsahil03x (approved):

PR #182, opened 2020-12-22
feat: Added gridview for channel list view and changed implementation…

deven98:
… for image sharing

### DIFF
--- a/lib/src/channel_list_view.dart
+++ b/lib/src/channel_list_view.dart
@@ -278,21 +278,19 @@ class _ChannelListViewState extends State<ChannelListView>
                 },
               );
             } else {
-              child = ListView.custom(
+              child = ListView.separated(
                 physics: AlwaysScrollableScrollPhysics(),
-                controller: _scrollController,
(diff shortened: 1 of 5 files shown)


lib/src/channel_list_view.dart
```diff
@@ -259,22 +266,35 @@ class _ChannelListViewState extends State<ChannelListView>
...
+                  childCount: (channels.length * 2) + 1,
+                  findChildIndexCallback: (key) {
+                    final ValueKey<String> valueKey = key;
+                    final index = channels.indexWhere(
+                        (channel) => 'CHANNEL-${channel.id}' == valueKey.value);
+                    return index != -1 ? (index * 2) : null;
+                  },
+                ),
+              );
+            }
```
xsahil03x, line 297: You might want to change it to normal `ListView.builder` as it sometimes throws some index error. I did similar in `UserListView`

deven98: That code was already there, did not add that.

xsahil03x (approved):
LGTM

PR #186, opened 2020-12-28
Hotfix/channels bloc

imtoori:
# Submit a pull request

## CLA

The problem is that if we have the channel in memory we just add it, but it may happen that the channel is being used in another list.
Using `shouldAddChannel` the user can filter in the channels they want to add when a new message arrives.
This will probably be removed if we manage to create a filter system that is able to do such a thing.


deven98 (approved):

xsahil03x (approved):
LGTM

PR #204, opened 2021-01-08
feat: Added message flagging

deven98:
Added message flagging


lib/src/message_widget.dart
```diff
@@ -726,6 +731,7 @@ class _MessageWidgetState extends State<MessageWidget> {
                   widget.onReplyTap != null,
               showThreadReply:
                   widget.showThreadReplyIndicator && widget.onThreadTap != null,
+              showFlagButton: widget.showFlagButton && !isMyMessage,
```
imtoori, line 734: @deven98 why putting the condition `!isMyMessage` here if we put it in the messagelistview too?

imtoori (changes requested):

PR #224, opened 2021-01-20
Doc fixes 2.3

deven98:


packages/flutter_widgets/lib/src/giphy_attachment.dart
```diff
@@ -65,8 +67,9 @@ class GiphyAttachment extends StatelessWidget {
                   Padding(
                     padding: const EdgeInsets.all(8.0),
                     child: GestureDetector(
-                      onTap: () {
-                        Navigator.push(context, MaterialPageRoute(builder: (_) {
+                      onTap: () async {
```
Nash0x7E2, line 70: Consider extracting to a method. It helps make the build method more readable.

PR #230, opened 2021-01-22
[Flutter-SDK] Migrate to push enum

Nash0x7E2:
# Submit a pull request

## CLA

- Migration for #229

### DIFF
--- a/packages/flutter_widgets/example/lib/notifications_service.dart
+++ b/packages/flutter_widgets/example/lib/notifications_service.dart
@@ -56,7 +56,7 @@ void initNotifications(Client client) {
     if (connector.token.value != null) {
       client.addDevice(
         connector.token.value,
-        Platform.isAndroid ? 'firebase' : 'apn',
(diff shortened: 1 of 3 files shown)


packages/flutter_widgets/example/lib/notifications_service.dart
```diff
@@ -56,7 +56,7 @@ void initNotifications(Client client) {
     if (connector.token.value != null) {
       client.addDevice(
         connector.token.value,
-        Platform.isAndroid ? 'firebase' : 'apn',
+        Platform.isAndroid ? PushProvider.firebase : PushProvider.apn,
```
xsahil03x, line 59: I think converting this into a global getter will be better, something like `getPlatformPushProvider`

imtoori: just in the example or in the sdk?
because actually you might want to use firebase on ios too

xsahil03x: nvm then

PR #250, opened 2021-02-02
Hotfix/video

deven98:
Added video fixes


packages/stream_chat_flutter/lib/src/message_widget.dart
```diff
@@ -142,6 +142,8 @@ class MessageWidget extends StatefulWidget {
   /// Function called when quotedMessage is tapped
   final OnQuotedMessageTap onQuotedMessageTap;
 
+  final Map<String, VideoPackage> videoPackages;
```
imtoori, line 145: Let's add a comment here

packages/stream_chat_flutter/lib/src/message_list_view.dart
```diff
@@ -239,6 +239,8 @@ class _MessageListViewState extends State<MessageListView> {
 
   final MessageListController _messageListController = MessageListController();
 
+  Map<String, VideoPackage> videoPackages = {};
```
imtoori, line 242: This can be final

imtoori (changes requested):

packages/stream_chat_flutter/lib/src/message_list_view.dart
deven98: It's in the state, I don't think this matters that much.

packages/stream_chat_flutter/lib/src/message_widget.dart
deven98: Yep, adding it.

imtoori (approved):
LGTM

PR #251, opened 2021-02-02
Sample app

imtoori:
Moved sample app to separate repo
Created new sample app with tutorial steps


packages/stream_chat_flutter/example/pubspec.yaml
```diff
@@ -13,30 +27,53 @@ dependencies:
     path: ../
   stream_chat_persistence:
     path: ../../stream_chat_persistence
```
Nash0x7E2, line 29: These relative paths should be removed.

Nash0x7E2 (approved):
LGTM

packages/stream_chat_flutter/example/pubspec.yaml
imtoori: Are you sure?
I probably agree on the persistence package, but why should we remove the relative import of the sdk?

Nash0x7E2: Flutter is usually installed at `sdk: flutter`, not `path: ../`. A user may not have the framework installed in that directory.

imtoori: expand the diff :)

PR #257, opened 2021-02-04
update examples using latest llc version

imtoori:


packages/stream_chat_flutter/pubspec.yaml
```diff
@@ -11,7 +11,7 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
-  stream_chat_flutter_core: ^1.0.0-beta
+  stream_chat_flutter_core: ^1.0.2-beta
```
xsahil03x, line 14: This should be 1.0.1-beta

packages/stream_chat_flutter/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.0.1-beta
+
+- Update llc dependency
```
xsahil03x, line 3: Should be stream chat core instead of llc

xsahil03x (changes requested):
Some minor changes.

xsahil03x (approved):
LGTM

PR #282, opened 2021-02-25
Fix/send anim duration

deven98:
Added message input theme and send animation duration

### DIFF
--- a/packages/stream_chat_flutter/lib/src/message_actions_modal.dart
+++ b/packages/stream_chat_flutter/lib/src/message_actions_modal.dart
@@ -596,7 +596,8 @@ class _MessageActionsModalState extends State<MessageActionsModal> {
       elevation: 2,
       clipBehavior: Clip.hardEdge,
       isScrollControlled: true,
-      backgroundColor: StreamChatTheme.of(context).channelTheme.inputBackground,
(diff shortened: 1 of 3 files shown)


imtoori (changes requested):
LGTM
but I would move some properties that are not in the `MessageInputTheme`, but are related to the `MessageInput`
for example: 
- sendButtonColor
- actionButtonColor
- sendButtonIdleColor
- actionButtonIdleColor
- inputBackground

that are in the `ChannelTheme` as of now

xsahil03x:
Just an opinion but I don't feel `sendAnimationDuration` is something that should be part of the theme. An extra field in the `MessageInput` seems better to me as a `User`.

imtoori:
I get it, but the issue is that it won't be applied to the messageinput while in edit mode

deven98:
Don't we use the message input internally in edit anyway?

xsahil03x:
Yes

imtoori:
yeah that's what I'm saying
if we put it as a param of MessageInput the user will need to customize it in two places, if we put in the theme it will be easier in my opinion
what do you think?

imtoori (approved):
LGTM

PR #283, opened 2021-02-25
fix: channel header text style

deven98:
Fixed channel title

### DIFF
--- a/packages/stream_chat_flutter/lib/src/image_actions_modal.dart
+++ b/packages/stream_chat_flutter/lib/src/image_actions_modal.dart
@@ -103,11 +103,11 @@ class ImageActionsModal extends StatelessWidget {
                         color: StreamChatTheme.of(context).colorTheme.accentRed,
                       ),
                       () {
-                        var channel = StreamChannel.of(context).channel;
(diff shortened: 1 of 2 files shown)


packages/stream_chat_flutter/lib/src/image_actions_modal.dart
```diff
@@ -103,11 +103,11 @@ class ImageActionsModal extends StatelessWidget {
                         color: StreamChatTheme.of(context).colorTheme.accentRed,
                       ),
                       () {
+                        var channel = StreamChannel.of(context).channel;
+                        channel.updateMessage(
+                            message..attachments.removeAt(currentIndex));
```
imtoori, line 108: This should not be in this pr, right?

imtoori (changes requested):
LGTM, but we need to remove the image action thing
it's in another pr

packages/stream_chat_flutter/lib/src/image_actions_modal.dart
deven98: Think I branched from the wrong thing, fixed.

PR #285, opened 2021-02-25
add platform type to userAgent header

imtoori:


packages/stream_chat/lib/src/platform_detector/platform_detector.dart
```diff
@@ -0,0 +1,47 @@
...
+      case PlatformType.Linux:
+        return 'linux';
+      case PlatformType.Fuchsia:
+        return 'fuchsia';
+      default:
+        return '';
+    }
+  }
+
+  PlatformType get type => currentPlatform;
```
xsahil03x, line 46: I think we should just make this getter static too.

xsahil03x (commented):
LGTM

ruggi:
👏

PR #294, opened 2021-03-01
add pana check gh action

imtoori:


Nash0x7E2 (approved):
Looks good 👍  Do you think it's worth displaying the passing pub score? I notice we set the passing score to 100/110.

imtoori:
It's displayed actually, there is also a report of the result 
check this out https://github.com/GetStream/stream-chat-flutter/pull/294/checks?check_run_id=2005377256

the max is just 100 btw, we set it to the maximum

PR #310, opened 2021-03-03
fix: generate image thumbnails and use transparent image as placehold…

imtoori:
…er for cached network images

### DIFF
--- a/packages/stream_chat_flutter/lib/src/attachment/giphy_attachment.dart
+++ b/packages/stream_chat_flutter/lib/src/attachment/giphy_attachment.dart
@@ -1,6 +1,9 @@
+import 'dart:typed_data';
+
 import 'package:cached_network_image/cached_network_image.dart';
 import 'package:flutter/material.dart';
 import 'package:stream_chat_flutter_core/stream_chat_flutter_core.dart';
+import 'package:transparent_image/transparent_image.dart';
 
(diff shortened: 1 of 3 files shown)


packages/stream_chat_flutter/lib/src/attachment/giphy_attachment.dart
```diff
@@ -1,6 +1,9 @@
+import 'dart:typed_data';
+
 import 'package:cached_network_image/cached_network_image.dart';
 import 'package:flutter/material.dart';
 import 'package:stream_chat_flutter_core/stream_chat_flutter_core.dart';
+import 'package:transparent_image/transparent_image.dart';
 
 import '../full_screen_media.dart';
```
Nash0x7E2, line 8: May want to change these to package imports, one less thing to change in our lints 😉

Nash0x7E2 (approved):
LGTM

packages/stream_chat_flutter/lib/src/attachment/giphy_attachment.dart
imtoori: I don't think we should do this 
I'm the first one who would do it because I like spaghetti and everything similar (spaghetti code, spaghetti repo and so on...), but generally speaking I don't think it's a good practice

PR #317, opened 2021-03-06
fix: Method being called during build

Nash0x7E2:
# Submit a pull request

## CLA

- [x] Added/Updated documentation where applicable 

### Bug 
As reported:  "When we use ChannelListView and we override the onImageTap trying to push our own view, it throws an error. `onImageTap` is called when the widget is building."

### Fix 
Add lambda which calls `widget.onImageTap` when pressed.


imtoori (changes requested):
can you fix analysis errors?

Nash0x7E2:
@imtoori I don't think the analysis errors are directly related to this PR. Maybe it's worth opening another PR for that?

xsahil03x:
I agree, let's merge this and create another for the analysis.

imtoori:
yes I changed the branch to develop and I'll merge it
we can fix the analysis later

PR #320, opened 2021-03-08
fix: remove notification badge logic from the sdk and move it to samp…

imtoori:
Remove the logic from the SDK and move it to the sample app

Why?
- we remove a dependency
- the package was asking the notification permissions right away and that's not always what our customers want


xsahil03x:
Where do we exactly show this in UI?

imtoori:
@xsahil03x 
we don't 
it's related to the ios app badge

xsahil03x (approved):
LGTM

PR #323, opened 2021-03-09
feat: Changed attachment builder structure

deven98:
Custom attachment support


packages/stream_chat_flutter/lib/src/message_widget.dart
```diff
@@ -837,6 +840,22 @@ class _MessageWidgetState extends State<MessageWidget>
         [];
 
     if (images.length > 1) {
```
xsahil03x, line 842: Now that we are passing `List<Attachment>` instead of a single `Attachment` to the `AttachmentWidget`, I would suggest we move the attachment `grouping logic` to their respective widget instead of handling it in `MessageWidget`.

PR #340, opened 2021-03-17
fix: unread count not updating while the chat is open

### LINKED ISSUE #338: Unreadcount not updating
**Describe the bug**
client.state.**totalUnreadCountStream** and channel.state.**unreadCount** are not counting correctly while the chat is opened

**Version of the package**
tested with 1.4.0-beta and 1.3.0-beta

**To Reproduce**
1. Enter in a channel page and while you are with the chat open, ask the other user to send you a few messages
2. Back to the channels list you will notice that all messages received while you were on the channel are still being considered as unread on client.state.**totalUnreadCountStream** and channel.state.**unreadCount**
3. You can test it with example 6: https://github.com/GetStream/stream-chat-flutter/blob/master/packages/stream_chat_flutter/example/lib/tutorial-part-6.dart

**Expected behavior**
If you are with the chat opened in the moment that you received the message, it means that you have read them, so they shouldn't be added to the unreadCount

gz#10259

imtoori:
fixes #338

### DIFF
--- a/packages/stream_chat_flutter/pubspec.yaml
+++ b/packages/stream_chat_flutter/pubspec.yaml
@@ -1,7 +1,7 @@
 name: stream_chat_flutter
 homepage: https://github.com/GetStream/stream-chat-flutter
 description: Stream Chat official Flutter SDK. Build your own chat experience using Dart and Flutter.
-version: 1.5.0
+version: 1.5.1
 repository: https://github.com/GetStream/stream-chat-flutter
 issue_tracker: https://github.com/GetStream/stream-chat-flutter/issues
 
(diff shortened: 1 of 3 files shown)


xsahil03x (approved):
LGTM

packages/stream_chat_flutter/pubspec.yaml
```diff
@@ -1,7 +1,7 @@
 name: stream_chat_flutter
 homepage: https://github.com/GetStream/stream-chat-flutter
 description: Stream Chat official Flutter SDK. Build your own chat experience using Dart and Flutter.
-version: 1.5.0
+version: 1.5.1
```
xsahil03x, line 4: I think v1.5.0+1 would be better as it's a hotfix

xsahil03x: https://dart.dev/tools/pub/versioning

PR #345, opened 2021-03-19
fix: ws disconnection

imtoori:
- removed conditional imports for web
- fix disconnection while connecting
- add test case


Nash0x7E2 (approved):
LGTM

PR #355, opened 2021-03-25
fix: Fixed sendingbutton.outside case

deven98:
Fixed sending button padding issue

### DIFF
--- a/packages/stream_chat_flutter/lib/src/message_input.dart
+++ b/packages/stream_chat_flutter/lib/src/message_input.dart
@@ -491,16 +491,17 @@ class MessageInputState extends State<MessageInput> {
 
   Expanded _buildTextInput(BuildContext context) {
     final theme = StreamChatTheme.of(context);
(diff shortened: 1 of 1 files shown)


packages/stream_chat_flutter/lib/src/message_input.dart
```diff
@@ -495,6 +495,12 @@ class MessageInputState extends State<MessageInput> {
       child: Center(
         child: Container(
           clipBehavior: Clip.antiAlias,
+          margin: (widget.sendButtonLocation == SendButtonLocation.inside
```
Nash0x7E2, line 498: nitpick: Consider extracting logic to a variable at the top of our build function. It's a nice way to keep the build method clean 😉

Nash0x7E2 (approved):
Minor nitpick but looks good 👍 . LGTM

imtoori (approved):
LGTM

PR #361, opened 2021-03-30
refactor(ui-kit): update giphy interaction

xsahil03x:


imtoori (approved):
text overflow wasn't handled
I added it and now LGTM

PR #367, opened 2021-03-31
feat: exposed textbuilder callback

deven98:


imtoori (changes requested):
you should pass it in the parent message as well

imtoori (commented):
LGTM

PR #377, opened 2021-04-02
fix: attachments upload when size exceeds

### LINKED ISSUE #371: Files Larger Than 20MB Can Still Be Sent and Become Permanently Stuck
**Describe the bug**
Attempting to attach a file larger than 20MB shows the error but allows the file to be sent.
Upload then subsequently fails and the failed file becomes stuck permanently, with no ability to remove.

**What package are you using? What version?**
v1.5.1 

**What platform is it about?**
- [X] Android
- [ ] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS
- [ ] Linux

**To Reproduce**
1. Attach video larger than 20MB.
2. Dismiss warning message.
3. File attachment still exists and can be sent.
4. Send
5. Sending fails and becomes stuck in chat with no ability to remove.

**Expected behavior**
Do not allow attachments that are larger than the file size limit.

**Screenshots**
Screenshot_20210331-132550
Screenshot_20210331-132606

deven98:
Fixes #371

### DIFF
--- a/packages/stream_chat_flutter/lib/src/message_input.dart
+++ b/packages/stream_chat_flutter/lib/src/message_input.dart
@@ -1989,8 +1989,6 @@ class MessageInputState extends State<MessageInput> {
       extraData: extraDataMap.isNotEmpty ? extraDataMap : null,
     );
 
-    _attachments[attachment.id] = attachment;
-
     if (file.size > _kMaxAttachmentSize) {
       if (attachmentType == 'Video') {
         final mediaInfo = await VideoService.compressVideo(file.path);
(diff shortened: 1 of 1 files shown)


packages/stream_chat_flutter/lib/src/message_input.dart
```diff
@@ -1988,8 +1989,6 @@ class MessageInputState extends State<MessageInput> {
       extraData: extraDataMap.isNotEmpty ? extraDataMap : null,
     );
 
-    _attachments[attachment.id] = attachment;
```
xsahil03x, line 1991: I think we added the attachment at the start because the compression may take some time and we want to show it ASAP as the selected item in message input.

deven98: I think there's a a better way to do that. If we want to do this, let's also add a check at the end checking for file sizes.

xsahil03x (approved):
LGTM

PR #381, opened 2021-04-02
feat: migrate to NNBD(non-nullable by default)

xsahil03x:
- feat: migrate to NNBD(non-nullable by default)

### Plan

**Phase 1. Upgrade dependencies to null-safe compatible dependencies for each stream library**
  - [X] stream_chat
  - [x] stream_chat_flutter
  - [x] stream_chat_flutter_core
  - [x] stream_chat_persistence

**Phase 2. Upgrade to sound null safety for each stream library**
  - [x] stream_chat
  - [x] stream_chat_flutter
  - [x] stream_chat_flutter_core
  - [x] stream_chat_persistence


packages/stream_chat/lib/src/api/websocket.dart
```diff
@@ -107,43 +107,43 @@ class WebSocket {
...
   /// The current connection status value
-  ConnectionStatus get connectionStatus => _connectionStatusController.value;
+  ConnectionStatus? get connectionStatus => _connectionStatusController.value;
 
   /// This notifies of connection status changes
   Stream<ConnectionStatus> get connectionStatusStream =>
       _connectionStatusController.stream;
 
-  String _path;
+  late final String _path;
```
ryanheise, line 116: I'm observing:

```
LateInitializationError: Field '_path@2300330865' has already been initialized.
```

because `_path` is updated multiple times elsewhere in this class.

imtoori: Thanks! That was not right, I just updated it

PR #390, opened 2021-04-09
feat(ui): [CDS-239] NNBD 

imtoori:


packages/stream_chat/lib/src/api/channel.dart
```diff
@@ -193,9 +193,9 @@ class Channel {
       state?._channelState.channel?.extraData ?? _extraData;
 
   /// Channel extra data as a stream
-  Stream<Map<String, dynamic>?>? get extraDataStream {
+  Stream<Map<String, dynamic>>? get extraDataStream {
     _checkInitialized();
-    return state?.channelStateStream.map((cs) => cs.channel?.extraData);
+    return state?.channelStateStream.map((cs) => cs.channel?.extraData ?? {});
```
xsahil03x, line 198:
```suggestion
    return state?.channelStateStream.map((cs) => cs.channel?.extraData ?? _extraData);
```

deven98:
Mentions are highlighted initially, after which they go back to black 

image

deven98:
Might not be a null safety thing, but padding for MessageInput is off 

image

packages/stream_chat_flutter/lib/src/message_list_view.dart
```diff
@@ -1008,19 +1008,20 @@ class _MessageListViewState extends State<MessageListView> {
             break;
           case ReturnActionType.reply:
             FocusScope.of(context).unfocus();
-            widget.onMessageSwiped(message);
+            widget.onMessageSwiped!(message);
```
deven98, line 1011: We should probably have a null check for widget.onMessageSwiped instead of a bang here

deven98:
image

Saving doesn't work. If it isn't supported in the example, we should show a better error message of what went wrong or remove the option.

deven98:
May be a non-issue:

Are two different types supposed to be cojoined?

image

packages/stream_chat_flutter/lib/src/deleted_message.dart
```diff
@@ -5,25 +5,25 @@ import 'package:stream_chat_flutter/src/stream_chat_theme.dart';
...
+    required this.messageTheme,
     this.borderRadiusGeometry,
     this.shape,
     this.borderSide,
     this.reverse = false,
   }) : super(key: key);
 
   /// The theme of the message
-  final MessageTheme messageTheme;
+  final MessageTheme? messageTheme;
```
deven98, line 17: MessageTheme is required here, can we make it non-nullable?

packages/stream_chat_flutter/lib/src/image_header.dart
```diff
@@ -32,9 +32,9 @@ class ImageHeader extends StatelessWidget implements PreferredSizeWidget {
 
   /// Creates a channel header
   ImageHeader({
-    Key key,
-    this.message,
-    this.urls,
+    Key? key,
+    required this.message,
+    this.urls = const [],
```
deven98, line 37: Are we using the urls anywhere?

deven98: Can we add a type to the currentIndex too?

packages/stream_chat_flutter/lib/src/stream_chat.dart
```diff
@@ -32,21 +32,21 @@ import 'dart:ui' as ui;
 /// Use [StreamChat.of] to get the current [StreamChatState] instance.
 class StreamChat extends StatefulWidget {
   final StreamChatClient client;
-  final Widget child;
-  final StreamChatThemeData streamChatThemeData;
+  final Widget? child;
```
deven98, line 35: child is required, can make this non nullable

packages/stream_chat_flutter/lib/src/thread_header.dart
```diff
@@ -61,30 +61,30 @@ class ThreadHeader extends StatelessWidget implements PreferredSizeWidget {
...
-  final Widget subtitle;
+  final Widget? subtitle;
 
   /// Leading widget
-  final Widget leading;
+  final Widget? leading;
 
   /// AppBar actions
-  final List<Widget> actions;
+  final List<Widget>? actions;
```
deven98, line 82: Probably can make this non nullable and a const [] by default

packages/stream_chat_flutter/lib/src/upload_progress_indicator.dart
```diff
@@ -5,20 +5,27 @@ import 'stream_chat_theme.dart';
...
-    Key key,
-    @required this.uploaded,
-    @required this.total,
-    this.progressIndicatorColor = const Color(0xffb2b2b2),
-    this.padding = const EdgeInsets.only(top: 5, bottom: 5, right: 11, left: 5),
+  UploadProgressIndicator({
+    Key? key,
+    required this.uploaded,
+    required this.total,
+    Color? progressIndicatorColor,
```
deven98, line 17: Unsure of why this can't be    

`this.progressIndicatorColor = const Color(0xffb2b2b2),`

packages/stream_chat_flutter/lib/src/url_attachment.dart
```diff
@@ -9,17 +9,17 @@ class UrlAttachment extends StatelessWidget {
   final EdgeInsets textPadding;
 
   UrlAttachment({
-    @required this.urlAttachment,
-    @required this.hostDisplayName,
-    @required this.textPadding,
+    required this.urlAttachment,
+    required this.hostDisplayName,
+    required this.textPadding,
```
deven98, line 14: Can we give this a default value? Not sure why this is required before or after

packages/stream_chat_flutter/lib/src/utils.dart
```diff
@@ -46,17 +46,17 @@ Future<bool> showConfirmationDialog(
               if (icon != null) icon,
               SizedBox(height: 26.0),
               Text(
-                title,
+                title!,
```
deven98, line 49: Add default text or make NN

packages/stream_chat_flutter/lib/src/utils.dart
```diff
@@ -46,17 +46,17 @@ Future<bool> showConfirmationDialog(
...
                 style: StreamChatTheme.of(context).textTheme.headlineBold,
               ),
               SizedBox(height: 7.0),
               Text(
-                question,
+                question!,
```
deven98, line 54: Add default text or make NN

packages/stream_chat_flutter/lib/src/utils.dart
```diff
@@ -69,7 +69,7 @@ Future<bool> showConfirmationDialog(
                           Navigator.of(context).pop(false);
                         },
                         child: Text(
-                          cancelText,
+                          cancelText!,
```
deven98, line 72: Add default text or make NN

packages/stream_chat_flutter/lib/src/utils.dart
```diff
@@ -90,7 +90,7 @@ Future<bool> showConfirmationDialog(
                           Navigator.pop(context, true);
                         },
                         child: Text(
-                          okText,
+                          okText!,
```
deven98, line 93: Add default text or make NN

packages/stream_chat_flutter/lib/src/utils.dart
```diff
@@ -110,17 +110,17 @@ Future<bool> showConfirmationDialog(
...
   BuildContext context, {
-  String title,
-  Widget icon,
-  String details,
-  String okText,
-  StreamChatThemeData theme,
+  String? title,
+  Widget? icon,
+  String? details,
+  String? okText,
```
deven98, line 118: Same as before, text is used with bangs and no default value

packages/stream_chat_flutter/lib/src/video_thumbnail_image.dart
```diff
@@ -9,17 +9,17 @@ import 'stream_svg_icon.dart';
 import 'video_service.dart';
 
 class VideoThumbnailImage extends StatefulWidget {
-  final String video;
-  final double width;
-  final double height;
-  final BoxFit fit;
+  final String? video;
```
deven98, line 12: Required but nullable

packages/stream_chat_flutter/lib/src/message_text.dart
```diff
@@ -7,43 +8,45 @@ import 'utils.dart';
...
-  final void Function(String) onLinkTap;
-  final MessageTheme messageTheme;
+  final void Function(User)? onMentionTap;
+  final void Function(String)? onLinkTap;
+  final MessageTheme? messageTheme;
 
   @override
   Widget build(BuildContext context) {
-    final text = _replaceMentions(message.text).replaceAll('\n', '\\\n');
+    final text = _replaceMentions(message.text)!.replaceAll('\n', '\\\n');
```
deven98, line 25: Consider making the _replaceMentions method non nullable and accept a NN string, pass a default value (message.text ?? '') when calling it

packages/stream_chat_flutter/lib/src/thread_header.dart
imtoori: I wouldn't do this, if you look at the AppBar widget they don't do that

packages/stream_chat_flutter/lib/src/stream_chat.dart
imtoori: I wouldn't do it
instead I would use a default empty widget 

problem is we cant to be able to use it in the MaterialApp builder property and the passed widget is nullable
we can talk about it on a call, let me know

deven98:
LGTM

PR #406, opened 2021-04-26
[CDS-191] feat(llc): add support for type safe filters

xsahil03x:

### DIFF
--- a/packages/stream_chat/lib/src/models/filter.dart
+++ b/packages/stream_chat/lib/src/models/filter.dart
@@ -1,0 +1,197 @@
+// ignore_for_file: non_constant_identifier_names, constant_identifier_names
+
+const _groupOperators = [
+  FilterOperator.and,
+  FilterOperator.or,
+  FilterOperator.nor,
+];
+
+/// Possible operators to use in filters.
+enum FilterOperator {
+  /// Matches values that are equal to a specified value.
+  equal,
+
+  /// Matches all values that are not equal to a specified value.
+  notEqual,
+
+  /// Matches values that are greater than a specified value.
+  greater,
+
+  /// Matches values that are greater than a specified value.
+  greaterOrEqual,
+
+  /// Matches values that are less than a specified value.
+  less,
+
+  /// Matches values that are less than or equal to a specified value.
+  lessOrEqual,
+
+  /// Matches any of the values specified in an array.
+  in_,
+
+  /// Matches none of the values specified in an array.
+  notIn,
+
+  /// Matches values by performing text search with the specified value.
+  query,
+
+  /// Matches values with the specified prefix.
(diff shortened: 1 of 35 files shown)


packages/stream_chat/lib/src/models/filter.dart
```diff
@@ -0,0 +1,197 @@
+// ignore_for_file: non_constant_identifier_names, constant_identifier_names
```
Nash0x7E2, line 1: 👀

packages/stream_chat/lib/src/models/filter.dart
```diff
@@ -0,0 +1,197 @@
...
+        FilterOperator.nor: '\$nor',
+      }[this]!;
+}
+
+/// Stream supports a limited set of filters for querying channels,
+/// users and members. The example below shows how to filter for channels
+/// of type messaging where the current user is a member
+///
+/// ```dart
+/// val filter = Filter.and(
```
Nash0x7E2, line 80: `val` is not valid Dart syntax :))

Nash0x7E2 (commented):
Looks good thus far 👍

packages/stream_chat/lib/src/models/filter.dart
```diff
@@ -0,0 +1,197 @@
...
+///     Filter.equal('type', 'messaging'),
+///     Filter.in_('members', [user.id])
+/// )
+/// ```
+/// See <a href="https://getstream.io/chat/docs/query_channels/?language=dart" target="_top">Query Channels Documentation</a>
+class Filter {
+  const Filter.__({
```
imtoori, line 87: what do you think about giving a name to these internal constructors?

packages/stream_chat/lib/src/models/filter.dart
xsahil03x:
```suggestion
/// final filter = Filter.and(
```

packages/stream_chat/lib/src/models/filter.dart
xsahil03x: @Nash0x7E2 It's to suppress lint because of the underscore after `in`, I added the `_` because we do not have any support for [escaping reserved keyword](https://github.com/dart-lang/language/issues/271) in dart.
```dart
  /// Matches any of the values specified in an array.
  in_,
```

packages/stream_chat/lib/src/models/filter.dart
xsahil03x: Any particular use-case for that? I don't really think we need to name them.

imtoori: we don't actually need to, it was just for readability

this-is-allan:
oh nice

PR #407, opened 2021-04-28
fix: pinMessage and add docs

deven98:
Fixed pinMessage functionality and add docs for it

### DIFF
--- a/packages/stream_chat/lib/src/client.dart
+++ b/packages/stream_chat/lib/src/client.dart
@@ -1333,6 +1333,8 @@ class StreamChatClient {
   }
 
   /// Pins provided message
+  /// [timeoutOrExpirationDate] can either be a [DateTime] or a value in seconds
+  /// to be added to [DateTime.now]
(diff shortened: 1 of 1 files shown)


packages/stream_chat/lib/src/client.dart
```diff
@@ -1350,9 +1352,11 @@ class StreamChatClient {
...
       pinExpires = timeoutOrExpirationDate.toUtc();
     } else if (timeoutOrExpirationDate is num) {
-      pinExpires = DateTime.now().add(
-        Duration(seconds: timeoutOrExpirationDate.toInt()),
-      );
+      pinExpires = DateTime.now()
+          .add(
+            Duration(seconds: timeoutOrExpirationDate.toInt()),
+          )
+          .toUtc();
```
xsahil03x, line 1359: This wasn't necessary btw, as we already convert to UTC in Message constructor.

deven98: In that case the top one would be redundant too, right?

PR #421, opened 2021-05-06
fix: [CDS-262] move .of calls to didChangeDependencies

imtoori:


packages/stream_chat_flutter_core/lib/src/message_list_core.dart
```diff
@@ -167,21 +167,23 @@ class MessageListCoreState extends State<MessageListCore> {
...
-    _streamChannel = StreamChannel.of(context);
-    if (!_initialized && _isThreadConversation) {
-      _initialized = true;
-      _streamChannel.getReplies(widget.parentMessage!.id);
+    final newStreamChannel = StreamChannel.of(context);
+
+    if (newStreamChannel != _streamChannel) {
+      _streamChannel = newStreamChannel;
+      if (_isThreadConversation) {
+        _streamChannel!.getReplies(widget.parentMessage!.id);
```
deven98, line 183: Do we want to do this in didUpdateWidget and once somewhere else while starting? Logic is the parent message ID is no longer valid for a new stream channel in the inherited widget, and they would need to change the parent message ID too, which would better be suited in didUpdate than here.

deven98 (approved):
LGTM

PR #431, opened 2021-05-14
fix: [CDS-299] mentioned users

imtoori:


packages/stream_chat/lib/src/models/reaction.dart
```diff
@@ -50,7 +50,7 @@ class Reaction {
 
   /// Reaction custom extraData
   @JsonKey(includeIfNull: false)
-  final Map<String, Object>? extraData;
+  final Map<String, Object?>? extraData;
```
xsahil03x, line 53: should we make this non-nullable too?

d3xvn (approved):
LGTM

PR #433, opened 2021-05-17
fix: filter.operator can be null

imtoori:
the filter will be encoded like a group operator would


xsahil03x:
@imtoori can you also add some test for this change

packages/stream_chat/lib/src/models/filter.dart
```diff
@@ -164,6 +166,11 @@ class Filter extends Equatable {
     String? key,
   }) = Filter.__;
 
+  /// Creates a custom [Filter] from a raw value
```
xsahil03x, line 169: maybe also add an example in the doc

xsahil03x (approved):
LGTM

PR #448, opened 2021-05-21
fix: make timeoutOrExpirationDate optional

### LINKED ISSUE #447: Allow null timeout in pinMessage for unbounded expiration
**Is your feature request related to a problem? Please describe.**

The pre-null-safety version used to allow a `null` timeout to effectively make it unbounded.

In the null safety preview this parameter was made non-nullable so you can't indicate `null` anymore.

**Describe the solution you'd like**

Although the parameter is now non-nullable, the internal logic in that method is still the same and has code to handle the `null` case:

```dart
  Future<UpdateMessageResponse> pinMessage(
    Message message,
    Object timeoutOrExpirationDate,
  ) {
    assert(() {
      if (timeoutOrExpirationDate is! DateTime &&
          timeoutOrExpirationDate is! num) {
        throw ArgumentError('Invalid timeout or Expiration date');
      }
      return true;
    }(), 'Check whether time out is valid');

    DateTime? pinExpires;
    if (timeoutOrExpirationDate is DateTime) {
      pinExpires = timeoutOrExpirationDate.toUtc();
    } else if (timeoutOrExpirationDate is num) {
      pinExpires = DateTime.now()
          .add(
            Duration(seconds: timeoutOrExpirationDate.toInt()),
          )
          .toUtc();
    }
    return updateMessage(
      message.copyWith(pinned: true, pinExpires: pinExpires),
    );
  }
```

So even though `null` is not a possible parameter, anything that is not a `DateTime` or a `num` will be treated like the old `null` behaviour. So we could for example pass in an empty string `''` instead of `null` and we can trigger that internal logic to allow an unbounded expiration.

So it would be nice to just fix this by making the parameter nullable and removing the null assertion check to allow it.

**Describe alternatives you've considered**

I can specify a really large expiration date, but I think the old behaviour was more elegant.

One more alternative "proposal" would be to make this second parameter optional, defaulting to null. This would be a backwards compatible change.

**Additional context**

None.

imtoori:
fixes #447

### DIFF
--- a/packages/stream_chat/lib/src/api/channel.dart
+++ b/packages/stream_chat/lib/src/api/channel.dart
@@ -517,7 +517,7 @@ class Channel {
         throw ArgumentError('Invalid timeout or Expiration date');
       }
       return true;
-    }(), 'Check for invalid token or expiration date');
+    }(), 'Check whether time out is valid');
 
     DateTime? pinExpires;
     if (timeoutOrExpirationDate is DateTime) {
--- a/packages/stream_chat/lib/src/api/channel.dart
+++ b/packages/stream_chat/lib/src/api/channel.dart
@@ -507,9 +507,9 @@ class Channel {
 
   /// Pins provided message
   Future<UpdateMessageResponse> pinMessage(
-    Message message,
+    Message message, [
     Object? timeoutOrExpirationDate,
-  ) {
+  ]) {
     assert(() {
       if (timeoutOrExpirationDate is! DateTime &&
           timeoutOrExpirationDate != null &&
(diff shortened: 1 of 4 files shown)


xsahil03x:
@imtoori maybe also add some tests so we don't make this mistake sometime again in future :p

packages/stream_chat/lib/src/api/channel.dart
```diff
@@ -507,17 +507,17 @@ class Channel {
...
+  ]) {
     assert(() {
       if (timeoutOrExpirationDate is! DateTime &&
           timeoutOrExpirationDate != null &&
           timeoutOrExpirationDate is! num) {
         throw ArgumentError('Invalid timeout or Expiration date');
       }
       return true;
-    }(), 'Check for invalid token or expiration date');
+    }(), 'Check whether time out is valid');
```
deven98, line 520: Can you correct this to 'timeout' instead of 'time out'?

xsahil03x: sal vatore

imtoori: lol ok

deven98:
LGTM

PR #453, opened 2021-05-21
feat: parse images when pasted

### LINKED ISSUE #395: Pasting images or files from clipboard into message input does not work
Right now it is not possible to paste copied images or files in to the message input, which is a ubiquitous feature on imessage, slack, etc and one that our users expect

imtoori:
I found nothing on this googling around so I decided to do it trying to paste an image in ios and see what gets pasted and how to parse it
Let me know what you think about it, it's a workaround, but it works fine

I didn't find a way to do it for files 

fixes #395

### DIFF
--- a/packages/stream_chat_flutter/lib/src/message_input.dart
+++ b/packages/stream_chat_flutter/lib/src/message_input.dart
@@ -691,6 +691,8 @@ class MessageInputState extends State<MessageInput> {
   String? _previousValue;
 
   void _onChanged(BuildContext context, String s) {
+    _parsePastedImages(s);
+
     if (s == _previousValue) {
       return;
     }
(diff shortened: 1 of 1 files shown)


imtoori:
alspo ta

imtoori:
also, take a look at this https://github.com/flutter/flutter/issues/20796

we may even decide that we don't want to do it based on the fact that is not supported by the framework

xsahil03x:
> also, take a look at this [flutter/flutter#20796](https://github.com/flutter/flutter/issues/20796)
> 
> we may even decide that we don't want to do it based on the fact that is not supported by the framework

yeah, if the framework doesn't support this atm I don't think we should put our efforts into this.

PR #462, opened 2021-05-27
refactor!: v2.0.0

xsahil03x:


packages/stream_chat/lib/src/core/http/interceptor/logging_interceptor.dart
```diff
@@ -0,0 +1,339 @@
...
+  /// you can also write log in a file.
+  void Function(InterceptStep step, Object object) logPrint;
+
+  @override
+  void onRequest(RequestOptions options, RequestInterceptorHandler handler) {
+    if (request) {
+      _printRequestHeader(_logPrintRequest, options);
+    }
+    if (requestHeader) {
+      _printMapAsTable(
```
Nash0x7E2, line 81: This is cool!

Nash0x7E2 (approved):
The code for this PR looks great! My only concern is around the lack of API documentation for public classes and methods. 

Let's make sure we keep track of this, and we add documentation to all methods before releasing it to pub-even as a pre-release. 

When it comes to adding docs, the content should answer the following questions:
- What is CLASS/METHOD?
- How is it used?
- How can I customize it?
- Where is it used?
- Example code?

imtoori (approved):
image

PR #465, opened 2021-05-31
feat: Swipe actions

deven98:
Added callbacks for default slidable actions


xsahil03x:
@deven98 I will suggest exposing actions (primary, secondary) parameter directly than `onDelete`, `onMore` callback.

deven98:
@xsahil03x That'll required exposing the Slidable IconSlidable action too, which I don't really want to do

imtoori (approved):
LGTM

PR #497, opened 2021-06-25
feat: make default message widget easier to reuse

imtoori:


packages/stream_chat_flutter/lib/src/message_list_view.dart
```diff
@@ -56,32 +59,37 @@ typedef ReplyTapCallback = void Function(Message);
...
-  bool? isLastUser;
+  late final bool isLastUser;
 
   /// True if the user message is the same of the next message
-  bool? isNextUser;
+  late final bool isNextUser;
 
   /// The message
-  Message message;
+  late final Message message;
```
xsahil03x, line 89: Not really needed
```suggestion
  final Message message;
```

packages/stream_chat_flutter/lib/src/message_list_view.dart
```diff
@@ -56,32 +59,37 @@ typedef ReplyTapCallback = void Function(Message);
...
 
   /// The index of the message
-  int index;
+  late final int index;
```
xsahil03x, line 92: Not really needed
```suggestion
  final int index;
```

packages/stream_chat_flutter/lib/src/message_list_view.dart
```diff
@@ -1130,6 +1074,21 @@ class _MessageListViewState extends State<MessageListView> {
...
+        MessageDetails(
+          userId,
+          message,
+          messages,
+          index,
+          messageWidget as MessageWidget,
+        ),
+        messages,
+      );
+    }
```
xsahil03x, line 1089: # Do not commit this
maybe something like this
```suggestion
    if (widget.messageBuilder != null) {
      messageWidget = widget.messageBuilder!(
        context,
        MessageDetails(
          userId,
          message,
          index,
        ),
        messages,
        messageWidget as MessageWidget,
      );
    }
```

packages/stream_chat_flutter/lib/src/message_list_view.dart
```diff
@@ -1130,6 +1072,21 @@ class _MessageListViewState extends State<MessageListView> {
...
+    if (widget.messageBuilder != null) {
+      messageWidget = widget.messageBuilder!(
+        context,
+        MessageDetails(
+          userId,
+          message,
+          messages,
+          index,
+        ),
+        messages,
```
xsahil03x, line 1084: two `messages` 😩

imtoori: yea but one is to populate the MessageDetails, the other one it's passed in the builder

xsahil03x: but not really needed as part of message details

imtoori: but we don't have any instance of the message list in the message details
it's just for computing the message details properties

xsahil03x: got it

deven98 (approved):
LGTM

PR #501, opened 2021-06-26
fix: queryMembers payload 

### LINKED ISSUE #498: Query Members Error
**Describe the bug**
When trying to query members I am getting the below error even though I have added a filter as shown below. This was working before changing my package to the ones below.

```
getMembers() {
    return widget.channel.queryMembers(
        sort: [chat.SortOption('last_message_at')],
        pagination: chat.PaginationParams(limit: 25),
        filter: chat.Filter.and([
          chat.Filter.equal('banned', false),
        ])).then((value) {
      setState(() {
        members.addAll(value.members);
        _isLoading = false;
      });
    });
  }
```

**What package are you using? What version?**
```
stream_chat_flutter: 
    git:
      url: https://github.com/GetStream/stream-chat-flutter.git
      ref: develop
      path: ./packages/stream_chat_flutter
dependency_overrides:
    stream_chat:
      git:
        url: https://github.com/GetStream/stream-chat-flutter.git
        ref: develop
        path: ./packages/stream_chat
    stream_chat_flutter_core:
      git:
        url: https://github.com/GetStream/stream-chat-flutter.git
        ref: develop
        path: ./packages/stream_chat_flutter_core
```

**What platform is it about?**
- [x] Android
- [ ] iOS
- [ ] Web
- [ ] Windows
- [ ] MacOS
- [ ] Linux

`[ERROR:flutter/lib/ui/ui_dart_state.cc(199)] Unhandled Exception: StreamChatNetworkError(code: 4, message: QueryMembers failed with error: "filter_conditions is a required field", statusCode: 400, data: ErrorResponse(code: 4, message: QueryMembers failed with error: "filter_conditions is a required field", statusCode: 400, moreInfo: https://getstream.io/chat/docs/api_errors_response))`

imtoori:
improved api tests checking for args

### DIFF
--- a/packages/stream_chat/lib/src/client/channel.dart
+++ b/packages/stream_chat/lib/src/client/channel.dart
@@ -989,11 +989,11 @@ class Channel {
   /// List the reactions for a message in the channel
   Future<QueryReactionsResponse> getReactions(
     String messageId, {
-    PaginationParams? options,
+    PaginationParams? pagination,
   }) =>
       _client.getReactions(
         messageId,
-        options: options,
+        pagination: pagination,
       );
 
   /// Retrieves a list of messages by ID
--- a/packages/stream_chat/lib/src/client/client.dart
+++ b/packages/stream_chat/lib/src/client/client.dart
@@ -1149,11 +1149,11 @@ class StreamChatClient {
   /// Get all the reactions for a [messageId]
   Future<QueryReactionsResponse> getReactions(
     String messageId, {
-    PaginationParams? options,
+    PaginationParams? pagination,
   }) =>
       _chatApi.message.getReactions(
         messageId,
-        options: options,
+        pagination: pagination,
       );
 
   /// Update the given message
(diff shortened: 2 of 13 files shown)


xsahil03x:
@imtoori maybe also update all the verify calls.

imtoori:
Tests would fail even if we don't modify the verify calls, so I think we're good

xsahil03x:
> Tests would fail even if we don't modify the verify calls, so I think we're good

but then `any(named: '')` was also good enough as we don't really have any logic around the API data.

imtoori:
But if you modified the payload the tests would have been successful even with a wrong payload
Now they fail if you modify the payload without changing the tests

imtoori:
It's like we were testing just the api method (get,post,put..)

imtoori:
And the url

xsahil03x (approved):
LGTM

PR #520, opened 2021-07-02
feat!: connection less client [CDS-335]

imtoori:


xsahil03x:
is this breaking?

packages/stream_chat/lib/src/client/client.dart
```diff
@@ -289,15 +324,15 @@ class StreamChatClient {
     } catch (e, stk) {
       if (e is StreamWebSocketError && e.isRetriable) {
         final event = await _chatPersistenceClient?.getConnectionInfo();
-        if (event != null) return event;
+        if (event != null) return event.me;
```
xsahil03x, line 327: we should return `ownUser` in this too instead of `event.me` as the user properties might have changed over time.

We can do something like:
```dart
return event.me.merge(ownUser);
```

packages/stream_chat/lib/src/client/client.dart
```diff
@@ -327,7 +362,8 @@ class StreamChatClient {
         _ws.connectionStatusStream.skip(1).listen(_connectionStatusHandler);
 
     try {
-      return await _ws.connect(user);
+      await _ws.connect(user);
+      return state.user!;
```
xsahil03x, line 366: maybe?

```dart
      final event = await _ws.connect(user);
      return event.me.merge(ownUser);
```

imtoori: I think there's no need for this

imtoori:
it's breaking
connectUser calls return `OwnUser?` and not `Event`

packages/stream_chat/lib/src/client/client.dart
xsahil03x: I just don't want to miss any user information

imtoori: yeah but state.user is updated using the events and it's the latest version in the backend right?

xsahil03x (approved):
LGTM

PR #558, opened 2021-07-19
feat(ui, core): message list pagination limit [CDS-352]

### LINKED ISSUE #539: PaginationParams in MessageListCore
**Is your feature request related to a problem? Please describe.**
Just like in the ChannelListCore, can we get PaginationParams for MessageListCore as well, because ultimately it should be UI call as many as messages they want to show?

Edit: I noticed now that, MessageListCore is returning all messages available in the channel.

**Describe the solution you'd like**
Introduce PaginationParams in MessageListCore
`pagination: PaginationParams(limit: 10)`

I am using  `stream_chat_flutter_core: 2.0.0-nullsafety.9`

@imtoori @deven98 @xsahil03x

imtoori:
fixes #539


packages/stream_chat_flutter/lib/src/message_input.dart
```diff
@@ -2309,6 +2309,7 @@ class __PickerWidgetState extends State<_PickerWidget> {
 
           return InkWell(
             onTap: () async {
+              print('on tap');
```
xsahil03x, line 2312: Remove print :}

xsahil03x (approved):
LGTM

packages/stream_chat_flutter/lib/src/message_input.dart
imtoori:
```suggestion
```

PR #560, opened 2021-07-20
feat(ui): Replace random png

deven98:
Added new default avatar


imtoori:
can you add tests?

xsahil03x:
@deven98 maybe also add some tests for Users with non-alphanumeric characters in their name

imtoori (dismissed):
LGTM

PR #561, opened 2021-07-20
feat(ui): reverse message list view

imtoori:


packages/stream_chat_flutter/lib/src/message_list_view.dart
```diff
@@ -700,8 +710,11 @@ class _MessageListViewState extends State<MessageListView> {
                       );
                     }
                   },
-                  child: StreamSvgIcon.down(
-                    color: _streamTheme.colorTheme.textHighEmphasis,
+                  child: Transform.rotate(
+                    angle: widget.reverse ? 0 : pi,
+                    child: StreamSvgIcon.down(
+                      color: _streamTheme.colorTheme.textHighEmphasis,
+                    ),
```
xsahil03x, line 717: We should add `StreamSvgIcon.up` instead of Transforming it

xsahil03x (approved):
LGTM

PR #568, opened 2021-07-21
fix(ui): list background

imtoori:


packages/stream_chat_flutter/lib/src/message_search_list_view.dart
```diff
@@ -144,61 +144,69 @@ class _MessageSearchListViewState extends State<MessageSearchListView> {
...
+
+    final backgroundColor =
+        MessageSearchListViewTheme.of(context).backgroundColor;
+    return backgroundColor != null
+        ? ColoredBox(
+            color: backgroundColor,
+            child: messageSearchListCore,
+          )
+        : messageSearchListCore;
+  }
```
xsahil03x, line 209: LGTM, but I would have written this similar to message-list-view for consistency and easier to read.

packages/stream_chat_flutter/lib/src/user_list_view.dart
```diff
@@ -160,38 +160,44 @@ class _UserListViewState extends State<UserListView>
...
 
+    final backgroundColor = UserListViewTheme.of(context).backgroundColor;
+
+    final child = backgroundColor != null
+        ? ColoredBox(
+            color: backgroundColor,
+            child: userListCore,
+          )
+        : userListCore;
+
```
xsahil03x, line 200: LGTM, but I would have written this similar to message-list-view for consistency and easier to read.

xsahil03x (approved):
LGTM

PR #589, opened 2021-07-29
chore(repo): 2.1.1

imtoori:


xsahil03x (approved):
LGTM, should we also bump the version in localization though?

PR #602, opened 2021-08-02
feat(localization): support for Japanese and Korean

sachaarbonel:
I don't guarantee any of those translations to be accurate @EricSSartorius


packages/stream_chat_localizations/lib/src/stream_chat_localizations_ja.dart
```diff
@@ -0,0 +1,349 @@
...
+  String get sendLabel => '送信';
+
+  @override
+  String get withText => 'と';
+
+  @override
+  String get inText => 'に';
+
+  @override
+  String get youText => 'お客様';
```
EricSSartorius, line 334: This may be better as あなた and remove the comment

packages/stream_chat_localizations/lib/src/stream_chat_localizations_ko.dart
```diff
@@ -0,0 +1,352 @@
...
+  String get sendLabel => '보내기';
+
+  @override
+  String get withText => '함께';
+
+  @override
+  String get inText => '에';
+
+  @override
+  String get youText => '고객님';
```
EricSSartorius, line 338: This may be better as 당신 and remove the comment

imtoori (approved):
LGTM

PR #607, opened 2021-08-04
feat(llc, ui): add name, image property to User and OwnUser

HayesGordon:
Add `name`, `image` property to `User` and `OwnUser` classes.

Setting the `name` in the constructor will also set it on `extraData['name']`, using the "name" key.

The getter for `name` will return image from extraData['name'] if available, else will return `id`.

Setting the `image` in the constructor will also set it on `extraData['image']`, using the "image" key.

The getter for `image` will return image from extraData['image'].

---

Add additional testing for `OwnUser` and `User`.


xsahil03x (approved):
LGTM

PR #614, opened 2021-08-08
fix(llc, persistence): enable pragma foreign_keys

### LINKED ISSUE #604: Cascade deletion for `stream_chat_persistance` database is not working.
Hi! First of all, I'm very new to `moor` package and your `stream_chat_persistance` is a great example on how to setup and work with a database, thank you!

**Describe the bug**
While setting up moor, I found out that cascade deletion that you declare here https://github.com/GetStream/stream-chat-flutter/blob/master/packages/stream_chat_persistence/lib/src/entity/messages.dart#L81 is not working because `PRAGMA foreign_keys = ON` is not set. I was able to reproduce it with the next code example:

```dart
final userId = '123';
final channelId = '435:world';
final _client = StreamChatPersistenceClient();
await _client.connect(userId);

final cannel = ChannelModel(id: channelId, cid: channelId);
await _client.updateChannels([cannel]);

final message = Message(id: '3465');
await _client.updateMessages(channelId, [message]);

await _client.deleteChannels([channelId]);

final existedMessages = await _client.getMessagesByCid(channelId)
final existedChannels = await _client.getChannelByCid(channelId);

print(existedMessages); // prints `Message instance`, but in the docs it says that all messages of a cid will be deleted
print(existedChannels); // prints null
```
reproduces with `stream_chat_persistence: ^2.1.1` and `stream_chat: ^2.1.1`.

Another part that proves the problem is that for upsert you use `mode: InsertMode.insertOrReplace` which according to the docs deletes the row if it already exists which should trigger delete cascade, but it doesn't get triggered right now. If the pragma settings is on, then it works, **but leads to the data loss** (tested on my database 😞 ).

I'm not sure, that the solution I found for my database suits you, but I hope it helps.
1. I ended up setting pragma as described here: https://github.com/simolus3/moor/issues/41#issuecomment-504543642
    ```dart
    @override
    MigrationStrategy get migration => MigrationStrategy(
        beforeOpen: (details) async {
          await customStatement('PRAGMA foreign_keys = ON;');
        },
     );
    ```
2. Instead of `mode: InsertMode.insertOrReplace` I switched to `into(...).insertOnConflictUpdate`, which works just fine and doesn't perform deletions.

Also, could you please clarify one thing to me: I'm looking at your db migrations code https://github.com/GetStream/stream-chat-flutter/blob/master/packages/stream_chat_persistence/lib/src/db/moor_chat_database.dart#L58 and if I understand it correctly, on a schema version update it will drop all tables with their data and then create new empty ones. Am I right about this behaviour?

incendial:
# Submit a pull request

## CLA

Closes #604


incendial:
@xsahil03x while fixing the problem with foreign keys, I stepped into another problem with pinned messages. Right now reactions have a [foreign key for messages](https://github.com/GetStream/stream-chat-flutter/blob/develop/packages/stream_chat_persistence/lib/src/entity/reactions.dart#L13), but not the pinned messages which stops working right after foreign keys check is enabled (and thats why new changes in tests for that part are failing).

I'm not sure how this should be fixed (for example, I can add a separate foreign key for pinned messages and make both keys  nullable) and kinda think that it's not my decision to make. Maybe you have some ideas on what's the best solution for that problem?

xsahil03x:
Hey @incendial, thanks for the awesome PR 😄. And about the issue, I think the easiest would be to create a separate reaction table for pinned message reactions but the best I can think of is to somehow delete pinned messages table and merge it with the regular message table.

This is really a great start and I can also help you implement this in our upcoming sprint which starts on Wednesday.

Thanks Again ❤️

incendial:
> I think the easiest would be to create a separate reaction table for pinned message reactions but the best I can think of is to somehow delete pinned messages table and merge it with the regular message table.

@xsahil03x before taking a decision here and providing the implementation, could you clarify why a separate table for pinned messages was created at the first place? Is it some kind of an optimization? Or just looked as a good decision at its time?

> This is really a great start and I can also help you implement this in our upcoming sprint which starts on Wednesday.

No problem, I think I can do it by myself.

xsahil03x:
> @xsahil03x before taking a decision here and providing the implementation, could you clarify why a separate table for pinned messages was created at the first place? Is it some kind of an optimization? Or just looked as a good decision at its time?

We created a separate table because we were getting them as separate fields via API and also to make a clear separation with regular messages. But I think we should get rid of it now and handle them using regular messages table. We can add required methods for `pinnedMessages` inside `message_dao`

incendial:
> We created a separate table because we were getting them as separate fields via API and also to make a clear separation with regular messages. But I think we should get rid of it now and handle them using regular messages table. We can add required methods for pinnedMessages inside message_dao

Thanks for clarification! I'll take a closer look on how to implement this tomorrow.

incendial:
@xsahil03x looks like pinned messages have all the same properties as regular messages, so I removed pinned messages and  added `isPinned` parameter to some methods in messages_dao (ex. https://github.com/GetStream/stream-chat-flutter/pull/614/files#diff-87ddffbd692d1f1c8ef8d8ef3a0284eff59b5190d3afe95588c8c2c21b957388R30). What do you think about this approach? Or it's better to have separate methods for pinned messages instead?

Also, should I add `isPinned` to the `updateMessages` method too or it's the responsibility of a user to provide correct `pinned` value in message model?

xsahil03x:
Hey @incendial,

> added `isPinned` parameter to some methods in messages_dao (ex. [#614 (files)](https://github.com/GetStream/stream-chat-flutter/pull/614/files#diff-87ddffbd692d1f1c8ef8d8ef3a0284eff59b5190d3afe95588c8c2c21b957388R30)).

I think it's better to have different methods than having a boolean. we should just add `pinned` before the `message` for every method name.

> Also, should I add `isPinned` to the `updateMessages` method too or it's the responsibility of a user to provide correct `pinned` value in message model?

We don't really need `isPinned` for updating and adding messages in DB because we can easily filter them while querying using the `isPinned` field present inside the `MessageEntity`. 

Yes it's the user's responsibility to provide the pinned value in message model, we use `false` as the default value.

incendial:
> I think it's better to have different methods than having a boolean. we should just add pinned before the message for every method name.

Sure, I'll replace it with separate methods then.

incendial:
@xsahil03x hi, I think the PR is ready for review.

xsahil03x:
Thanks @incendial, I'll review it soon 😄

xsahil03x:
Hey @incendial, I just discussed the real reason we were maintaining a separate table for `PinnedMessages`. It was because we only fetch a limited number of messages on the first load and pinned messages can be way past those messages.

So I think we have to reset those changes and create a separate `Reaction` table for pinned messages. Don't worry, I will do those changes if you are busy.

xsahil03x:
Hey @incendial, Sorry again for mentioning this late. Our team lead was on a vacation and I totally forgot why we created a separate table for pinned messages. 🙏🏼 

Okay so the problem we faced at that time was, Let's say
1. We opened a newly installed app.
2. It fetches 50-80 messages (which includes pinned messages) and saved them in persistence.

Now there may be a case where the user only wants to see all the pinned messages. So he
1. Fetches all the messages from persistence.
2. Filtered them on the basis of the `pinned` flag.

Now the problem here is that we can't be sure if these messages are the only pinned messages of this channel. There may be some more old pinned messages but we are not showing them because we don't have them yet in our persistence.

And as I previously mentioned in one of my earlier comments.
> We are getting them as separate fields via API.

```dart
messages: [Messages] // this contains all the messages [regular + pinned] in a certain sort of order.
pinned_messages: [Messages] // this contains all the pinned messages of this channel.
```

incendial:
> Hey @incendial, Sorry again for mentioning this late. Our team lead was on a vacation and I totally forgot why we created a separate table for pinned messages. 🙏🏼

It's okay, don't get me wrong.

> Now the problem here is that we can't be sure if these messages are the only pinned messages of this channel. There may be some more old pinned messages but we are not showing them because we don't have them yet in our persistence.

I don't understand how a separate table solves this problem. 
If we look at the approach with 2 tables and let's assume we fetch 80 messages and 5 pinned messages in a separate field and save it to persistence. Then we'll have 80 in Messages table and 5 in PinnedMessages table.
If we look at the approach with 1 table and fetch the same amount of messages and pinned messages and save them, we'll end up with 85 messages in the table, but selecting regular messages will return 80 messages and selecting pinned messages will return 5.

So, both approaches don't tell anything about how many pinned message were not loaded, they only differ in how they are stored.

Maybe I'm missing something?

incendial:
> // this contains all the pinned messages of this channel.

Okay, I've missed your comments in a code block 😅. But why we can't just save them into messages table?

xsahil03x:
> > // this contains all the pinned messages of this channel.
> 
> Okay, I've missed your comments in a code block 😅. But why we can't just save them into messages table?

Because then it will create problems with pagination as our current implementation depends on using the `messageId` `lessThan`/`greaterThan` to filter next or previous messages.

So, then there might be cases where we will also get old pinned messages as part of message list.
```dart
MessageTable:

MessageDate | MessageId
	22 July		12	(Pinned) -> this one we get from pinned_messages
	34 Sep		42	(Regular)
	35 Sep		43	(Regular)
	35 Sep		44	(Pinned)
	35 Sep		45	(Regular)
```

If we paginate `lessThanOrEqual` to `44` we will get
```dart
messages: [
	Message(id: 12) // Very old message |
	Message(id: 42)						|
	Message(id: 43)						|-----> But the user will never expect the old pinned message
	Message(id: 44)						|
	Message(id: 45)						|
]
```

incendial:
But messages are requested from the DB with `..where(messages.pinned.equals(isPinned))` which happens before pagination calculation. How pinned messages can affect regular messages then?

Note: I'm just curious about the limitations and what is the best way to resolve this problem before changing the code (I'll become even sadder if we'll need to rewrite it one more time 😄😄😄). But if you think that it's better to have a separate table (which I'm totally ok with), just tell me and I'll stop with questions 😄.

xsahil03x:
Hey @incendial, I think you are getting confused between `RegularMessages` List and `PinnedMessages` List.

Regular Messages List -> Will contains all the messages no matter they are pinned or not.
```dart
messages: [
	Message(id: 12, pinned: true),
	Message(id: 42, pinned: false),
	Message(id: 43, pinned: false),
	Message(id: 44, pinned: true),
	Message(id: 45, pinned: false),
]
```
Pinned Messages List -> Will only contain pinned messages.
```dart
messages: [
	Message(id: 12, pinned: true),
	Message(id: 44, pinned: true),
]
```

incendial:
> Hey @incendial, I think you are getting confused between RegularMessages List and PinnedMessages List.

Ah, okay, but why you load pinned messages twice (full list as a separate field and partially inside regular messages)?

xsahil03x:
> Ah, okay, but why you load pinned messages twice (full list as a separate field and partially inside regular messages)?

We are not actually. The `regular` messages list contains a limited number of `recent` messages depending on the pagination limit. But on the other hand, the `pinned` messages list will contain all the pinned messages even if they are 2-3 months old.

incendial:
> We are not actually. The regular messages list contains a limited number of recent messages depending on the pagination limit. 

As I understood from the example above https://github.com/GetStream/stream-chat-flutter/pull/614#issuecomment-899579640, this list can also contain pinned messages. Will those pinned message be also included into the full pinned messages list?

imtoori:
@incendial yes, they are duplicated because the full message list contains a part of the full message list, while the pinned message list contains just the messages that are pinned
The problem with using a single table is that we want to have a continuous list of messages in the message table so that we can query the list without thinking about "holes" in the list

incendial:
@imtoori thanks for clarification!

I was writing a new question when I realized why you might need the approach described here: https://github.com/GetStream/stream-chat-flutter/pull/614#issuecomment-899579640 - I was not able to find a proof in your app examples, but you probably show recent messages in feed (or list) and thats where regular messages and pinned messages are mixed up. And then you have some kind of a separate place for all pinned messages and that is where you need all pinned messages of a channel.

@xsahil03x @imtoori thank you so much for you patience and answers, it was very interesting to learn about your product!

If you okay with timings I can update the request in a few days or so.

incendial:
@xsahil03x hi! I've reverted all changes for pinned messages table and added a new table, dao, tests, etc. for pinned message reactions. I've also noticed that the changes I made affects `ChatPersistenceClient` and I'm not sure, how it should be updated properly. Could you pick up the request at this point? I think it'll be more efficient.

imtoori:
@incendial yes we can take it from here, thanks a lot for the work done on this pr :)

incendial:
> thanks a lot for the work done on this pr :)

Thank you for your patience and help, sorry for not finishing it.

imtoori (approved):
LGTM

PR #617, opened 2021-08-10
feat(llc): add name, image get, set and update on channel 

HayesGordon:
This will add shortcuts to get, set and update the name and image of a `Channel`, instead of updating `extraData` directly.

### DIFF
--- a/packages/stream_chat/lib/src/client/channel.dart
+++ b/packages/stream_chat/lib/src/client/channel.dart
@@ -140,7 +140,7 @@ class Channel {
     if (_initializedCompleter.isCompleted) {
       throw StateError(
         'Once the channel is initialized you should use channel.updateName '
-        'to update the channel image',
+        'to update the channel name',
       );
     }
     _extraData.addAll({'name': name});
--- a/packages/stream_chat/lib/src/client/channel.dart
+++ b/packages/stream_chat/lib/src/client/channel.dart
@@ -63,145 +63,145 @@ class Channel {
     _extraData.addAll(extraData);
   }
 
-  /// Returns true if the channel is muted
+  /// Returns true if the channel is muted.
   bool get isMuted =>
       _client.state.currentUser?.channelMutes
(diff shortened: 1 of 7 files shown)


packages/stream_chat/lib/src/client/channel.dart
```diff
@@ -210,23 +252,43 @@ class Channel {
...
+  ///
+  /// The channel needs to be initialized.
+  ///
+  /// {@macro image}
+  Stream<String?> get imageStream {
+    _checkInitialized();
+    return state!.channelStateStream.map(
+      (cs) =>
+          (cs.channel?.extraData['image'] as String?) ??
+          (_extraData['image'] as String?),
```
xsahil03x, line 278: this can be `image` instead, we already have a getter above

packages/stream_chat/lib/src/client/channel.dart
```diff
@@ -792,7 +860,36 @@ class Channel {
...
+    _checkInitialized();
+
+    return _client.updateChannelPartial(
+      id!,
+      type,
+      set: {
+        'image': image,
+      },
+    );
+  }
```
xsahil03x, line 890: can be replaced by 
```dart
return updatePartial(set:{'image':image});
```

packages/stream_chat/lib/src/client/channel.dart
```diff
@@ -63,145 +120,171 @@ class Channel {
...
+
+  /// Shortcut to set channel name.
+  ///
+  /// {@macro name}
+  set name(String? name) {
+    if (_initializedCompleter.isCompleted) {
+      throw StateError(
+        'Once the channel is initialized you should use channel.updateName '
+        'to update the channel image',
```
xsahil03x, line 143:
```suggestion
        'to update the channel name',
```

HayesGordon:
The latest changes you made look good!

xsahil03x (approved):
LGTM

PR #619, opened 2021-08-11
refactor(ui): use channel `.nameStream` and `.imageStream` instead of `.extraDataStream`

xsahil03x:


packages/stream_chat_flutter/lib/src/channel_avatar.dart
```diff
@@ -108,7 +107,7 @@ class ChannelAvatar extends StatelessWidget {
                   imageUrl: channelImage,
                   errorWidget: (_, __, ___) => Center(
                     child: Text(
-                      extraData['name']?[0] ?? '',
+                      channel.name?[0] ?? '',
```
funwithflutter, line 110: I'm not sure if this will work? Will channel.name automatically update? Or are we fine with displaying an outdated name, and just getting updates for the image?

xsahil03x: `channel.name` is a getter so it will always give back the latest name

funwithflutter: Will it though?  Does the `Channel` get updated when the channel state changes? Or is only new values emitted from the different streams?

HayesGordon:
LGTM

HayesGordon (dismissed):

imtoori (approved):
LGTM

PR #620, opened 2021-08-11
feat(ui): add possibility to limit attachments in `MessageInput`

### LINKED ISSUE #582: Attachments limiting
It would be nice to have a way to set attachments to be 1 per message or X number up to 10.

xsahil03x:
Fixes #582


imtoori (approved):
LGTM

PR #624, opened 2021-08-13
fix(ui): Fix `MessageInput` textfield enabling behaviour

jiahan-wu:


xsahil03x:
Hey @jiahan-wu, thanks for the fix 😄 
Can we do this instead?
```dart
 if (pickedFile != null) {
      final bytes = await pickedFile.readAsBytes();
      file = AttachmentFile(
        size: bytes.length,
        path: pickedFile.path,
        bytes: bytes,
      );
  }
```

and then it will be automatically called on line [1900](https://github.com/GetStream/stream-chat-flutter/pull/624/files#diff-ab9bbd9912f791b2dcaa78c30aa5a75289f7dce3a247f6dc3814067926f3bbc7R1900)

jiahan-wu:
@xsahil03x Sure.😄

packages/stream_chat_flutter/CHANGELOG.md
```diff
@@ -28,6 +28,10 @@ Here's the full naming breakdown:
 * `MessageTheme` is now `MessageThemeData`
 * `UserListViewTheme` is now `UserListViewThemeData`
 
+🐞 Fixed
+
+- Fixed `MessageInput` textField behaviour
```
xsahil03x, line 33:
```suggestion
- Fixed `MessageInput` textField case where `input` is not enabled if the file picked from the camera is null.
```

xsahil03x (approved):
LGTM

PR #631, opened 2021-08-18
ci(repo): add verify-semantic-changelog-update action workflow

xsahil03x:
# Verify Semantic Changelog Update

This is a [Github Action](https://github.com/GetStream/verify-semantic-changelog-update) that ensures your PR contains a changelog entry for all the user facing changes.

## Configuration

The action works without configuration, however you can provide options for customization.

```yml
        with:
          # Configure which types should be verified.
          # Default: fix, feat, *! (Any breaking change type)
          types: |
            fix
            feat
          # Configure which scopes are allowed along with their path.
          # Default: Only checks for the top-level changelog entry
          scopes: |
            {
              "llc": "packages/stream_chat",
              "ui": "packages/stream_chat_flutter",
              "core": "packages/stream_chat_flutter_core",
              "localization": "packages/stream_chat_flutter_localizations",
              "persistence": "packages/stream_chat_persistence"
            }
          # Configure the changelog file path/name.
          # Default: changelog.md
          path: |
            changelog.md
```

### DIFF
--- a/.github/workflows/pr_title.yml
+++ b/.github/workflows/pr_title.yml
@@ -1,4 +1,4 @@
-name: 'PR Title is Conventional and Semantic'
+name: 'PR is Conventional and Semantic'
 on:
   pull_request_target:
     types:
--- a/.github/workflows/pr_title.yml
+++ b/.github/workflows/pr_title.yml
@@ -1,4 +1,4 @@
-name: 'PR Title is Conventional'
+name: 'PR Title is Conventional and Semantic'
 on:
   pull_request_target:
     types:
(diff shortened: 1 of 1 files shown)


.github/workflows/pr_title.yml
```diff
@@ -1,4 +1,4 @@
-name: 'PR Title is Conventional'
+name: 'PR Title is Conventional and Semantic'
```
xsahil03x, line 1:
```suggestion
name: 'PR is Conventional and Semantic'
```

imtoori (approved):
LGTM

PR #652, opened 2021-08-31
fix(llc): unread count not updating

imtoori:


funwithflutter (commented):
I would rename the method to be _computeUnread, as it's happening in two places now. Don't know if it will be easy to add a test anywhere to ensure that the unread is updated. Otherwise it looks good to me

PR #655, opened 2021-09-01
refactor(ui): removed clipping from user avatar

deven98:


packages/stream_chat_flutter/CHANGELOG.md
```diff
@@ -46,6 +46,7 @@ breakdown:
 * `UserListViewTheme` is now `UserListViewThemeData`
 
 - Updated core dependency.
+- `UserAvatar` no longer uses a clipper by default.
```
imtoori, line 49: this is in the wrong place
I think we don't really need to add an entry for this

imtoori (changes requested):

PR #658, opened 2021-09-03
feat(ui): refactor overlays

deven98:


packages/stream_chat_flutter/lib/src/message_input.dart
```diff
@@ -472,7 +473,28 @@ class MessageInputState extends State<MessageInput> {
         child: child,
       );
     }
-    return child;
+    return AnchoredOverlay(
+      showOverlay: _commandsOverlay != null ||
+          _emojiOverlay != null ||
+          _mentionsOverlay != null,
```
xsahil03x, line 479: Can't we create 3 booleans for this instead?

imtoori: I suggested we create three different widget instead of using just one Overlay this way
What do you think about this?

For example the `EmojiOverlay` could have all the logic about picking emojis 

I think it falls also under the spike call we had yesterday

deven98: Unlike the earlier impl, in this case we only use these booleans twice and only in the same code block, so LOC would not change either way, I'd suggest this being as it is.

packages/stream_chat_flutter/lib/src/attachment/giphy_attachment.dart
```diff
@@ -48,6 +48,7 @@ class GiphyAttachment extends AttachmentWidget {
   }
 
   Widget _buildSendingAttachment(BuildContext context, String imageUrl) {
+    print('sending attachment GIPHY $imageUrl');
```
deven98: Extra print

deven98:
```suggestion
```

packages/stream_chat_flutter/lib/src/message_widget.dart
```diff
@@ -1153,10 +1157,15 @@ class _MessageWidgetState extends State<MessageWidget>
   }
 
   Widget _parseAttachments() {
+    if (widget.message.isEphemeral) {
+      print('EPHEMERAL');
+    }
```
deven98:
```suggestion
```

imtoori: this is outdated

packages/stream_chat_flutter/lib/src/attachment/giphy_attachment.dart
imtoori: this is outdated

deven98:
Screenshot 2021-09-20 at 4 12 26 PM

Overlay padding seems to be off for some reason

PR #660, opened 2021-09-06
fix(llc, ui, core): list pagination

xsahil03x:


packages/stream_chat_flutter_core/lib/src/message_search_list_core.dart
```diff
@@ -38,7 +38,7 @@ class MessageSearchListCore extends StatefulWidget {
   /// * [errorBuilder]
   /// * [loadingBuilder]
   /// * [childBuilder]
-  const MessageSearchListCore({
+  MessageSearchListCore({
```
xsahil03x, line 41: add const back :|

imtoori (approved):
LGTM

PR #662, opened 2021-09-07
feat(llc): add `Filter.contains` and `Filter.empty`

imtoori:


packages/stream_chat/lib/src/core/models/filter.dart
```diff
@@ -157,13 +161,20 @@ class Filter extends Equatable {
...
+
   /// Creates a custom [Filter] if there isn't one already available.
   const factory Filter.custom({
     required Object value,
     String? operator,
     String? key,
   }) = Filter.__;
 
+  /// An empty filter
+  factory Filter.empty() => const Filter.raw(value: {});
```
xsahil03x, line 176: const factory maybe

imtoori: Only redirecting factory constructors can be declared to be `const`

docusaurus/docs/Flutter/guides/understanding_filters.mdx
```diff
@@ -107,6 +107,36 @@ The 'exists' filter matches values that exist, or don't exist, based on the spec
...
+```dart
+Filter.raw(value: {
+  'members': [
+    ..._selectedUsers.map((e) => e.id),
+    chatState.currentUser!.id,
+  ],
+  'distinct': true,
+});
+```
+
```
xsahil03x, line 139: can you also add example for `Filter.custom`?

imtoori: done

xsahil03x (approved):
LGTM

Issue #673, opened 2021-09-10
[Help] Search is working in development; not working in Testflight & AppleStore


Abacaxi-Nelson:
### Which packages are you using?

stream_chat, stream_chat_persistance

### On what platforms did you experience the issue?

iOS

### What version are you using?

stream_chat_flutter: ^2.2.1
stream_chat_persistence: ^2.2.0

### What happened?

Its weird, but after half day trying to figure it out; i convince my self to ask for some help.
Same app:
![IMG-2365](https://user-images.githubusercontent.com/556990/132860300-0291de0e-751a-4f8a-afb6-b9066e0d4ce8.PNG)
![IMG-2364](https://user-images.githubusercontent.com/556990/132860278-edc56389-32c5-473c-8ca5-2ebd71561d30.jpg)
 
First one is Testflight version, when search a user dont work.
Second screen in my dev env, when search works...

One thing that coud explain, is the api call is not triggered; cause i dont see the log in 'Chat Logs/Query Users' for testflight version; but see it for dev app..

### Steps to reproduce

```bash
Using your example app..
Rest of the app workin great :)
If you have an idea, please share :)
```

### Supporting info to reproduce

```dart
await chatClient.disconnectUser();
      await chatClient.connectUser(
        chat.User(
          id: uid,
          extraData: {
            'image': user.imageUrl ??
                'XXXX',
            'name': user.userName,
          },
        ),
        response.body,
      );
```
Using your example app..

### Flutter doctor output

```shell
Flutter 2.5
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Abacaxi-Nelson:
really strange.
launching in debug mode work.
launching in release/profile mode dont work: ```queryUsers is not called when name is typed```
https://github.com/GetStream/stream-chat-flutter/blob/18c6fd842c86767461adbd5adfd71770d34922f1/packages/stream_chat_flutter_core/lib/src/users_bloc.dart#L65

xsahil03x:
@Abacaxi-Nelson check if you have the internet permission available in the profile and release flavor.

Abacaxi-Nelson:
Hey @xsahil03x 
I do have it cause rest of the app is working.
Im able to display my chanels, and even send message in them.

Abacaxi-Nelson:
hey @xsahil03x 
i think i narrow down the problem.
[...]

PR #682, opened 2021-09-14
fix(core): match oldWidget custom type params using jsonEncode in didUpdate.

xsahil03x:
Fixes: #673


imtoori (approved):

imtoori:
Merging this even if the analysis is failing 
we'll open another PR for this since it's mostly due to flutter 2.5

PR #686, opened 2021-09-15
feat(core, ui): Add `Filter.empty` as the default filter in `UserListView`.

xsahil03x:


packages/stream_chat_flutter/lib/src/user_list_view.dart
```diff
@@ -76,6 +76,7 @@ class UserListView extends StatefulWidget {
   /// The query filters to use.
   /// You can query on any of the custom fields you've defined on the [Channel].
   /// You can also filter other built-in channel fields.
+  // TODO: Make it non-nullable in a future breaking release
```
imtoori, line 79: we already have breaking changes in the next release, don't we?

xsahil03x: not anymore, I deprecated it to deprecation instead
https://github.com/GetStream/stream-chat-flutter/pull/688

imtoori (approved):

Issue #674, opened 2021-09-10
Null check operator used on a null value


marcossevilla:
### Which packages are you using?

stream_chat_flutter

### On what platforms did you experience the issue?

iOS

### What version are you using?

package:stream_chat_flutter - ^2.0.0

### What happened?

Throws this error:

```
══╡ EXCEPTION CAUGHT BY SCHEDULER LIBRARY ╞══
The following _CastError was thrown during a scheduler callback:
Null check operator used on a null value

When the exception was thrown, this was the stack:
#0      ItemScrollController.jumpTo (package:scrollable_positioned_list/src/scrollable_positioned_list.dart:195:25)
#1      _MessageListViewState.didChangeDependencies.<anonymous closure> (package:stream_chat_flutter/src/message_list_view.dart:1202:28)
#2      SchedulerBinding._invokeFrameCallback (package:flutter/src/scheduler/binding.dart:1143:15)
#3      SchedulerBinding.handleDrawFrame (package:flutter/src/scheduler/binding.dart:1088:9)
#4      AutomatedTestWidgetsFlutterBinding.pump.<anonymous closure> (package:flutter_test/src/binding.dart:1006:9)
#7      TestAsyncUtils.guard (package:flutter_test/src/test_async_utils.dart:71:41)
#8      AutomatedTestWidgetsFlutterBinding.pump (package:flutter_test/src/binding.dart:993:27)
#9      WidgetTester.pumpWidget.<anonymous closure> (package:flutter_test/src/widget_tester.dart:554:22)
#12     TestAsyncUtils.guard (package:flutter_test/src/test_async_utils.dart:71:41)
#13     WidgetTester.pumpWidget (package:flutter_test/src/widget_tester.dart:551:27)
#14     main.<anonymous closure>.<anonymous closure> 
```

in this widget test:

```dart
late stream_chat_flutter.StreamChatClient client;
late stream_chat_flutter.Channel channel;

setUp(() {
  client = MockStreamChatClient();
  when(() => client.on(any(), any(), any(), any()))
      .thenAnswer((_) => const Stream.empty());
  channel = MockChannel();
  when(() => channel.initialized).thenAnswer((_) async => true);
  when(() => channel.on(any(), any(), any(), any())).thenAnswer(
    (_) => const Stream.empty(),
  );
  final channelClientState = MockChannelClientState();
  when(() => channel.state).thenReturn(channelClientState);
  when(() => channelClientState.threadsStream).thenAnswer(
[...]

PR #693, opened 2021-09-17
fix(ui): check scrollController is attached before calling jump.

deven98:
Fixes: #674


packages/stream_chat_flutter/lib/src/message_list_view.dart
```diff
@@ -1199,10 +1199,12 @@ class _MessageListViewState extends State<MessageListView> {
       initialAlignment = _initialAlignment;
 
       WidgetsBinding.instance!.addPostFrameCallback((timeStamp) {
-        _scrollController?.jumpTo(
-          index: initialIndex,
-          alignment: initialAlignment,
-        );
+        if (_scrollController?.isAttached ?? false) {
```
imtoori, line 1202:
```suggestion
        if (_scrollController?.isAttached == true) {
```

consider this change please, I think it's cleaner

imtoori (approved):

PR #696, opened 2021-09-20
fix(ui): fix MessageListView header and footer while reverse: false

imtoori:


packages/stream_chat_flutter/lib/src/message_list_view.dart
```diff
@@ -1199,12 +1213,10 @@ class _MessageListViewState extends State<MessageListView> {
       initialAlignment = _initialAlignment;
 
       WidgetsBinding.instance!.addPostFrameCallback((timeStamp) {
-        if (_scrollController?.isAttached == true) {
-          _scrollController?.jumpTo(
-            index: initialIndex,
-            alignment: initialAlignment,
-          );
-        }
```
xsahil03x, line 1207: ?

imtoori: I don't know 🤦🏼

HayesGordon (approved):
LGTM

PR #698, opened 2021-09-21
refactor(ui): use portal in messageinput overlays

imtoori:


packages/stream_chat_flutter/lib/src/mentions_overlay.dart
```diff
@@ -0,0 +1,101 @@
...
+  final MentionTileBuilder? mentionsTileBuilder;
+
+  @override
+  Widget build(BuildContext context) {
+    final _streamChatTheme = StreamChatTheme.of(context);
+
+    Future<List<Member>>? queryMembers;
+
+    if (query.isNotEmpty) {
+      queryMembers = channel
```
HayesGordon, line 38: We can't do this in the build. This will be triggered unnecessarily and also results in the current flickering. This should be a StateFullWidget that only calls the future when the query changes, use didUpdateWidget

packages/stream_chat_flutter/lib/src/mentions_overlay.dart
```diff
@@ -0,0 +1,101 @@
...
+                  .toList(),
+              const SizedBox(
+                height: 8,
+              ),
+            ],
+          ),
+        ),
+      ),
+    );
+    ;
```
HayesGordon, line 99: Unnecessary semi-colon ;

packages/stream_chat_flutter/lib/src/message_input.dart
```diff
@@ -472,7 +461,34 @@ class MessageInputState extends State<MessageInput> {
         child: child,
       );
     }
-    return child;
+
+    return PortalEntry(
```
HayesGordon, line 465: I think we should rather refactor this to be one PortalEntry, and then conditionally show the correct build function in the `portal`.

packages/stream_chat_flutter/lib/src/message_input.dart
```diff
@@ -878,10 +887,15 @@ class MessageInputState extends State<MessageInput> {
...
             .contains('@')) {
-      _mentionsOverlay = _buildMentionsOverlayEntry();
-      if (_mentionsOverlay != null) {
-        Overlay.of(context)!.insert(_mentionsOverlay!);
+      if (!_showMentionsOverlay) {
+        setState(() {
+          _showMentionsOverlay = true;
+        });
       }
+    } else if (_showMentionsOverlay) {
```
HayesGordon, line 895: This can just be an else

packages/stream_chat_flutter/lib/src/message_input.dart
```diff
@@ -898,146 +912,38 @@ class MessageInputState extends State<MessageInput> {
...
         _chosenCommand = matchedCommandsList[0];
         textEditingController.clear();
+
+        if (_showCommandsOverlay) {
+          setState(() {
+            _commandEnabled = true;
+            _showCommandsOverlay = false;
+          });
+        }
+      } else if (!_showCommandsOverlay) {
```
HayesGordon, line 922: Can be just an else

packages/stream_chat_flutter/lib/src/user_avatar.dart
```diff
@@ -66,6 +66,12 @@ class UserAvatar extends StatelessWidget {
     final placeholder =
         this.placeholder ?? streamChatTheme.placeholderUserImage;
 
+    final backupGradientAvatar = ClipRRect(
```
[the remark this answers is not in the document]
imtoori: you sure about this?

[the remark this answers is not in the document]
imtoori: if it's not in the tree it shouldn't be built, right?

packages/stream_chat_flutter/lib/src/mentions_overlay.dart
imtoori: yes, we're working on this

packages/stream_chat_flutter/lib/src/message_input.dart
imtoori: I don't think we can do this

packages/stream_chat/lib/src/client/channel.dart
```diff
@@ -1773,9 +1773,8 @@ class ChannelClientState {
   List<Message> get messages => _channelState.messages;
 
   /// Channel message list as a stream.
-  Stream<List<Message>> get messagesStream => channelStateStream
-      .map((cs) => cs.messages)
-      .distinct(const ListEquality().equals);
```
xsahil03x, line 1778: @imtoori why we did this change?

imtoori: there was a problem with ephemeral messages not updating related to this

packages/stream_chat_flutter/lib/src/message_input.dart
```diff
@@ -1261,300 +1152,78 @@ class MessageInputState extends State<MessageInput> {
...
+    var tileBuilder = widget.userMentionsTileBuilder;
+    if (tileBuilder == null && widget.mentionsTileBuilder != null) {
+      tileBuilder = (context, user) {
+        final member = Member(
+          user: user,
+          userId: user.id,
+          createdAt: user.createdAt,
+          updatedAt: user.updatedAt,
+        );
+        return widget.mentionsTileBuilder!.call(context, member);
```
deven98, line 1177:
```suggestion
        return widget.mentionsTileBuilder!(context, member);
```

deven98 (approved):
LGTM

PR #699, opened 2021-09-21
feat(ui, core): added callback for modifying channel lists

deven98:


packages/stream_chat_flutter_core/lib/src/channel_list_core.dart
```diff
@@ -192,15 +208,16 @@ class ChannelListCoreState extends State<ChannelListCore> {
...
-            EventType.notificationMessageNew,
-            EventType.channelVisible,
-          )
-          .listen((event) => loadData());
+      _subscription =
+          client.on().where((e) => _eventMap.containsKey(e.type)).listen((e) {
+        final list = _eventMap[e.type]?.call(_channelsBloc.channels, e);
+
+        if (list != null) {
+          _channelsBloc.channelsController.add(list);
```
imtoori, line 218: I think we could make `channelsController` private adding a method to do this

imtoori (changes requested):
This looks good, just a couple of nitpicks
Also please let's add some tests

PR #702, opened 2021-09-21
docs(doc): add versioning policy

HayesGordon:


HayesGordon:
If this looks fine, then we can add it to CONTRIBUTION and our Docs.

CONTRIBUTING.md
```diff
@@ -177,6 +177,26 @@ Develop is merged into master after the team performs various automated and QA t
 
 ---
 
+# Versioning Policy
+
+All of the Stream Chat packages follow [semantic versioning (semver)](https://semver.org/).
+
+That means that with a version number x.y.z (major.minor.patch):
+
+- When releasing critical bug fixes, we make a patch release by changing the z number (ex: 3.6.2 to 3.6.3).
```
deven98, line 186: Small doubt, but do we want to use the word 'critical' here?

xsahil03x: https://semver.org/#spec-item-6
```
Patch version Z (x.y.Z | x > 0) MUST be incremented if only backwards compatible bug fixes are introduced. A bug fix is defined as an internal change that fixes incorrect behavior.
```

HayesGordon: Yeah we can change it. I'll modify it to what @xsahil03x suggested

PR #704, opened 2021-09-22
chore(ui,core,llc,persistence,localization): remove analysis warnings due to flutter 2.5

imtoori:


HayesGordon (approved):
I updated some extra docs to use `limit` instead of `paginationParams`. The rest LGTM

PR #708, opened 2021-09-27
feat(llc): added not exists filter

deven98:


packages/stream_chat/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## Upcoming
+
+✅ Added
+
+- `Filter.notExists` now replaces `Filter.exists(key, exists: false)`.
```
imtoori, line 5:
```suggestion
- Added `Filter.notExists`.
```

packages/stream_chat/lib/src/core/models/filter.dart
```diff
@@ -40,9 +40,12 @@ enum FilterOperator {
   /// Matches values with the specified prefix.
   autoComplete,
 
-  /// Matches values that exist/don't exist based on the specified boolean value.
+  /// Matches values that exist.
```
imtoori, line 43:
```suggestion
  /// Matches values that exist/don't exist based on the specified boolean value.
```

packages/stream_chat/lib/src/core/models/filter.dart
```diff
@@ -163,10 +166,13 @@ class Filter extends Equatable {
   factory Filter.autoComplete(String key, String text) =>
       Filter._(operator: FilterOperator.autoComplete, key: key, value: text);
 
-  /// Matches values that exist/don't exist based on the specified boolean value.
+  /// Matches values that exist.
```
imtoori, line 169:
```suggestion
  /// Matches values that exist/don't exist based on the specified boolean value.
```

imtoori (changes requested):
a couple of changes
also add it to the docs please

packages/stream_chat/lib/src/core/models/filter.dart
```diff
@@ -43,6 +43,9 @@ enum FilterOperator {
   /// Matches values that exist/don't exist based on the specified boolean value.
   exists,
 
+  /// Matches values that don't exist.
+  notExists,
+
```
xsahil03x, line 48: we don't really need this

deven98: Yeah fair enough, it was done in another way before, it changed to make it non breaking

PR #709, opened 2021-09-27
fix(ui): thread reply scrolling main list

### LINKED ISSUE #687: Users lose their place in the conversation after replying in threads.

stream_chat_flutter

iOS, Android

2.2.1

### What happened?

When replying in a thread from a parent MessageListView, sending a reply resets the scroll position of the parent MessageListView to the bottom, so users lose their place in the chat after replying. I'm guessing a fix for this would be to not automatically scroll to the bottom when a thread reply is added.

### Steps to reproduce

```bash
1. Override threadBuilder with your own ThreadPage implementation
2. Find a chat with a bunch of messages in it and scroll up to the top
3. Long press one of the first messages and tap on Thread Reply
4. Add a reply to the thread in ThreadPage
5. Exit ThreadPage and notice how the chat is now at the bottom and you've lost your place in the chat message list.
```

_No response_

### Relevant log output

_No response_

### Flutter analyze output

_No response_

### Flutter doctor output

_No response_

### Code of Conduct

imtoori:
fixes #687

### DIFF
--- a/packages/stream_chat_flutter/CHANGELOG.md
+++ b/packages/stream_chat_flutter/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Upcoming
+
+🐞 Fixed
+
+-[[#687]](https://github.com/GetStream/stream-chat-flutter/issues/687): Fix Users losing their place in the conversation after replying in threads.
+
 ## 3.0.0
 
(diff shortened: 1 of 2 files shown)


xsahil03x (commented):
LGTM

xsahil03x (approved):

PR #722, opened 2021-10-05
chore(llc,core,ui): 3.1.0

imtoori:


packages/stream_chat_flutter/pubspec.yaml
```diff
@@ -17,15 +17,15 @@ dependencies:
   diacritic: ^0.1.3
   dio: ^4.0.0
   ezanimation: ^0.5.0
-  file_picker: ^3.0.1
+  file_picker: ^4.1.3
```
HayesGordon, line 20: Maybe mention in the Changelog that this is updated, as this is a major upgrade

packages/stream_chat/pubspec.yaml
```diff
@@ -28,8 +28,8 @@ dependencies:
 dev_dependencies:
   build_runner: ^2.0.1
   freezed: ^0.14.1+3
-  json_serializable: ^4.1.0
+  json_serializable: ^5.0.2
   mocktail: ^0.1.1
   # Remove fixed version once the below issue is fixed
   # https://github.com/flutter/flutter/issues/91059
```
xsahil03x, line 34: we can remove these lines

imtoori:
```suggestion
```

PR #725, opened 2021-10-06
fix(ui): fix message list initial index and alignment

imtoori:


packages/stream_chat_flutter/lib/src/message_list_view.dart
```diff
@@ -305,10 +305,18 @@ class _MessageListViewState extends State<MessageListView> {
...
+              ? streamChannel!.channel.state?.threads[widget.parentMessage!.id]
+              : streamChannel!.channel.state?.messages)
+          ?.where(widget.messageFilter ?? defaultFilter(context))
+          .toList();
+
+      final totalMessages = messages?.length ?? 0;
+      final messageIndex = messages
+              ?.indexWhere((e) => e.id == streamChannel!.initialMessageId) ??
+          -1;
+      print('messageIndex: $messageIndex');
```
deven98, line 318:
```suggestion
```

packages/stream_chat_flutter/lib/src/message_list_view.dart
```diff
@@ -1218,10 +1226,26 @@ class _MessageListViewState extends State<MessageListView> {
     if (newStreamChannel != streamChannel) {
       streamChannel = newStreamChannel;
       _messageNewListener?.cancel();
-      initialIndex = _initialIndex;
-      initialAlignment = _initialAlignment;
 
-      WidgetsBinding.instance!.addPostFrameCallback((timeStamp) {
+      WidgetsBinding.instance!.addPostFrameCallback((timeStamp) async {
```
deven98, line 1230: Quick q: Why does this need to be async?

imtoori: it does not 
```suggestion
      WidgetsBinding.instance!.addPostFrameCallback((timeStamp) {
```

PR #735, opened 2021-10-13
feat(localization): add Nepali language support

shivamyadavnp:
Added support for the Nepali language which is one of the most used languages in Nepal and somehow it is even in practice with huge numbers in India.


imtoori:
closing this
feel free to reopen it when it's ready

PR #742, opened 2021-10-19
feat(ui, core): Added MessageInputController and general refactor

deven98:


packages/stream_chat/lib/src/core/models/message.dart
```diff
@@ -72,16 +72,19 @@ class Message extends Equatable {
     this.pinnedBy,
     this.extraData = const {},
     this.deletedAt,
-    this.status = MessageSendingStatus.sent,
+    this.status = MessageSendingStatus.sending,
```
HayesGordon, line 75: Why was this changed? Note that the `fromJson` is setting `MessageSendingStatus.sent`

imtoori: because we need that in order to have the right behaviour in the message input
a new created message has a sending status, while if you get it from backend we set it as send

PR #750, opened 2021-10-26
fix(llc): fix channel unread count

imtoori:


developerjamiu (commented):
This does not fix the issue at all. In fact it sets the unread count as zero all the time (after every message).

imtoori:
@developerjamiu can you be more specific?
It fixes the issue, I don't see it when I run the sample app

developerjamiu:
When a new message is sent to a room by a user (whether in the foreground or background), the unreadCountStream keeps returning 0 for the other users.

imtoori:
there are many reasons for that to happen
Are the other users member of the channel?

developerjamiu:
Yes, there are.

imtoori:
I'm quite sure there's an explanation :/ 
Can you try our sample app and see if it works there?
or even this example? https://github.com/GetStream/stream-chat-flutter/blob/develop/packages/stream_chat_flutter/example/lib/tutorial_part_4.dart
it's working fine for me, so we need to confirm you're not doing anything wrong

developerjamiu:
Alright, thanks!

PR #771, opened 2021-11-15
feat(ui): added customisation options for attachment actions modal and related widgets

### LINKED ISSUE #766: AttachmentActionsModal should have configurable actions

stream_chat_flutter

_No response_

We don't want to show the 'Show in Chat' or 'Reply' actions that are hardcoded into the  `AttachmentActionsModal` when the user has tapped the image from a channel page.  I don't understand the purpose of the 'show in chat' action when you have just tapped the image from the current chat channel and can simply close the gallery view.  I understand its utility in a search context, but I find the 'Show in Chat' option confusing when I tapped it from the channel.  

Additionally, when tapped from the channel page, I'm not sure of a simple solution for `onShowMessage` in order to get the 'Show in Chat' option to just close the attachment detail view.  I ended up resorting to a less than ideal hack/workaround using route names, since in this case I know they came from the channel page.
```
  onShowMessage: (message, channel) {
    Navigator.of(context).popUntil((route) =>
        route.settings?.name ==
        RouteNameBuilder.chatChannelByCid(channel.cid));
  },
```

And the `Reply` option doesn't make sense for us at all because we don't support quoted replies or threads right now, so it would be nice to be able to remove it!

Ability to configure the actions generated by the `AttachmentActionsModal` class.

_No response_

### Additional context

Sorry if I'm missing something in the documentation about how to configure this, or just generally not understanding the point of these actions in the context described above.  Loving stream and stream flutter 👍 

### Code of Conduct

deven98:
Fixes #766

### DIFF
--- a/.github/workflows/dart_code_metrics.yaml
+++ b/.github/workflows/dart_code_metrics.yaml
@@ -33,7 +33,7 @@ jobs:
           flutter-version: ${{ env.flutter_version }}
 
       - name: "Install Tools"
-        run: flutter pub global activate melos 1.0.0-dev.6
+        run: flutter pub global activate melos 1.0.0-dev.10
       - name: "Bootstrap Workspace"
         run: melos bootstrap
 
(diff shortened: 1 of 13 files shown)


imtoori (approved):
LGTM 
there is some room of improvement regarding the FullScreenMedia, but that's probably gonna be breaking so let's do it in the future

PR #775, opened 2021-11-17
feat(ui): stream channel listview

xsahil03x:


packages/stream_chat_flutter_core/lib/src/channels_bloc.dart
```diff
@@ -16,6 +16,10 @@ import 'package:stream_chat_flutter_core/src/stream_controller_extension.dart';
 /// using Flutter's [BuildContext].
 ///
 /// API docs: https://getstream.io/chat/docs/flutter-dart/query_channels/
+@Deprecated(
+  "'ChannelsBloc' is deprecated and shouldn't be used. "
+  "Please use 'StreamChannelListView' instead.",
```
xsahil03x, line 21:
```suggestion
  "Please use 'StreamChannelListController' instead.",
```

imtoori: makes sense

PR #785, opened 2021-11-24
fix(ui): listen for read events in the message widget

imtoori:


packages/stream_chat_flutter/lib/src/message_widget.dart
```diff
@@ -558,8 +564,6 @@ class _MessageWidgetState extends State<MessageWidget>
 
   bool get showTimeStamp => widget.showTimestamp;
 
-  bool get isMessageRead => widget.readList?.isNotEmpty == true;
```
deven98, line 561: Wouldn't this be breaking?

packages/stream_chat_flutter/lib/src/message_widget.dart
```diff
@@ -97,14 +97,20 @@ class MessageWidget extends StatefulWidget {
...
     this.attachmentPadding = EdgeInsets.zero,
-    this.allRead = false,
+    @Deprecated('''
+    allRead is now deprecated and it will be removed in future releases. 
+    The MessageWidget now listens for read events on its own.
+    ''') this.allRead = false,
+    @Deprecated('''
+    readList is now deprecated and it will be removed in future releases. 
+    The MessageWidget now listens for read events on its own.
+    ''') this.readList,
```
deven98, line 113: The PR looks good. The only issue I see is this: it is breaking without being literal code breaking - we do not use the values of readList after taking it in as a parameter. To make it non breaking, we would likely make it work as it is in default conditions (this is obviously problematic since the current implementation is buggy) or remove the parameters altogether to make it breaking but remove effectively dead parameters.

deven98 (approved):
LGTM

PR #786, opened 2021-11-24
fix(ui): do not use `withData: true` property

### LINKED ISSUE #760: FilePicker is called with withData=true, crashing on big files

stream_chat_flutter

Android

3.1.1

### What happened?

When picking large files (videos with a >500 MB on my device, but this will vary), the app crashes because default `MessageInput` widget uses the following code to get a file from the FilePicker plugin:
```
  final res = await FilePicker.platform.pickFiles(
        type: type,
        withData: true,
      );
```
According to issues in the FilePicker github, the `withData` loads the full file in the phone memory, which is I believe not necessary and setting it to `false` seems to work fine too: https://github.com/miguelpruivo/flutter_file_picker/issues/392

### Steps to reproduce

```bash
1. Use stream chat flutter MessageInput Widget
2. Click Attachments and select large file 
3. Phone crashes with OutOfMemory exception
```

_No response_

### Relevant log output

_No response_

### Flutter analyze output

_No response_

### Flutter doctor output

_No response_

### Code of Conduct

imtoori:
fixes #760

### DIFF
--- a/packages/stream_chat_flutter/CHANGELOG.md
+++ b/packages/stream_chat_flutter/CHANGELOG.md
@@ -12,7 +12,8 @@
 - Fixed `MessageWidget` null errors associated with `channel.memberCount`.
 - Fixed adding attachments on web.
 - [[#767]](https://github.com/GetStream/stream-chat-flutter/issues/767): Fix `MessageInput` focus behaviour when sending messages.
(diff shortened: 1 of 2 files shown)


deven98 (approved):
LGTM

PR #793, opened 2021-11-26
fix(llc): ChannelEvent.membersCount default to 0

imtoori:


packages/stream_chat/lib/src/client/client.dart
```diff
@@ -1432,6 +1432,8 @@ class ClientState {
         .listen((Event event) async {
       final eventChannel = event.channel!;
       await _client.chatPersistenceClient?.deleteChannels([eventChannel.cid]);
+      final channel = channels[eventChannel.cid];
```
deven98, line 1435:
```suggestion
      channels[eventChannel.cid]?.dispose();
```

packages/stream_chat/lib/src/client/client.dart
```diff
@@ -1432,6 +1432,8 @@ class ClientState {
...
+      channel?.dispose();
```
deven98, line 1436:
```suggestion
```

deven98 (approved):
LGTM

PR #795, opened 2021-11-29
chore(llc,core,ui): 3.3.0

imtoori:


deven98:
Screenshot 2021-11-29 at 7 51 00 PM

We likely shouldn't be supplying these parameters since internally they aren't used either way (even if the user provides custom properties which should ideally override them). This will also fix analysis to some extent.

imtoori:
wouldn't it be breaking in that case? @deven98

deven98:
> wouldn't it be breaking in that case? @deven98

I'm not saying remove the parameters, I'm saying don't pass them on since we don't use them internally in the widget.

imtoori:
yeah that's right, I just pushed it @deven98

deven98:
> yeah that's right, I just pushed it @deven98

I think we can remove this the second place as well. One sec.

PR #891, opened 2022-02-08
fix(llc): decrease parent message reply count only once

bstolinski:
# Submit a pull request

## CLA

### Issue
When a message in a thread is deleted, the reply counter of the parent message decreases by two instead of one.

The video shows how the second user deletes `message 3` and `message 2` one by one and how this affects the reply counter in the parent's message.

https://user-images.githubusercontent.com/53336850/153065893-8086a66c-f0f9-4c02-b2c1-257353408e31.mov

### Solution
Every time a message is deleted, the listener from `_listenMessageUpdated` is called and `event` already has a message in it with a reduced reply counter:

```
void _listenMessageUpdated() {
  _subscriptions.add(_channel
      .on(
    EventType.messageUpdated,
    EventType.reactionUpdated,
  )
      .listen((event) {
    final oldMessage =
        messages.firstWhereOrNull((it) => it.id == event.message?.id);
    final message = event.message!.copyWith(
      ownReactions: oldMessage?.ownReactions,
    );
    addMessage(message);
    if (message.pinned) {
      _channelState = _channelState.copyWith(
        pinnedMessages: [
          ..._channelState.pinnedMessages,
          message,
        ],
      );
    }
  }));
}
```

Then the listener from `_listenMessageDeleted` calls `removeMessage` (in this case `hardDelete` is always `true`):

```
void _listenMessageDeleted() {
  _subscriptions.add(_channel.on(EventType.messageDeleted).listen((event) {
    final message = event.message!;
    if (event.hardDelete == true) {
      removeMessage(message, hardDelete: true);
    } else {
      addMessage(message);
    }
  }));
}
```

and `removeMessage` finds the parent message of the deleted message and decreases the reply counter one more time:

```
/// Remove a [message] from this [channelState].
void removeMessage(Message message, {bool hardDelete = false}) {
  final parentId = message.parentId;
  // i.e. it's a thread message
  // 1. Remove the thread message
  // 2. Reduce total reply count of parent message
  if (parentId != null) {
    final allMessages = [...messages];
    final parentMessage = allMessages.firstWhereOrNull(
      (it) => it.id == parentId,
    );
    // return if message not available in the memory
    if (parentMessage == null) return;
    final replyCount = parentMessage.replyCount;
    // return if reply count is null or zero
    if (replyCount == null || replyCount == 0) return;
    addMessage(parentMessage.copyWith(replyCount: replyCount - 1));
    updateThreadInfo(
      parentId,
      threads[parentId]!
        ..removeWhere(
          (e) => e.id == message.id,
        ),
    );
  } else {
    // Remove regular message
    final allMessages = [...messages];
    if (hardDelete) {
      allMessages.removeWhere((e) => e.id == message.id);
      _channelState = _channelState.copyWith(messages: allMessages);
    } else if (allMessages.remove(message)) {
      _channelState = _channelState.copyWith(messages: allMessages);
    }
  }
}
```

`removeMessage` should decrease the reply counter only when executing the action on messages (`sendAction`) returns the message from the server as `null`. This solution works for me, but I'm not 100% sure if it should be fixed this way.


xsahil03x:
Hey @bstolinski, can you test and merge this PR if it fixes your issue?
https://github.com/squaddy/stream-chat-flutter/pull/2

xsahil03x:
Hey @bstolinski @TomBeckett can you please also update the changelog?

bstolinski:
Hi @xsahil03x, in your solution `removeMessage` calls` updateThreadInfo` giving the messages without the one that was removed and that's correct, but inside `updateThreadInfo` algorithm gets the old messages from the thread (with the one that was removed) and adds to the given messages.

```
/// Update threads with updated information about messages.
void updateThreadInfo(String parentId, List<Message> messages) {
  final newThreads = Map<String, List<Message>>.from(threads);
  if (newThreads.containsKey(parentId)) {
    newThreads[parentId] = [
      ...messages,
      ...newThreads[parentId]!.where(
        (newMessage) => !messages.any((m) => m.id == newMessage.id),
      ),
    ]..sort(_sortByCreatedAt);
  } else {
    newThreads[parentId] = messages;
  }
  _threads = newThreads;
}
```

xsahil03x:
Hey @bstolinski , I will take a look and merge it soon 👍🏼

xsahil03x:
Hey @bstolinski , The changes LGTM 👍🏼 I just added some more cases, can you look at this PR and merge it?
https://github.com/squaddy/stream-chat-flutter/pull/3  

I will merge this PR afterward.

Thanks,
Sahil

bstolinski:
Hi @xsahil03x, looks good 🚀 PR merged, thanks!

xsahil03x (dismissed):
LGTM

xsahil03x (approved):
LGTM

PR #911, opened 2022-02-21
perf(ui): improve full screen media animation

HayesGordon:
- Reduce build calls
- Add intervals/curves to improve the animation feel
- Expose controllers


imtoori (approved):
LGTM

PR #912, opened 2022-02-21
fix(ui): sample app version bump

sachaarbonel:
# Submit a pull request

## CLA

### DIFF
--- a/packages/stream_chat_flutter/example/pubspec.yaml
+++ b/packages/stream_chat_flutter/example/pubspec.yaml
@@ -27,9 +27,12 @@ dependencies:
   cupertino_icons: ^1.0.3
   flutter:
     sdk: flutter
-  stream_chat_flutter: ^3.4.0
-  stream_chat_localizations: ^2.1.0
-  stream_chat_persistence: ^3.1.0
+  stream_chat_flutter: 
+    path: ../
+  stream_chat_localizations: 
+    path: ../../stream_chat_localizations
+  stream_chat_persistence: 
+    path: ../../stream_chat_persistence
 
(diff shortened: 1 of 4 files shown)


packages/stream_chat_flutter/example/pubspec.yaml
```diff
@@ -27,9 +27,9 @@ dependencies:
   cupertino_icons: ^1.0.3
   flutter:
     sdk: flutter
-  stream_chat_flutter: ^2.2.1
-  stream_chat_localizations: ^1.1.0
-  stream_chat_persistence: ^2.2.0
+  stream_chat_flutter: ^3.4.0
+  stream_chat_localizations: ^2.1.0
+  stream_chat_persistence: ^3.1.0
```
imtoori, line 32: @sachaarbonel isn't it better to use the local version from path?

imtoori (changes requested):

packages/stream_chat_flutter/example/pubspec.yaml
xsahil03x: isn't really necessary as we run them after `melos bs` which does the same thing.

imtoori: yeah but then once we push a new release we need to update this as well and it's very easy to forget

xsahil03x: fine then 👍🏼

imtoori: @sachaarbonel can you update this?

PR #918, opened 2022-02-28
refactor(ui): move portal location to message list layer

HayesGordon:
Instead of creating a portal for each message widget, create a portal for each message list. It cannot be moved higher in the widget tree, otherwise it will result in the reactions overlaying other components (for example, the AppBarr)


imtoori (changes requested):
image
I just want to highlight that this pr causes this
In my opinion this shouldn't happen
What do you think?

HayesGordon:
Agreed, closed this for now until we get a better solution.

PR #938, opened 2022-03-01
feat(ui): Add Channel Header customization

S-ecki:
# Submit a pull request

## CLA

- added properties `elevation` and `centerTitle` on `ChannelHeader`
- previous values are taken as defaults if properties are null (so change does not affect existing designs)
- this was done to enable users to style the `ChannelHeader` like an `AppBar` (e.g. elevation = 4 and centerTitle = false by default)


imtoori (changes requested):
hey @S-ecki thanks for the PR
I polished it a bit

I think `centerTitle` does not do anything right now, can you confirm this? Maybe we can remove it or find a way of making it effective.

Also, I think we should do the same for `ThreadHeader` and `ChannelListHeader`
What do you think?

S-ecki:
hey @imtoori, thanks for the answer!

`centerTitle` does indeed change something for me. I looked further into the issue and discovered that the Flutter `AppBar` sets the default for `centerTitle` based on the Platform. You can find the logic in Flutter´s `app_bar.dart` file on line 714 (`_getEffectiveCenterTitle()`).
TLDR: on Android, `centerTitle` is false. on iOS it is true unless there are 2 or more `actions` in the `AppBar`.

I would try to align `ChannelHeader`s `centerTitle` with Flutter `AppBar` `centerTitle`. 

The reason why I want it to be consistent is following: I have a mobile App with multiple Tabs, where each Tab has a AppBar with the same design. One of those Tabs contains Stream Chat and I want the ChannelHeader to be used as the AppBar. Thus, it needs to be styled exactly the same.

I did not use `ThreadHeader` or `ChannelListHeader` yet, but I would also keep them consistent with `AppBar` 👍

S-ecki:
hey @imtoori!

I tested around with the `centerTitle` property and came to the conclusion that the best way to handle it is to let the actual `AppBar` handle it. What I mean by that is a user can pass in a `centerTitle` property, which will be set. If no `centerTitle` is passed in, we pass `null` to the Flutter `AppBar`. By that, the value of `centerTitle` gets decided based on `_getEffectiveCenterTitle()` (like mentioned in my last comment).
This way, we can ensure that Stream Chat centerTitle properties are always aligned with Flutter AppBar.

I also added elevation and centerTitle to `ThreadHeader` and `ChannelListHeader`.

S-ecki:
Hey @imtoori 

are there any updates on this issue? Any improvements I could make?

imtoori:
hey @S-ecki I'm reviewing it now, sorry about the delay but I was focused on another task

imtoori:
@S-ecki the only problem I see is that I don't see any visual change after changing the `centerTitle` property to true/false
I think that's because we don't use just a `Text`, but a more complex widget that messes up with that property
The title stays always centered using the default configuration changing only the `centerTitle` property
Can you confirm this? I'm pretty sure we can implement a workaround for this

S-ecki:
Hey @imtoori, I unfortunately can confirm that the `centerTitle` property does not change anything visually, sorry!
Do you have an idea of how we can work around that?

imtoori:
I'm working on #995 at the moment and it's quite high in priority, I'll see what I can do later today hopefully

S-ecki:
sorry to bother you again @imtoori , any updates in sight? :)

imtoori:
hey @S-ecki not yet, I'm sorry, I'll work on this in the following days

peter-gy:
Hey @imtoori, I also find myself in need to have this issue resolved. If you have a vague idea about the approach to be taken to solve the problem, I would be glad to contribute!

packages/stream_chat_flutter/lib/src/channel_header.dart
```diff
@@ -61,9 +61,11 @@ class ChannelHeader extends StatelessWidget implements PreferredSizeWidget {
     this.showConnectionStateTile = false,
     this.title,
     this.subtitle,
+    this.centerTitle,
     this.leading,
     this.actions,
     this.backgroundColor,
+    this.elevation,
```
xsahil03x, line 68:
```dart
this.elevation = 1
```

imtoori:
```suggestion
    this.elevation = 1,
```

packages/stream_chat_flutter/lib/src/thread_header.dart
```diff
@@ -179,4 +191,21 @@ class ThreadHeader extends StatelessWidget implements PreferredSizeWidget {
 
   @override
   final Size preferredSize;
+
+  bool _getEffectiveCenterTitle(ThemeData theme) {
```
ayush221b, line 195: Since we are using the same function in `ChannelHeader` as well as `ThreadHeader` do you think it would be a good idea to make it a global helper function and use that in both places, instead of duplicating the code?

S-ecki:
thanks a lot for merging @imtoori! Is there a timeline for release 3.6.2?

imtoori:
We're currently focused on releasing v4 today, 3.7 will come after that

Issue #1071, opened 2022-04-11
Save File in full_screen_media freezes up the app, shows random progress and throws range error


bhanu-under25:
### Which packages are you using?

stream_chat, stream_chat_flutter

### On what platforms did you experience the issue?

iOS, Android

### What version are you using?

stream_chat_flutter: ^3.5.1

### What happened?

Full Screen media shows range error and shows random progress indicator.

Below are the steps to reproduce and it affects not only on the sample app but all other full screen media implementation.

I hope following video gives enough steps to reproduce in sample app code.

 - This a gif showing the range error.
 
![range error](https://user-images.githubusercontent.com/92363650/162725847-35aa3884-37dc-4222-93d8-6026c1c99be1.gif)

- This shows the random progress in dialog box, it freezes the app and dialog box is never dismissed.

![random](https://user-images.githubusercontent.com/92363650/162725963-2decb3e0-419b-4e67-8965-4478806bbc25.gif)

Let me know if you need any more information to reproduce this.

### Steps to reproduce

```bash
1. Go to '...'
2. Click on '...'
3. Scroll down to '...'
...
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ayush221b:
Hey @bhanu-under25 thank you for bringing this to our attention, I am able to reproduce the RangeError, working on a fix. I will update here once done.

bhanu-under25:
@ayush221b

You can reproduce the save file error too in flutter sample app.
- It shows download random progress.
- Doesn't get dismissed automatically.
- And because the showDialog used is non dismissable we can't navigate back.

Let me know if you need any more information regarding this.

Thank You.

bhanu-under25:
Thanks for solving.

PR #1076, opened 2022-04-13
fix(ui): Attachment actions on full screen were broken

ayush221b:
# Submit a pull request

## CLA

Solves #1071 

The FullScreen implementation and the adjoining widgets had some flaws, because of which we were getting some random unexpected behaviour. I have introduced the concept of StreamAttachmentPackage, that can be used when each individual attachment needs to be mapped to its message, I have also added a new progress UI and changed the way we track download completion for attachments. I have also made the necessary changes to the tests.


ayush221b:
1. This also removes the `prefer_expression_function_bodies` rule from our analysis options.
2. Also, this exports the file `src/multi_overlay.dart` because that file houses the class `OverlayOptions` which we require when a developer has to pass `customOverlays` to the MessageListView.

ibhanu:
fyi, it brings breaking changes!

imtoori:
@ibhanu that's why it's in v4

ibhanu:
how would it work for backward compatibility?
because old version have this issue.

imtoori:
it doesn't work :( this was the only way to fix it and we pushed it in v4 since it's breaking
I hope that's not a problem, the upgrade is not hard https://getstream.io/chat/docs/sdk/flutter/guides/migration_guide_4_0/

PR #1123, opened 2022-05-05
feat(ui): defaultGroupImage builder

### LINKED ISSUE #1095: Ability to override GroupAvatar

stream_chat_flutter

iOS, Android

It doesn't seem possible to replace GroupAvatars used throughout stream_chat_flutter. We've been able to customise single-user avatars by implementing `defaultUserImage` in `StreamChatTheme`. However, the way this is used for GroupAvatar just stacks the avatars side by side with some overlap.

We would like to be able to replace GroupAvatars with our own widget so that we can customise the look and feel of Stream Chat to match our app.

One way to do this would be to have a builder for group avatars in the same way the theme has for default avatars. Or a builder for replacing `StreamChannelAvatar` and we can decide what to do based on the number of users.

I _think_ right now it's possible to replace the avatars in different places. For example, putting our own widgets into `StreamChannelHeader.actions`. However, it would mean having to duplicate this work everywhere where Stream displays an avatar. It would be much cleaner to override it in the ThemeData.

### Additional context

_No response_

### Code of Conduct

GroovinChip:
Closes #1095 

## CLA

### DIFF
--- a/packages/stream_chat_flutter/CHANGELOG.md
+++ b/packages/stream_chat_flutter/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Upcoming
 
+✅ Added
+- [[#1095]](https://github.com/GetStream/stream-chat-flutter/issues/1095): Add the ability to override the default `StreamGroupAvatar`
+
 🐞 Fixed
 
 - Fixed attachment picker ui.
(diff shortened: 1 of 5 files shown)


imtoori (changes requested):
I think this is not the right approach.
`defaultUserImage` is used when the user image is not available, so that's not the right way to customize the user avatar, it's more a workaround they're doing.
I think they should just replace the user avatar using their own implementation; there are ways of doing this in basically every widget. The same is valid for GroupAvatar.
I think we should help them doing that.

GroovinChip:
My understanding of their request is that they would like to be able to globally set a _default_ group avatar that they then have the option of overriding in various places, because they do not want to have to set a _default_ group avatar more than once. I think they know that they can set group avatars via the widgets; their request to be able to set their default _once_. Am I misunderstanding their request?

GroovinChip:
Per offline conversation with @imtoori I am closing this PR in favor of a different approach.

PR #1128, opened 2022-05-06
feat(ui): `StreamChatConfiguration` class

### LINKED ISSUE #1125: Create a `StreamChatConfiguration` class to hold various global UI defaults
Per conversation with @imtoori, we should remove non-theme related UI configuration options from `StreamChatThemeData` and put them into their own class, `StreamChatConfiguration`, which will be a property in the `StreamChat` widget just like `StreamChatThemeData`.

`StreamChatConfiguration` should contain a method that will look up the widget tree to find the nearest instance, like InheritedWidgets do.

The following properties would be moved from `StreamChatThemeData` to `StreamChatConfiguration`:
* `final Widget Function(BuildContext, User) defaultUserImage`
* `final Widget Function(BuildContext, User)? placeholderUserImage`
* `final List<StreamReactionIcon> reactionIcons`

This class would be appropriate for any future global UI configuration options that are not strictly theme related.

GroovinChip:
# Submit a pull request

This PR contains a new `StreamChatConfiguration` class for use in `StreamChat` widgets.

Closes #1125 (see for more details)

## CLA


xsahil03x:
@GroovinChip can we also convert it into an inherited widget similar to the theme instead of accessing it via `StreamChat` widget.

imtoori:
@xsahil03x  I think it's better to leave it like this, attached to the StreamChat widget
Also, @GroovinChip, can you change the base branch to the desktop one? Or maybe create a v5 branch?

xsahil03x:
> > @xsahil03x I think it's better to leave it like this, attached to the StreamChat widget Also, @GroovinChip, can you change the base branch to the desktop one? Or maybe create a v5 branch?
> 
> I agree that this should not be an `InheritedWidget`- it would complicate usage.

@GroovinChip the reason I want it to be an inherited widget is that its use-cases are similar to the theme class. What if a user only wants to change the configuration for a sub widget tree? It's not possible atm with the current implementation.

imtoori:
Let's add also a parameter to enforce unique reactions

packages/stream_chat_flutter/lib/src/stream_chat_configuration.dart
```diff
@@ -0,0 +1,221 @@
+import 'package:flutter/material.dart';
+import 'package:stream_chat_flutter/stream_chat_flutter.dart';
+
+/// {@template streamChatConfigurationProvider}
+/// Inherited widget providing the [StreamChatConfiguration] to the widget tree
+/// {@endtemplate}
+class StreamChatConfigurationProvider extends InheritedWidget {
```
xsahil03x, line 7: I would rename this to `StreamChatConfiguration` and the data to `StreamChatConfigurationData`

imtoori: makes sense

packages/stream_chat_flutter/lib/src/stream_chat_configuration.dart
```diff
@@ -0,0 +1,222 @@
...
+  final Widget Function(BuildContext, User)? defaultUserImage;
+
+  /// The widget that will be built when the user image is loading.
+  final Widget Function(BuildContext, User)? placeholderUserImage;
+
+  /// Assets used for rendering reactions.
+  final List<StreamReactionIcon>? reactionIcons;
+
+  /// Whether a new reaction should replace the existing one.
+  final bool? enforceUniqueReactions;
```
xsahil03x, line 158: this can be non-nullable

packages/stream_chat_flutter/lib/src/stream_chat.dart
```diff
@@ -35,6 +35,7 @@ class StreamChat extends StatefulWidget {
     required this.client,
     required this.child,
     this.streamChatThemeData,
+    this.config,
```
xsahil03x, line 38:
```suggestion
    this.streamChatConfigData,
```

Issue #1134, opened 2022-05-11
add channel.membership support


imtoori:
channel.membership is the `Member` object related to the logged-in user

- add support for `channel.membership`
- add support for `channel.membershipStream` (if it makes sense)

PR #1165, opened 2022-05-19
feat(llc): add support for channel.membership

xsahil03x:
Fixes: #1134

### DIFF
--- a/packages/stream_chat/lib/src/core/models/channel_model.dart
+++ b/packages/stream_chat/lib/src/core/models/channel_model.dart
@@ -26,7 +26,6 @@ class ChannelModel {
     this.extraData = const {},
     this.team,
     this.cooldown = 0,
-    this.membership,
   })  : assert(
           (cid != null && cid.contains(':')) || (id != null && type != null),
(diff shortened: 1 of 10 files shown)


packages/stream_chat/lib/src/core/models/channel_model.dart
```diff
@@ -1,5 +1,6 @@
 import 'package:json_annotation/json_annotation.dart';
 import 'package:stream_chat/src/core/models/channel_config.dart';
+import 'package:stream_chat/src/core/models/member.dart';
```
xsahil03x, line 3:
```suggestion
```

packages/stream_chat/lib/src/core/models/channel_state.dart
```diff
@@ -42,6 +43,9 @@ class ChannelState {
   /// The list of channel reads
   final List<Read>? read;
 
+  ///
```
xsahil03x, line 46:
```suggestion
  /// Relationship of the current user to this channel.
```

PR #1255, opened 2022-07-07
fix(ui): fix jiffy locale initialization

imtoori:


packages/stream_chat_flutter/lib/src/stream_chat.dart
```diff
@@ -141,7 +141,7 @@ class StreamChatState extends State<StreamChat> {
   @override
   void didChangeDependencies() {
     final currentLocale = Localizations.localeOf(context);
-    final languageCode = currentLocale.languageCode;
+    final languageCode = currentLocale.toString();
```
xsahil03x, line 144: this is a bit confusing as `currentLocale.toString()` does not give back the language code but the string representation of the locale.

Better to remove the `languageCode` field or rename it to better represent the value.

xsahil03x (approved):
LGTM

PR #1279, opened 2022-07-22
fix(ui): increase default file limit to 100MB

imtoori:


packages/stream_chat_flutter/lib/src/message_input/stream_message_input.dart
```diff
@@ -26,7 +26,7 @@ import 'package:stream_chat_flutter/stream_chat_flutter.dart';
 
 const _kMinMediaPickerSize = 360.0;
 
-const _kDefaultMaxAttachmentSize = 20971520; // 20MB in Bytes
+const _kDefaultMaxAttachmentSize = 104857600; // 100MB in Bytes
```
xsahil03x, line 29: Can we make it more expressive on how we get this number?
maybe something like...
```suggestion
const _kDefaultMaxAttachmentSize = 1024 * 1024; // 100MB in Bytes
```

imtoori: There's a comments saying what it is, but ok
```suggestion
const _kDefaultMaxAttachmentSize = 100 * 1024 * 1024; // 100MB in Bytes
```

PR #1280, opened 2022-07-22
fix(ui): fix image download on iOS

imtoori:


packages/stream_chat_flutter/lib/src/attachment_actions_modal.dart
```diff
@@ -405,7 +405,8 @@ class AttachmentActionsModal extends StatelessWidget {
       onReceiveProgress: progressCallback,
     );
     final result = await ImageGallerySaver.saveFile(filePath!);
-    downloadedPathCallback?.call((result as Map)['filePath']);
+    print('filePath: $filePath');
```
ayush221b, line 408: I think we should remove the print statement

imtoori:
```suggestion
```

packages/stream_chat_flutter/CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-## Upcming
+## Upocming
```
deven98, line 1:
```suggestion
## Upcoming
```

PR #1296, opened 2022-08-04
fix(ui): fix `stopTyping` unhandled exceptions when network is off or spotty.

xsahil03x:


imtoori (changes requested):
Let's add some tests on the key stroke handler

PR #1321, opened 2022-08-29
fix(llc): only start listening to client event once connected.

xsahil03x:
Fixes: https://getstream.zendesk.com/agent/tickets/25465


imtoori (approved):
LGTM

PR #1349, opened 2022-10-10
feat(ui): sendingIndicatorBuilder

geweald:
# Submit a pull request

## CLA

- [ ] The code changes follow best practices

Similar to `usernameBuilder` and others it allows customizing sending indicator in `BottomRow` without touching the whole `BottomRow` of the message

### DIFF
--- a/packages/stream_chat_flutter/CHANGELOG.md
+++ b/packages/stream_chat_flutter/CHANGELOG.md
@@ -1,3 +1,9 @@
+## UPCOMING
+
+✅ Added
+
+- Added `sendingIndicatorBuilder` to customize sending indicator in `StreamMessageWidget`.
+
 ## 5.0.1
 
 🔄 Changed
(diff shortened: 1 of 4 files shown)


imtoori:
I don't believe this is the best solution for this kind of use case.

Based on what we're doing in other places in the codebase I would:

- add a `sendingIndicatorBuilder` to the `BottomRow` widget
- add a `copyWith` method to the `BottomRow` widget, so that we can modify an instance easily
- add the default `BottomRow` widget implementation as the third parameter in the `bottomRowBuilder` callback 

what do you think

cc @xsahil03x

xsahil03x:
Yes, I agree with @imtoori, as then there will be multiple builders for the whole bottom row elements.

xsahil03x:
Hey @geweald, you can mark them deprecated with some info about the new implementation.

Issue #1379, opened 2022-11-10
Issues with photo attachments on web 


tanphuccgl:
### Which packages are you using?

stream_chat_flutter

### On what platforms did you experience the issue?

Web

### What version are you using?

5.0.1

### What happened?

Actual result: Show warning icon immediately then upload successfully
Expected result: should not show warning icon, if need time to load then maybe show loading icon instead

https://user-images.githubusercontent.com/80146701/201003728-6fc99604-eeb6-4ae2-a6cb-344eef2080b5.mov

### Steps to reproduce

```bash
Refer to the attached video.
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

HayesGordon:
Hi @tanphuccgl thank you for reporting this. I can reproduce this, and it's an easy fix that we will push soon. It's also an opportunity to expand this widget with some extra customization options for the error condition.

tanphuccgl:
Yes thanks for it. By the way, can you look into this problem: https://github.com/GetStream/stream-chat-flutter/issues/1346 . @HayesGordon

HayesGordon:
@tanphuccgl I will take a look at this as soon as possible as well 👍

tanphuccgl:
ok, thank you @HayesGordon

PR #1381, opened 2022-11-17
fix(ui): cached attachment image not showing on web

HayesGordon:
Fixes https://github.com/GetStream/stream-chat-flutter/issues/1379


xsahil03x (approved):
LGTM

Issue #1386, opened 2022-11-21
desktop_drop upgrade to version 0.4.0 required


ramy9109:
### Which packages are you using?

stream_chat_flutter

### On what platforms did you experience the issue?

Android

### What version are you using?

latest

### What happened?

The Android Gradle plugin supports only Kotlin Gradle plugin version 1.5.20 and higher.
The following dependencies do not satisfy the required version:
project ':desktop_drop' -> org.jetbrains.kotlin:kotlin-gradle-plugin:1.3.50

### Steps to reproduce

```bash
1. Go to '...'
2. Click on '...'
3. Scroll down to '...'
...
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

HayesGordon:
Related to https://github.com/getstream/stream-chat-flutter/issues/1375 and https://github.com/GetStream/stream-chat-flutter/issues/1372

Issue #1375, opened 2022-11-07
Bump minimum compatible Kotlin version to 1.5.20 via desktop_drop


CalHoll:
Just want to ping this issue that's "waiting on customer response" since I'm not the original poster I wasn't sure if it would get overlooked: https://github.com/GetStream/stream-chat-flutter/issues/1372

It's holding up a deployment that would fix crashes in production so any priority it could be given would be appreciated!

CalHoll:
`desktop_drop` has merged a commit that fixes the issue but it's not released as a package yet, here's the workaround:

```
dependency_overrides:
  desktop_drop:
    git:
      url: https://github.com/MixinNetwork/flutter-plugins.git
      path: packages/desktop_drop
      ref: 833cda265f2b2e9fe4105d4c56386c5ccd82464a
```

Looks like `image_gallery_saver` is also a problem...

CalHoll:
Here's the fix for image_gallery_saver - it's from the package maintainer, the repo was updated 2 months ago with the fixes for the latest version of flutter but they haven't updated the package on pub.dev

```
image_gallery_saver:
    git:
      url: https://github.com/hui-z/image_gallery_saver.git
      ref: 6676c04a858761736dc6b166ef14887578c6218f
```

imtoori:
unfortunately we have only two options

- wait for them to release the new versions and tell people to use the dependency overrides for the moment
- fork the packages 

cc @xsahil03x

CalHoll:
desktop_drop upgraded to 0.4 woo! Excited to get rid of these overrides.

Issue #1372, opened 2022-10-30
Issues while building the app


vasilich6107:
### Which packages are you using?

stream_chat_flutter

### On what platforms did you experience the issue?

Android

### What version are you using?

latest

### What happened?

Running Gradle task 'bundleRelease'...                          

FAILURE: Build failed with an exception.

* What went wrong:
The Android Gradle plugin supports only Kotlin Gradle plugin version 1.5.20 and higher.
The following dependencies do not satisfy the required version:
project ':desktop_drop' -> org.jetbrains.kotlin:kotlin-gradle-plugin:1.3.50

### Steps to reproduce

```bash
switch android gradle plugin to v7.3.1 and try to release the app
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

imtoori:
hey @vasilich6107 
Apparently this is due to a dependency
Switching to a less recent gradle plugin should work, can you try that?

CalHoll:
This is holding up a deployment which would fix crashes in production.

I found my way here after my build started failing, it's based on your "desktop_drop" dependency.

Oddly it seems that the issue is because "desktop_drop" has "kotlin-gradle-plugin:1.3.50" stated in the example folder, I'm not sure how this works.

I'm not very familiar with gradle so I've been poking around in the dark to try to figure out what is causing a bunch of crashes on production. I've discovered I need to get my deps upgraded to the latest to fix the issues but there's conflicts, here's a summary of my configuration:

andoid/app/build.gradle:
```
...
     compileSdkVersion 33
...

 dependencies {
         coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:1.2.2'
...
}
```

android/build.gradle:
```
buildscript {
    ext.kotlin_version = '1.7.20'
    repositories {
        google() 
        mavenCentral()
    }

    dependencies {
        classpath 'com.android.tools.build:gradle:7.3.1'
        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
    }
}
```

Here's some of the stack trace related to the build failure:
```
[ +104 ms] FAILURE: Build failed with an exception.
[   +5 ms] * What went wrong:
[        ] The Android Gradle plugin supports only Kotlin Gradle plugin version 1.5.20 and higher.
[        ] The following dependencies do not satisfy the required version:
[        ] project ':desktop_drop' -> org.jetbrains.kotlin:kotlin-gradle-plugin:1.3.50
[...]

PR #1388, opened 2022-11-23
chore(ui): update desktop drop dependency

HayesGordon:
Resolves https://github.com/getstream/stream-chat-flutter/issues/1386, https://github.com/getstream/stream-chat-flutter/issues/1375, and https://github.com/GetStream/stream-chat-flutter/issues/1372


xsahil03x (dismissed):
LGTM

Issue #1359, opened 2022-10-21
Increase connectivity_plus version to 3.0.2


flodaniel:
### Which packages are you using?

stream_chat_flutter, stream_chat_persistance, stream_chat_localizations

### On what platforms did you experience the issue?

iOS, Android

### What version are you using?

latest

### What happened?

Increase dependency of connectivity_plus to latest (3.0.2), to allow our own project to keep dependencies up to date. 

`Because stream_chat_flutter >=5.0.0 depends on stream_chat_flutter_core ^5.0.0 which depends on connectivity_plus ^2.1.0, stream_chat_flutter >=5.0.0 requires connectivity_plus ^2.1.0.`

### Steps to reproduce

```bash
Update packages and run flutter pub get
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #1389, opened 2022-11-23
chore(core): update connectivity plus dependency

HayesGordon:
Resolves https://github.com/getstream/stream-chat-flutter/issues/1359


xsahil03x (approved):
LGTM

Issue #1346, opened 2022-10-06
Issues with video attachments on web


Ayush-gupta-josh:
### Which packages are you using?

stream_chat_flutter

### On what platforms did you experience the issue?

Web

### What version are you using?

5.0.1

### What happened?

There are two scenarios where video attachment breaks.
1. When we select any video - 
    - When we select any video of size within the size limit then it breaks due to `Unexpected null value`.
    - In my opinion, this code is causing an error because you have mentioned that there will be no `URL` in case of web and here you are making it a mandatory value.
    ![implement](https://user-images.githubusercontent.com/67139749/194247890-f2bb7e61-e766-4820-be39-6c91125792e0.png)

    - Screen breaks with this error
    ![image](https://user-images.githubusercontent.com/67139749/194247723-9438765f-6c58-4e56-8cf6-3cc21320b143.png)
    
2. When we select a large video -
   - When we select a large video then there is no error on the screen instead it shows an `Unexpected null value` error in the console.

Reference Video

https://user-images.githubusercontent.com/67139749/194246733-4e6e8257-9f41-43e5-b501-6782b574101b.mp4

https://user-images.githubusercontent.com/67139749/194248480-3acc88a8-7c85-484f-8f16-6df12d616896.mp4

### Steps to reproduce

```bash
Refer to the attached video.
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

tanphuccgl:
@Ayush-gupta-josh do you have any solution?

PR #1390, opened 2022-11-25
fix(ui): video thumbnails on web null error

HayesGordon:
Resolves https://github.com/GetStream/stream-chat-flutter/issues/1346

Also makes the widget a little safer wherever we use it internally - ensuring a default thumbnail will be returned in the event any of the paths or URLs are null. Ideally, this widget should be refactored.


xsahil03x (dismissed):
LGTM

Issue #1347, opened 2022-10-06
StreamImageAttachment onReplyMessage doesn't work 


EtcGonza:
### Which packages are you using?

stream_chat_flutter

### On what platforms did you experience the issue?

iOS, Android

### What version are you using?

  stream_chat_flutter: ^5.0.0
  stream_chat_localizations: ^4.0.0

### What happened?

Hi! I update my project to v5 and I have a widget that use **StreamImageAttachment** and **StreamImageGroup** widgets. And I want to customize the onReplyMessage action of these widgets. But when I do it nothing happens. It's like the function doesn't execute. This is the code of my widget:

```
class MessageImages extends StatelessWidget {
  final Message message;
  final List<Attachment> imagesAttachment;

  const MessageImages({
    required this.imagesAttachment,
    required this.message,
    Key? key,
  }) : super(key: key);

  @override
  Widget build(BuildContext context) {
    final widgetChatProvider =
        Provider.of<ChatWidgetMannagmentProvider>(context, listen: false);

    final _messageChatProvider =
        Provider.of<ChatMessagesProvider>(context, listen: false);

    const messageTheme = StreamMessageThemeData(
      messageTextStyle: TextStyle(
        color: CustomStylesHelper.whiteColor,
        fontSize: 14,
      ),
    );

    return imagesAttachment.length > 1
        ? StreamImageGroup(
            constraints: BoxConstraints(
              maxWidth: 400,
              minWidth: 400,
              maxHeight: MediaQuery.of(context).size.height * 0.3,
            ),
            onReplyMessage: (msg) => print(msg.text),
            onShowMessage: (msg, channel) =>
                widgetChatProvider.navigateToMessage(context, channel, message),
            images: imagesAttachment,
            message: message,
            messageTheme: messageTheme,
          )
        : SizedBox(
            width: MediaQuery.of(context).size.width * 1,
            height: MediaQuery.of(context).size.height * 0.2,
            child: StreamImageAttachment(
              constraints: BoxConstraints(
                minWidth: MediaQuery.of(context).size.width * 1,
                minHeight: MediaQuery.of(context).size.height * 0.2,
              ),
              messageTheme: messageTheme,
              message: message,
              attachment: imagesAttachment.first,
              showTitle: false,
              onReplyMessage: (msg) => print(msg.text),
              onShowMessage: (msg, channel) => widgetChatProvider
                  .navigateToMessage(context, channel, message),
            ),
          );
  }
}
```

[...]

PR #1391, opened 2022-11-25
fix(ui): on reply callback not used in AttachmentActionsModal

HayesGordon:
Resolves https://github.com/GetStream/stream-chat-flutter/issues/1347


xsahil03x (dismissed):
LGTM

PR #1440, opened 2023-01-26
fix(ui): temp file path on Windows

adasiewiczr:
Fixed temp file path on Windows. For some reason, the `tempFilePath ` contains an extra `/` at the start. Make sure it gets removed before the file is created.


leandroBorgesFerreira (dismissed):
I can't test this =|

xsahil03x (approved):

PR #1444, opened 2023-01-26
fix(localization): update some copies due to correctness

ccebrecos:
## CLA

Changed some copies to be more precise with some expressions.

### DIFF
--- a/packages/stream_chat_localizations/CHANGELOG.md
+++ b/packages/stream_chat_localizations/CHANGELOG.md
@@ -8,6 +8,7 @@
 🔄 Changed
 
 * Some of the `Spanish` translations have been updated/changed for better understanding.
+* Some of the `Catalan` translations have been updated/changed for better understanding.
 
 ## 4.0.0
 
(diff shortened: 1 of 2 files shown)


ccebrecos:
As no version with Catalan was generated didn't change the `CHANGELOG.md` as the previous statement `Added support for Catalan` is still valid.

I find quite weird to generate an entry in `changes` for that but adding in order to make the tests green.

adasiewiczr (approved):
hey @ccebrecos, thank you for your contribution!
It's indeed a bit weird but it's not a common situation. Adding an entry to the CHANGELOG helps keep a good history of changes as you can always check the message and find a PR related to the change. In a situation like this, it might be a good idea to add both PR numbers into a single CHANGELOG.
However, let's keep it as is and we'll improve our recommendations.

ccebrecos:
Oops. Didn't thought about that kind of workaround, I will use it onwards. Thanks @adasiewiczr!

Issue #1338, opened 2022-09-28
sharing emails with custom domains shouldn't trigger url preview


CalHoll:
### Which packages are you using?

stream_chat_flutter

### On what platforms did you experience the issue?

iOS, Android

### What version are you using?

The latest Beta

### What happened?

when sharing my custom email in the chat it triggered a url preview - hello@cheerly.app

### Steps to reproduce

```bash
share a custom domain url in chat (i.e. hello@cheerly.app)
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

HayesGordon:
Hi @CalHoll, thanks for reporting this! We will look into resolving this as soon as possible

xsahil03x:
Hey @CalHoll, added the possibility to filter some url's in [#1516](https://github.com/GetStream/stream-chat-flutter/pull/1516).

eg:
  ```dart
  StreamMessageInput(
    ogPreviewFilter: (matchedUri, messageText) {
      final url = matchedUri.toString();
      if (url.contains('hello@cheerly.app')) {
        // Return false to prevent the OG preview from being built.
        return false;
      }
      // Return true to build the OG preview.
      return true;
  ),
  ```

PR #1516, opened 2023-05-01
feat(ui): Added `StreamMessageInput.ogPreviewFilter` to allow users to filter out the og preview links.

xsahil03x:
Fixes: #1338


kanat (approved):
LGTM

Issue #1325, opened 2022-09-08
Too many log entries per single error when using external logging


mkobuolys:
### Please select which package this feature is related to

stream_chat_flutter

### Which platforms would this feature impact?

iOS, Android

### Is your feature request related to a problem?

When using external logging, a single stack trace provides too many error entries in the external logging service (this case, DataDog).
<img width="1758" alt="Screenshot 2022-09-08 at 09 56 07" src="https://user-images.githubusercontent.com/17415795/189055433-c79c25e6-a858-4f0c-826b-22630a4c5c67.png">

### Describe the solution that you'd like

It would be nice to enable/disable this pretty logging option to provide the whole error with the stack trace in a single log entry.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

xsahil03x:
Hey @mkobuolys, we added support for external interceptors in `StreamChatClient`.
It's currently in [develop](https://github.com/GetStream/stream-chat-flutter/pull/1522) and will be released soon.
You can now override the default logging interceptor with your own and log as per your choice.

  ```dart
    final client = StreamChatClient(
      chatApiInterceptors: [
        InterceptorsWrapper(
          onRequest: (options, handler) {
            // Do something before request is sent.
            return handler.next(options);
          },
          onResponse: (response, handler) {
            // Do something with response data
            return handler.next(response);
          },
          onError: (DioError e, handler) {
            // Do something with response error
            return handler.next(e);
          },
        ),
      ],
  );
  ```
  
Feel free to re-open if this does not fulfill your requirements.

Thanks,
Sahil

mkobuolys:
Oh, it sounds great. I will definitely check that out, thanks!

PR #1522, opened 2023-05-05
feat(llc): add support for external interceptors.

xsahil03x:
Fixes: #1265, #1325


packages/stream_chat/lib/src/core/api/stream_chat_api.dart
```diff
@@ -14,7 +15,7 @@ import 'package:stream_chat/src/core/http/token_manager.dart';
 
 export 'device_api.dart' show PushProvider;
 
-/// ApiClient that wraps every other specific api
+/// Api_client that wraps every other specific api
```
xsahil03x, line 18:
```suggestion
/// ApiClient that wraps every other specific api
```

Issue #1682, opened 2023-07-21
Sending System Message sends Regular Message instead


mingjunsiek:
### Which packages are you using?

stream_chat_flutter

### On what platforms did you experience the issue?

iOS, Android

### What version are you using?

stream_chat_flutter: ^6.5.0

### What happened?

Based on the flutter docs in getStream [website](https://getstream.io/chat/docs/flutter-dart/silent_messages/?language=dart), it states the following: 

<img width="816" alt="image" src="https://github.com/GetStream/stream-chat-flutter/assets/13094556/16476137-be51-41fe-b69f-ca3f241fd6b5">

Note the sentence: `The user who sends a system message client-side should have the Create System Message permission.` which meant that we can create system message on the client side.

I have already enabled the `Create System Message` permission on the GetStream dashboard for my project.

But when I used the following code to send the system message via Flutter, the message was somehow converted to a `regular` message instead of `system:
```
final message = Message(
  text: 'systemMessage',
  type: 'system',
);

await channel.sendMessage(message);
```
Screenshot:
<img width="360" alt="image" src="https://github.com/GetStream/stream-chat-flutter/assets/13094556/7efe1744-39a1-401f-9e86-36633c9e1a7d">
- Message stored in GetStream
  - <img width="400" alt="image" src="https://github.com/GetStream/stream-chat-flutter/assets/13094556/0e68e659-e433-4bf5-9f41-c0ae127e257a">

### Steps to reproduce

```bash
1. Create a Message object with type = 'system'

final message = Message(
  text: 'systemMessage',
  type: 'system',
);
2. Send this message to the channel

await channel.sendMessage(message);

3. Message is turned to a regular message instead of a system message.
```

### Supporting info to reproduce

- Printing the Message Object
  - Message(91f2a6ba-5121-4637-b321-93feb1b6ad84, Replying will automatically move you to the next stage in the hiring process., system, [], [], null, null, null, null, null, null, null, 0, null, null, false, false, null, null, null, null, null, null, null, null, false, null, null, null, {}, MessageSendingStatus.sending, null)
  - <img width="360" alt="image" src="https://github.com/GetStream/stream-chat-flutter/assets/13094556/7efe1744-39a1-401f-9e86-36633c9e1a7d">
- Message stored in GetStream
  - <img width="400" alt="image" src="https://github.com/GetStream/stream-chat-flutter/assets/13094556/0e68e659-e433-4bf5-9f41-c0ae127e257a">

[...]

PR #1685, opened 2023-07-24
feat(llc): add support for setting message.type.

xsahil03x:
Fixes: #1682

### DIFF
--- a/packages/stream_chat/CHANGELOG.md
+++ b/packages/stream_chat/CHANGELOG.md
@@ -1,3 +1,15 @@
+## Upcoming
+
+✅ Added
+
+- Added support for setting `Message.type`. [#1682](https://github.com/GetStream/stream-chat-flutter/issues/1682)
+  ```
+  It is now possible to send system messages. System messages differ from normal messages in the way they are
+  presented to the user. Like the name says, system messages are normally send from the system itself, but a user is
+  able to send it as well by specifying type: 'system' with the message. The user who sends a system message client-side
+  should have the Create System Message permission. Server-side system messages don't need that permission.
+  ```
+
 ## 6.6.0
 
 🔄 Changed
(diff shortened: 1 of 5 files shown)


luis901101:
Thanks for this fix, but now that I upgraded to use latest:
```dart
  stream_chat_flutter: ^6.8.0
  stream_chat_persistence: ^6.7.0
  stream_chat_localizations: ^5.8.0
```

I'm having 403 errors when I send messages...
```
DioException ║ Status: 403 Forbidden...

╔ DioExceptionType.badResponse
║    {
║         code: 17,
║         "SendMessage failed with error: "SendMessage failed with error: "SendMessage failed with error: "SendMessage fai
║         led with error: "SendMessage failed with error: "SendMessage failed with error: "SendMessage failed with error:
║         "SendMessage failed with error: "SendMessage failed with error: "SendMessage failed with error: "creating system
║          messages not allowed"""""""""""
║         StatusCode: 403,
║         duration: "0.00ms",
║         more_info: "https://getstream.io/chat/docs/api_errors_response",
║         details: []
║    }
╚════════════════════════════════════════════════════════════════════════════════════════════════════════════════════════╝
```


I checked the `StreamMessageInput` widget in the `stream_chat_flutter: ^6.8.0` and I've found this:

```dart
/// Sends the current message
  Future<void> sendMessage() async {
    if (_timeOut > 0 ||
        (_effectiveController.text.trim().isEmpty &&
            _effectiveController.attachments.isEmpty)) {
      return;
    }

    final streamChannel = StreamChannel.of(context);
    final channel = streamChannel.channel;
    var message = _effectiveController.value.copyWith(
      type: 'system'
    );

```

maybe this is the cause... 

Anyway I just checked the same line 1333 in the repo and I don't see that `.copyWith(...)`

https://github.com/GetStream/stream-chat-flutter/blob/51e005deacc98788b43635be1a2ac25a52ababb5/packages/stream_chat_flutter/lib/src/message_input/stream_message_input.dart#L1323-L1333

PR #1772, opened 2023-11-12
demo of status change not working

talksik:
An action button in the appbar tries to change a parameter of the local user within `extraData` and it does that successfully. However, for some reason, the updated user received on the stream listener has the updated property online set to false.

Sorry for the noise, I had to add Linux as a platform because that's my development machine. It should be enough to only focus on `example/lib/main.dart` as that's the only place I had changes.

This issue is part of an email conversation I have with Eve from the getstream team. I believe this is a critical bug so hoping this will get resolved soon! :) If you want me to take down this PR, please let me know 👍🏾

### DIFF
--- a/packages/stream_chat/example/.metadata
+++ b/packages/stream_chat/example/.metadata
@@ -4,7 +4,27 @@
 # This file should be version controlled and should not be manually edited.
 
 version:
-  revision: 78910062997c3a836feee883712c241a5fd22983
-  channel: stable
+  revision: "2524052335ec76bb03e04ede244b071f1b86d190"
+  channel: "stable"
 
 project_type: app
+
+# Tracks metadata for the flutter migrate command
+migration:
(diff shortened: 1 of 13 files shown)


talksik:
Any update on this?

esarbanis:
Hi @talksik ! Thanks for providing this reproduction demo. Can you please switch it to `draft` and provide the link to it from an opened issue? It is a bit hard to track issues if they do not appear in the issue list :pray:

talksik:
@esarbanis Yes sir, I have changed this to draft and created an issue: https://github.com/GetStream/stream-chat-flutter/issues/1810

esarbanis:
Thank you @talksik I appreciate it!

deven98:
Closing this as we've solved the original issue. Thanks for the report and the example!

PR #1840, opened 2024-01-31
feat(ui): audio message recording components

esarbanis:
# Remaining:
- [x] Export all theming
- [X] Animations when appearing and disappearing
- [x] Finalize component API
- [x] Test coverage
- [ ] Prevent keyboard from collapsing when open
- [ ] Fix message action button spacing 
- [x] Fix height of `MessgeInput` when recording
- [x] Create rollup file for the various components export
- [x] Replace `Timer.periodic` with `Future.delayed` when possible

## CLA

https://github.com/GetStream/stream-chat-flutter/assets/633903/f196f39e-f2ad-4b7d-9b01-7810ef1961ca


deven98:
Screenshot 2024-02-13 at 7 46 14 PM

I'm not sure if this is specific to this PR but the spacing between message input elements seems off. 
Testing on Android Pixel 5.

packages/stream_chat_flutter/lib/src/message_input/audio_message_send_button/stream_audio_message_overlays.dart
```diff
@@ -0,0 +1,118 @@
...
+          fontSize: 15,
+        ),
+        child: Container(
+          height: 28,
+          width: MediaQuery.of(context).size.width,
+          alignment: Alignment.center,
+          decoration: BoxDecoration(
+            color: audioRecordingMessageTheme.audioButtonBannerColor,
+          ),
+          child: const Text('Hold to start recording.'),
```
deven98, line 33: Is this something we need to support i18n for?

deven98:
Screenshot 2024-02-13 at 7 50 55 PM

Let's add some right padding for the overlay. Maybe even an Opacity Tween of some kind?

deven98:
Could we find a way to prevent the keyboard collapsing (if open) when the audio button is hit?

deven98:
The height of the audio recording bar (when recording) is not the same as the `MessageInput` leading to it collapsing slightly when recording. Visible in your video but somehow more visible on Android.

packages/stream_chat_flutter/lib/src/message_input/audio_message_send_button/stream_audio_message_send_button.dart
```diff
@@ -0,0 +1,224 @@
+import 'dart:async';
+
+import 'package:flutter/services.dart';
+import 'package:flutter/widgets.dart';
+import 'package:stream_chat_flutter/src/message_input/audio_message_send_button/stream_audio_message_controllers.dart';
+import 'package:stream_chat_flutter/src/message_input/audio_message_send_button/stream_audio_message_gesture_state.dart';
+import 'package:stream_chat_flutter/src/message_input/audio_message_send_button/stream_audio_message_overlays.dart';
```
deven98, line 7: Let's create a single export file for the audio_message_send_button folder, and maybe even export it from stream_chat_flutter?

packages/stream_chat_flutter/lib/src/message_input/audio_message_send_button/stream_audio_message_send_button.dart
```diff
@@ -0,0 +1,224 @@
...
+import 'package:stream_chat_flutter/stream_chat_flutter.dart';
+
+/// A callback called when the user starts recording an audio message.
+typedef OnRecordingStart = void Function();
+
+/// A callback called when the user ends recording an audio message.
+typedef OnRecordingEnd = void Function();
+
+/// A callback called when the user cancels recording an audio message.
+typedef OnRecordingCanceled = void Function();
```
deven98, line 17: Why not use `VoidCallback` here? (Not objecting to this, just curious) - since the user only sees the parameter name and using a custom callback is a bit confusing since they have no clue what parameters to use when passing the callback.

packages/stream_chat_flutter/lib/src/message_input/audio_message_send_button/stream_audio_message_controllers.dart
```diff
@@ -0,0 +1,186 @@
...
+        timer.cancel();
+        if (mounted) {
+          setState(() {
+            iconColor =
+                audioRecordingMessageTheme.recordingIndicatorColorActive;
+            _startTime = DateTime.now();
+          });
+        }
+      },
+    );
```
deven98, line 49: If this is meant to run once (assuming by the `cancel()`), why can't we use `Future.delayed()`?

packages/stream_chat_flutter/lib/src/message_input/audio_message_send_button/stream_audio_message_controllers.dart
```diff
@@ -0,0 +1,186 @@
...
+
+    _timer = Timer.periodic(
+      const Duration(seconds: 1),
+      (timer) {
+        if (_startTime == null) {
+          return;
+        }
+        setState(() {});
+      },
+    );
```
deven98, line 59: If this is for setting clock count, can we use a `ValueNotifier` coupled with a builder for it instead of a full `setState()` as it is a stateful widget anyway?

packages/stream_chat_flutter/lib/src/message_input/audio_message_send_button/stream_audio_message_controllers.dart
```diff
@@ -0,0 +1,186 @@
...
+
+class _StreamAudioMessageControllersState
+    extends State<StreamAudioMessageControllers> with TickerProviderStateMixin {
+  Color? iconColor;
+  DateTime? _startTime;
+  Timer? _timer;
+  late final AnimationController _controller;
+
+  @override
+  void didChangeDependencies() {
```
deven98, line 27: Apart from the theme, why are we initialising in didChangeDeps for controllers and such? Any situations where initState was failing?

packages/stream_chat_flutter/lib/src/message_input/audio_message_send_button/stream_audio_message_overlays.dart
esarbanis: Yes, but I want to gather all text by the end of this component, and deal with it in a separate PR after the functionality

deven98: Okay, makes sense.

packages/stream_chat_flutter/lib/src/message_input/audio_message_send_button/stream_audio_message_send_button.dart
esarbanis: I wasn't sure if I had to create a rollup file, since I haven't seen any on the source. I would be happy to do it though 🙂

packages/stream_chat_flutter/lib/src/message_input/audio_message_send_button/stream_audio_message_controllers.dart
esarbanis: This will change with the implementation of the recording function, I will revisit it then

packages/stream_chat_flutter/lib/src/message_input/audio_message_send_button/stream_audio_message_controllers.dart
esarbanis: There was an exception thrown when disposing the `_controller`, than indicated to put it in `didChangeDependencies`

esarbanis:
> Let's add some right padding for the overlay. Maybe even an Opacity Tween of some kind?

In the design there no padding on the right, or an opacity change

> Could we find a way to prevent the keyboard collapsing (if open) when the audio button is hit?

This will disorient the user as now their finger is going to long tap in another place than where the button will be.

> The height of the audio recording bar (when recording) is not the same as the MessageInput leading to it collapsing slightly when recording. Visible in your video but somehow more visible on Android.

I must have missed it, I tested in iOS mostly 👍

packages/stream_chat_flutter/lib/src/message_input/audio_message_send_button/stream_audio_message_send_button.dart
esarbanis: I don't really like using other SDK's code as ours, even if it is from Flutter itself. Also this is provisioning for changes in the future 😉

esarbanis:
@deven98 the keyboard is not such an easy fix, I timeboxed it for about an hour and didn't make any significant progress, also I have checked the left action buttons both in iPhone Simulator and real device, as well as android emulator and real device and I cannot make it appear as in your screenshot ... 🤔  
Screenshot 2024-02-14 at 2 11 05 PM

deven98 (approved):
LGTM

PR #1913, opened 2024-05-01
feat(core): migrate connectivity_plus library

esarbanis:
# Submit a pull request

## CLA

### DIFF
--- a/packages/stream_chat_flutter_core/lib/src/stream_chat_core.dart
+++ b/packages/stream_chat_flutter_core/lib/src/stream_chat_core.dart
@@ -65,7 +65,7 @@ class StreamChatCore extends StatefulWidget {
   /// Stream of connectivity result
   /// Visible for testing
   @visibleForTesting
-  final Stream<ConnectivityResult>? connectivityStream;
+  final Stream<List<ConnectivityResult>>? connectivityStream;
 
   @override
   StreamChatCoreState createState() => StreamChatCoreState();
@@ -108,7 +108,7 @@ class StreamChatCoreState extends State<StreamChatCore>
   /// The current user as a stream
   Stream<User?> get currentUserStream => client.state.currentUserStream;
 
-  StreamSubscription<ConnectivityResult>? _connectivitySubscription;
+  StreamSubscription<List<ConnectivityResult>>? _connectivitySubscription;
 
   var _isInForeground = true;
   var _isConnectionAvailable = true;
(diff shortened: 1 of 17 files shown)


packages/stream_chat_flutter_core/lib/src/stream_chat_core.dart
```diff
@@ -121,13 +121,14 @@ class StreamChatCoreState extends State<StreamChatCore>
...
-    Stream<ConnectivityResult>? connectivityStream,
+    Stream<List<ConnectivityResult>>? connectivityStream,
   ]) {
     if (_connectivitySubscription == null) {
       connectivityStream ??= Connectivity().onConnectivityChanged;
       _connectivitySubscription =
           connectivityStream.distinct().listen((result) {
-        _isConnectionAvailable = result != ConnectivityResult.none;
+        _isConnectionAvailable =
+            result.length == 1 && result.first != ConnectivityResult.none;
```
Brazol, line 131: I think if there is more than one element in result it still can be connected. Or even when there is more than one then it's almost certainly connected?

For iOS it can be `something` + `vpn` for example
https://github.com/fluttercommunity/plus_plugins/pull/2599


```suggestion
        _isConnectionAvailable =
            result.length > 1 || (result.length == 1 && result.first != ConnectivityResult.none);
```

packages/stream_chat_flutter_core/pubspec.yaml
```diff
@@ -11,7 +11,7 @@ environment:
 
 dependencies:
   collection: ^1.17.2
-  connectivity_plus: ">=4.0.2 <6.0.0"
+  connectivity_plus: ">=4.0.2 <7.0.0"
```
Brazol, line 14: Shouldn't we require 6.0.0 now? It won't comple for versions before that because of the result being a list now.

```suggestion
  connectivity_plus: ">=6.0.1 <7.0.0"
```

packages/stream_chat_flutter_core/pubspec.yaml
Brazol, line 1: Connectivity Plus from version 6.0.1 requires Flutter 3.19.0 and Dart 3.3.0. Is it fine for us for v7? If so shouldn't we bump those in our pubspec also?

```
Plugin now requires the following:

Flutter >=3.19.0
Dart >=3.3.0
```

esarbanis: This is aimed towards beta

Brazol: Oh it's a merge to beta so we can require 3.19 there, still shouldn't we bump those in pubspecs also?

esarbanis: Will do on a separate PR, I also tested this change with flutter 3.16 and 3.19, compiled and all tests were passing, also the example app was running with no problem

Brazol (approved):
LGTM

PR #1947, opened 2024-06-11
fix(ui): extends predicates to mobile

### LINKED ISSUE #1796: Support sendMessageKeyPredicate on Mobile

stream_chat_flutter

iOS, Android

I have mobile users with external keyboards who would like to be able to submit messages using a key combination. (Specifically, CMD+Enter on iOS)

It looks like send message key handling was explicitly omitted from mobile in this commit: https://github.com/GetStream/stream-chat-flutter/pull/1222/commits/d3fff37fcdd7adcc1defb63eeeb1d48e422fedbb

I would like to be able to set a sendMessageKeyPredicate and have it be enabled on all platforms. I believe that if omitting that behavior from mobile was desired it should be done when building the predicate.

_No response_

### Additional context

_No response_

### Code of Conduct

deven98:
Fixes #1796 

Earlier implementation was only listening for keystrokes on desktop. New implementation also does the same by default but allows the user to customise and add their own send and clear predicates.


kanat (approved):
LGTM

PR #1953, opened 2024-06-14
fix(ui): added calculation for reaction picker tail

### LINKED ISSUE #1879: Reaction Picker Tail is always shown when the Reaction Picker is enabled in the default message widget

stream_chat_flutter, stream_chat_persistance, stream_chat_localizations

iOS, Android

7.1.0

### What happened?

When using the default message widget to display messages, the reaction picker tail is always shown when 'showReactionPicker' is set to true. ~~The tail should only be shown when a message has reactions.~~ **_Edit: I realize that I had originally misunderstood what I was seeing. The tail is part of the message picker itself. It should never be visible in the message list at all. I do not know why this is occurring._**

Screenshot 2024-03-12 at 11 51 02 AM

Disabling the reaction picker hides the tail, but that disables the ability to react to messages entirely.

I believe the issue lies in the following code in the StreamMessageWidget class:

`showReactionPickerTail: widget.showReactionPicker,`

### Steps to reproduce

```bash
1. View any channel displaying messages using the default widget with showReactionPicker set to true.
2. Every message will have the reaction tail shown, regardless of the reaction count.
```

_No response_

### Relevant log output

_No response_

### Flutter analyze output

_No response_

### Flutter doctor output

_No response_

### Code of Conduct

deven98:
Fixes #1879 

Reaction picker tails are currently visible in the list view and also show when reaction picker is disabled. This PR exposes a new property for enabling and disabling the tail as well as a default calculation for when the tail should be enabled.

### DIFF
--- a/packages/stream_chat_flutter/CHANGELOG.md
+++ b/packages/stream_chat_flutter/CHANGELOG.md
@@ -2,13 +2,17 @@
 
 ✅ Added
 
-- Copying a message now replaces the User IDs with user names.
-- Exported thumbnail widgets from the package.
 - Added `customAttachmentBuilders` parameter for `StreamAttachmentWidgetBuilder.defaultBuilders`.
 - `attachmentBuilders` parameter for `StreamMessageWidget` now only expects custom builders.
 - Added `StreamMediaAttachmentBuilder` widget to show media attachments in a message.
+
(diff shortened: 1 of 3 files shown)


Nash0x7E2 (approved):
LGTM

PR #1980, opened 2024-07-08
fix(core): fixed background events being called in foreground

### LINKED ISSUE #1802: onBackgroundEventReceived is called when app goes in foreground

stream_chat_flutter, stream_chat_persistance, stream_chat_localizations

Android

  stream_chat_flutter: 6.12.0
  stream_chat_localizations: 5.12.0
  stream_chat_persistence: 6.10.0

### What happened?

I receive a message.new event in `onBackgroundEventReceived` when i should not as the app is in foreground.
I also receive old message.new events, so that even if i have the chat view open, seeing a message from another user, as soon as the app goes to the background, i get the message.new event from that message.

### Steps to reproduce

```bash
1. Go to the chat view
2. Receive a message from another user
3. Background and foreground the app
4. Message.new event is received even though the app is in foreground.
```

_No response_

### Relevant log output

_No response_

### Flutter analyze output

_No response_

### Flutter doctor output

_No response_

### Code of Conduct

deven98:
Fixes #1802

### DIFF
--- a/packages/stream_chat_flutter/CHANGELOG.md
+++ b/packages/stream_chat_flutter/CHANGELOG.md
@@ -3,6 +3,7 @@
 🐞 Fixed
 
 - Fixed null errors in web from markdown.
+- Fixed bug causing background events to be sent in foreground.
 
 ## 7.3.0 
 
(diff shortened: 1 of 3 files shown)


Brazol (approved):
LGTM

PR #1981, opened 2024-07-08
fix(llc): fixed user parsing leading to incorrect online presence

### LINKED ISSUE #1810: Status change doesn't update properly and sets user offline.

stream_chat

Linux

6.9

### What happened?

An action button in the appbar tries to change a parameter of the local user within extraData and it does that successfully. However, for some reason, the updated user received on the stream listener has the updated property online set to false.

Sorry for the noise, I had to add Linux as a platform because that's my development machine. It should be enough to only focus on example/lib/main.dart as that's the only place I had changes.

[Associated PR](https://github.com/GetStream/stream-chat-flutter/pull/1772)

### Steps to reproduce

```bash
1. Go to associated PR linked above.
2. Run the app
3. Click the test button in the navbar.
4. _You will see the status of the user change to offline._
```

_No response_

### Relevant log output

_No response_

### Flutter analyze output

_No response_

### Flutter doctor output

_No response_

### Code of Conduct

deven98:
Fixes #1810

### DIFF
--- a/packages/stream_chat/CHANGELOG.md
+++ b/packages/stream_chat/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 🐞 Fixed
 
-- Fixed incorrect user object and online presence when user was updated.
+- Fixed incorrect user object and online presence when user is updated.
 
 ## 7.3.0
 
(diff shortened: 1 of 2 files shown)


Brazol (approved):
LGTM

PR #2005, opened 2024-08-21
feat(llc): allow setting http client adapter to use native http

### LINKED ISSUE #1989: Allow using native HTTP

stream_chat

iOS, Android, MacOS

The HTTP Implementation in dart:io has many known issues. 

- with VPNs: https://github.com/dart-lang/sdk/issues/41376
- Bad File descriptor erros when the app goes into background: https://github.com/dart-lang/http/issues/197
- No support for HTTP 2.0

`stream_chat` uses dio, with uses the http implementation from dart:io by default. It can be configured to use other implementations as well though, for example using https://pub.dev/packages/native_dio_adapter

But `stream_chat` does not expose the Dio client via exports, so it's not possible for users to change that.
Even when importing private APIs, it's difficult and error-prone to do so, as one would need to pass their own 
`StreamChatApi` with a custom implementation of `StreamHttpClient` to the `StreamChatClient`.

I would like `stream_chat` to allow for a way to configure dio's `httpClientAdapter`.

A flexible way to do so would be to allow passing in a `Dio Function() dioFactory`, where users would be able to do something like

```dart
dioFactory: () {
   final dio = Dio();
   dioClient.httpClientAdapter = NativeAdapter();
   return dio;
}
```

I would also be ok with defaulting to `NativeAdapter`, but this would introduce a new dependency to stream_chat and slightly increase bundle size for users who don't already depend on `cupertino_http` or `cronet_http`, so I can imagine that this is not a desired solution.

### Additional context

It seems that it was an architectural decision to make using dio an implementation detail which is not exposed to the user, and this would break that.
I understand that, but given the limitations of the of the current approach I think that overall, this is still worth it. 

### Code of Conduct

knaeckeKami:
# Submit a pull request

## CLA

fixes #1989

### DIFF
--- a/packages/stream_chat/CHANGELOG.md
+++ b/packages/stream_chat/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## unreleased
 
 ✅ Added
-- Add `httpClientAdapter` in order to use native http implementations
+- Added `httpClientAdapter` to `StreamHttpClient` that enables the use native http implementations
 
 ## 8.0.0
 
--- a/packages/stream_chat/CHANGELOG.md
+++ b/packages/stream_chat/CHANGELOG.md
@@ -1,3 +1,8 @@
+## unreleased
+
+✅ Added
+- Add `httpClientAdapter` in order to use native http implementations
+
 ## 8.0.0
 
 🐞 Fixed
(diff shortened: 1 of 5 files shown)


packages/stream_chat/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## unreleased
+
+✅ Added
+- Add `httpClientAdapter` in order to use native http implementations
```
Brazol, line 4:
```suggestion
- Added `httpClientAdapter` to `StreamHttpClient` that enables the use native http implementations
```

Brazol (approved):
LGTM

deven98 (approved):
LGTM

deven98:
Thanks for the PR @knaeckeKami! This should be merged soon and the changes should be released in the next version. You can of course use the git version if you need the same immediately. Thanks!

PR #2012, opened 2024-08-28
fix(ui): added option to use native picker for mobile

deven98:
Adds the `useNativePickerOnMobile` boolean to force use of native picker on mobile platforms.

### DIFF
--- a/packages/stream_chat_flutter/lib/src/message_input/attachment_picker/stream_attachment_picker_bottom_sheet.dart
+++ b/packages/stream_chat_flutter/lib/src/message_input/attachment_picker/stream_attachment_picker_bottom_sheet.dart
@@ -78,6 +78,7 @@ Future<T?> showStreamAttachmentPickerModalBottomSheet<T>({
   bool useRootNavigator = false,
   bool isDismissible = true,
   bool enableDrag = true,
+  bool useNativeAttachmentPickerOnMobile = false,
   RouteSettings? routeSettings,
   AnimationController? transitionAnimationController,
   Clip? clipBehavior = Clip.hardEdge,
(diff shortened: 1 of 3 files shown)


packages/stream_chat_flutter/lib/src/message_input/attachment_picker/stream_attachment_picker_bottom_sheet.dart
```diff
@@ -115,7 +116,7 @@ Future<T?> showStreamAttachmentPickerModalBottomSheet<T>({
               currentPlatform == TargetPlatform.linux ||
               currentPlatform == TargetPlatform.windows;
 
-          if (isWebOrDesktop) {
+          if (isWebOrDesktop || useNativeAttachmentPickerOnMobile) {
             return webOrDesktopAttachmentPickerBuilder.call(
```
Brazol, line 120: Maybe we should rename this now `webOrDesktopAttachmentPickerBuilder`? It's pretty confusing looking at the code why it works for native mobile scenarios also, as the whole flow from this point only refers to `desktop` and `web`. 
Maybe `nativeAttachmentPickerBuilder` would be better?

deven98: I agree to the `nativeAttachmentPickerBuilder` name but there's several other associated config objects that I would not be sure how to rename. We can discuss this.

Nash0x7E2: Maybe a comment linking to the docs for this with more info would be helpful here. Easy to get an impression that everything native is handled automatically by the SDK.

Brazol (approved):

PR #2045, opened 2024-11-06
fix(ui): added scrolling to reactions

d3xvn:
Added scrolling if reaction count is high.


xsahil03x:
Closing in favor of #2248

Issue #2030, opened 2024-10-01
video_thumbnail: Namespace not specified. Specify a namespace in the module's build file.


mellow0412:
### Which packages are you using?

stream_chat_flutter

### On what platforms did you experience the issue?

Android

### What version are you using?

stream_chat_flutter - 8.1.0

### What happened?

I am getting a compile time error describing that the video_thumbnail package needs to specify namespace in its build file.

### Steps to reproduce

```bash
Using the gradle-8.3-all produces this issue when its compiled for android 
My flutter version is Flutter (Channel stable, 3.24.2)
```

### Relevant log output

```shell
FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring project ':video_thumbnail'.
> Could not create an instance of type com.android.build.api.variant.impl.LibraryVariantBuilderImpl.
   > Namespace not specified. Specify a namespace in the module's build file. See https://d.android.com/r/tools/upgrade-assistant/set-namespace for information about setting the namespace.

     If you've specified the package attribute in the source AndroidManifest.xml, you can use the AGP Upgrade Assistant to migrate to the namespace value in the build file. Refer to https://d.android.com/r/tools/upgrade-assistant/agp-upgrade-assistant for general information about using the AGP Upgrade Assistant.
```

### Flutter doctor output

```shell
PS D:\StudioProjects\chat_app> flutter doctor
Doctor summary (to see all details, run flutter doctor -v):
[√] Flutter (Channel stable, 3.24.2, on Microsoft Windows [Version 10.0.26100.1457], locale en-IN)
[√] Windows Version (Installed version of Windows is version 10 or higher)
[√] Android toolchain - develop for Android devices (Android SDK version 35.0.0)
[√] Chrome - develop for the web
[X] Visual Studio - develop Windows apps
    X Visual Studio not installed; this is necessary to develop Windows apps.
      Download at https://visualstudio.microsoft.com/downloads/.
      Please install the "Desktop development with C++" workload, including all of its default components
[√] Android Studio (version 2024.1)
[√] Connected device (3 available)
[√] Network resources
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

mellow0412:
I found a workaround to fix it temporarily

editing the video_thumbnail android/app/build.gradle fixes it
add this line in the android/app/build.gradle inside android config

android {
  ...

  namespace = "xyz.justsoft.video_thumbnail"
  ...
}

[...]

PR #2049, opened 2024-11-18
fix(ui): namespace not specified for video_thumbnail package.

xsahil03x:
Fixes: #2030


pcj:
Would be nice to have new release with this one... `8.2.1`?

Issue #2080, opened 2025-01-11
How to create a new channel with stream_chat_flutter 9.x?


mauriziopinotti:
With stream_chat_flutter 8.x I used to create channels following the example app:

```dart
// in initState()
_channel = StreamChat.of(context).client.channel('messaging');

// then after the user selects the members for the new channel
  _channel = chatState.client.channel('messaging', extraData: {
    'members': [
      ..._selectedUsers.map((e) => e.id),
      chatState.currentUser!.id,
    ],
    'team': myteam,
  });
```

However, this no longer works with 9.x, I get

```
StreamChatNetworkError(code: 4, message: GetOrCreateChannel failed with error: "When using member based IDs specify at least 2 members", statusCode: 400, data: ErrorResponse(code: 4, message: GetOrCreateChannel failed with error: "When using member based IDs specify at least 2 members", statusCode: 400,
```

The example app does not have a create feature, how should channels be created in 9.x?

szechyjs:
I too am having the issue. For me it is happening on the load of the new chat page, before I even call the api to create a new channel. It looks like the root is coming from loading an empty channel?

The create channel code is essentially copied directly from https://github.com/GetStream/flutter-samples/blob/develop/packages/stream_chat_v1/lib/pages/new_chat_screen.dart perhaps updating that to the latest version will solve this issue.

```
[ERROR:flutter/runtime/dart_vm_initializer.cc(40)] Unhandled Exception: StreamChatNetworkError(code: 4, message: GetOrCreateChannel failed with error: "When using member based IDs specify at least 2 members", statusCode: 400, data: ErrorResponse(code: 4, message: GetOrCreateChannel failed with error: "When using member based IDs specify at least 2 members", statusCode: 400, moreInfo: https://getstream.io/chat/docs/api_errors_response))
#0      StreamHttpClient.post (package:stream_chat/src/core/http/stream_http_client.dart:150:7)
<asynchronous suspension>
#1      ChannelApi.queryChannel (package:stream_chat/src/core/api/channel_api.dart:35:22)
<asynchronous suspension>
#2      Channel.query (package:stream_chat/src/client/channel.dart:1662:24)
<asynchronous suspension>
```

mauriziopinotti:
For anyone stumbling upon this issue, I have sort of fixed mine by changing the way I create new channels.

In 8.x I used to follow the official example that does something like

```dart
// Create channel in initState()
  @override
  void initState() {
    super.initState();

[...]

PR #2105, opened 2025-02-11
feat(repo): move sample app to monorepo

xsahil03x:
Moving the [sample_app_v1](https://github.com/GetStream/flutter-samples/tree/develop/packages/stream_chat_v1) for easier maintenance and keeping it up-to-date.


README.md
```diff
@@ -20,8 +20,8 @@ Stream allows developers to rapidly deploy scalable feeds and chat messaging wit
 
 For upgrading from V7 to V8, please refer to the [V8 Migration Guide](https://getstream.io/chat/docs/sdk/flutter/guides/migration_guide_8_0/)
 
-## Sample apps and demos 
-Our team maintains a dedicated repository for full fledged sample applications and demos. Consider checking out [GetStream/flutter-samples](https://github.com/GetStream/flutter-samples) to learn more or get started by looking at our latest [Stream Chat demo](https://github.com/GetStream/flutter-samples/tree/main/packages/stream_chat_v1). 
+## Sample apps and demos
+Check out our [sample_app](sample_app) to see Stream Chat's core features in action. For more sample apps and demos, visit [GetStream/flutter-samples](https://github.com/GetStream/flutter-samples).
```
renefloor, line 24: Don't we want to archive `GetStream/flutter-samples`? This repo is also mentioned on line 67

.gitignore
```diff
@@ -82,13 +82,25 @@ Podfile.lock
...
+**/linux/flutter/ephemeral/
+
 # Generated files
 *.dart.js
 *.info.json
 *.js
 *.js_
 *.js.deps
 *.js.map
+*.wasm
```
renefloor, line 103: Is it not enough to ignore the build folders?

melos.yaml
```diff
@@ -2,7 +2,17 @@ name: stream_chat_flutter
...
+  - packages/*
+  - packages/*/example
+
+categories:
+  sample_app:
+    - sample_app
+  packages:
+    - packages/*
+  examples:
+    - packages/*/example
```
renefloor, line 15: Didn't know about categories yet :)

renefloor (approved):

README.md
xsahil03x: We will archive the sample app but we still have some extra demo apps in the repo. Do we want to archive them all?

.gitignore
xsahil03x: Will these files always be present in build folders? I was not sure so added them specifically and added exceptions for the `sql-wasm.wasm` for the sample app.

README.md
renefloor: The `chatty` app is officially "no longer maintained". The iMessage app could be nice example, but it's based on stream_chat_flutter_core version 3, so would need a bit of work to get up to date.

.gitignore
renefloor: I would expect that when these files are not in the build folder they are useful to keep. But I would have to see which files these are and why they are there.

szechyjs:
Now that this has been merged, can we fix #2080?

PR #2131, opened 2025-03-06
refactor(llc, core, ui): update X-Stream-Client header for enhanced client identification

xsahil03x:


packages/stream_chat/lib/src/core/http/system_environment_manager.dart
```diff
@@ -0,0 +1,60 @@
...
+/// {@endtemplate}
+class SystemEnvironmentManager {
+  /// {@macro systemEnvironmentManager}
+  SystemEnvironmentManager({
+    SystemEnvironment environment = const SystemEnvironment(
+      sdkName: 'stream-chat',
+      sdkIdentifier: 'dart',
+      sdkVersion: PACKAGE_VERSION,
+    ),
+  }) : _environment = environment;
```
renefloor, line 17: We don't want to have people change these values. What do you think of creating 2 constructors like this? (simplified)

```suggestion
  SystemEnvironmentManager.llc() : sdkIdentifier = 'dart';
  SystemEnvironmentManager.core() : sdkIdentifier = 'core';
  SystemEnvironmentManager.flutter() : sdkIdentifier = 'flutter';
```

For Core and Flutter I would add an `AppConfig` parameter, which only contains the app specific settings, such as appName and appVersion. We can make more variables required there.

packages/stream_chat_flutter_core/lib/src/stream_chat_core.dart
```diff
@@ -113,11 +109,80 @@ class StreamChatCoreState extends State<StreamChatCore>
   var _isInForeground = true;
   var _isConnectionAvailable = true;
 
+  Future<SystemEnvironment> get _getSystemEnvironment async {
+    String? osVersion;
+    String? deviceModel;
+
+    try {
+      switch (CurrentPlatform.type) {
```
renefloor, line 117: Following my other comment, I would move this to an `AppConfig` class which can be used both by core and flutter.

packages/stream_chat/lib/src/core/http/system_environment_manager.dart
xsahil03x: I don't think the `llc` should not know about the packages where it's being used.

xsahil03x: What we can do is make the `sdkName` and `sdkVersion` field private and keep the other open. wdyt?

packages/stream_chat_flutter_core/lib/src/stream_chat_core.dart
xsahil03x: We can move the fields in the AppConfig class but we can't move this logic as it's flutter specific and we can't add flutter related dependency in llc

martinmitrevski (approved):
LGTM! ✅

PR #2142, opened 2025-03-13
refactor(llc, ui): move slowMode logic inside StreamMessageInputController

xsahil03x:
Resolves: FLU-68

The PR aims to move the slow mode logic from the `StreamMessageInput` widget to the `StreamMessageInputController` so that it can be directly accessed and handled through the controller.

Additionally, it improves the current cooldown logic in Channel and maintains the cooldown through out the app restart.

https://github.com/user-attachments/assets/a71fdf2e-631d-4f8d-86f4-ebb5bb934f3a

### DIFF
--- a/packages/stream_chat/lib/src/client/channel.dart
+++ b/packages/stream_chat/lib/src/client/channel.dart
@@ -261,7 +261,7 @@ class Channel {
   /// Remaining cooldown duration in seconds for the channel.
   ///
   /// Returns 0 if there is no cooldown active.
-  int get remainingCooldown {
+  int getRemainingCooldown() {
     _checkInitialized();
 
     final cooldownDuration = cooldown;
@@ -283,7 +283,7 @@ class Channel {
     "Use a combination of 'remainingCooldown' and 'currentUserLastMessageAt'",
   )
   DateTime? get cooldownStartedAt {
-    if (remainingCooldown == 0) return null;
+    if (getRemainingCooldown() <= 0) return null;
     return currentUserLastMessageAt;
   }
 
--- a/packages/stream_chat/lib/src/client/channel.dart
+++ b/packages/stream_chat/lib/src/client/channel.dart
@@ -258,8 +258,34 @@ class Channel {
     return state!.channelStateStream.map((cs) => cs.channel?.cooldown ?? 0);
   }
 
+  /// Remaining cooldown duration in seconds for the channel.
+  ///
+  /// Returns 0 if there is no cooldown active.
+  int get remainingCooldown {
+    _checkInitialized();
+
+    final cooldownDuration = cooldown;
+    if (cooldownDuration <= 0) return 0;
+
+    final userLastMessageAt = currentUserLastMessageAt;
+    if (userLastMessageAt == null) return 0;
+
+    if (ownCapabilities.contains(PermissionType.skipSlowMode)) return 0;
+
+    final currentTime = DateTime.timestamp();
+    final elapsedTime = currentTime.difference(userLastMessageAt).inSeconds;
+
+    return max(0, cooldownDuration - elapsedTime);
+  }
+
   /// Stores time at which cooldown was started
-  DateTime? cooldownStartedAt;
+  @Deprecated(
+    "Use a combination of 'remainingCooldown' and 'currentUserLastMessageAt'",
+  )
+  DateTime? get cooldownStartedAt {
+    if (remainingCooldown == 0) return null;
+    return currentUserLastMessageAt;
+  }
 
   /// Channel creation date.
   DateTime? get createdAt {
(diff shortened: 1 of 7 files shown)


packages/stream_chat/lib/src/client/channel.dart
```diff
@@ -258,8 +258,34 @@ class Channel {
...
+
+    final currentTime = DateTime.timestamp();
+    final elapsedTime = currentTime.difference(userLastMessageAt).inSeconds;
+
+    return max(0, cooldownDuration - elapsedTime);
+  }
+
   /// Stores time at which cooldown was started
-  DateTime? cooldownStartedAt;
+  @Deprecated(
```
renefloor, line 282: Is there really a need to deprecate this method? It's still working as expected right?

packages/stream_chat/lib/src/client/channel.dart
```diff
@@ -258,8 +258,34 @@ class Channel {
...
+    "Use a combination of 'remainingCooldown' and 'currentUserLastMessageAt'",
+  )
```
renefloor, line 284: Is there really a need to deprecate this getter? It still works as expected right? And there isn't a lot to maintain here.

packages/stream_chat/lib/src/client/channel.dart
```diff
@@ -258,8 +258,34 @@ class Channel {
     return state!.channelStateStream.map((cs) => cs.channel?.cooldown ?? 0);
   }
 
+  /// Remaining cooldown duration in seconds for the channel.
+  ///
+  /// Returns 0 if there is no cooldown active.
+  int get remainingCooldown {
```
renefloor, line 264: I have the feeling this is quite a heavy getter. Usually when you call a get you expect a simple value.
Maybe we could change it to a method and rename to `getRemainingCooldown` or `calculateRemainingCooldown`? So devs are aware it is not recommended to call it in the build methods.

packages/stream_chat_flutter/lib/src/message_input/stream_message_input.dart
```diff
@@ -527,6 +521,20 @@ class StreamMessageInputState extends State<StreamMessageInput>
...
+
+    WidgetsBinding.instance.endOfFrame.then((_) {
+      if (!mounted) return;
+
+      // Call the listener once to make sure the initial state is reflected
+      // correctly in the UI.
+      _onChangedDebounced.call();
+
+      // Resumes the cooldown if the channel has currently an active cooldown.
+      if (_isEditing case false) {
```
renefloor, line 533: Why not just ` if (!_isEditing)`?

packages/stream_chat_flutter/lib/src/message_input/stream_message_input.dart
```diff
@@ -527,6 +521,20 @@ class StreamMessageInputState extends State<StreamMessageInput>
...
+        final channel = StreamChannel.of(context).channel;
```
renefloor, line 534: I think we should add this to `didChangeDependencies`, for when the channel changes.

`didChangeDependencies` is always called after `initState` so we can remove it here. In `didChangeDependencies` we also have access to context, so I wonder if the `WidgetsBinding.instance.endOfFrame` is needed there.

packages/stream_chat/lib/src/client/channel.dart
xsahil03x: This isn't being used anymore and has no specific use, it was a SDK specific hack created earlier to store the cooldown start time so that we can resume it when we go back to the channel. This however was not enough to also resume it if we are restarting the app.

packages/stream_chat_flutter/lib/src/message_input/stream_message_input.dart
xsahil03x: We only need to call this once at the start of this widget, didChangeDependencies can get called multiple time in a lifecycle of a widget.

packages/stream_chat/lib/src/client/channel.dart
xsahil03x: Renamed to `getRemainingCooldown()`

packages/stream_chat_flutter/lib/src/message_input/stream_message_input.dart
renefloor: But for example if you would have a master detail view, you could change the channel and still have this state. So you need to get the channel again from the context. You are depending here on the `StreamChannel.of(context)` which gets the current channel from the tree. We have to make sure that if it updates in the tree we also update it here.
I do agree that it would be good to check if the channel actually changed, so maybe store the channelId in the widget state so you can compare if the id of the channel changed and only do this than.

`StreamChannel.of(context)` uses `findAncestorStateOfType` and in the docs there they also recommend to use `findAncestorStateOfType`.

In general I always have the feeling that there is a potential code smell when you need `WidgetsBinding.instance.addPostFrameCallback` or something similar.

packages/stream_chat/lib/src/client/channel.dart
renefloor: I don't know if anybody uses this, but I agree it's probably not a very useful method.

packages/stream_chat_flutter/lib/src/message_input/stream_message_input.dart
xsahil03x: Updated

renefloor (approved):

Issue #2113, opened 2025-02-14
How should I change the dialog background color of StreamGalleryHeader?


Vincent8725:
### Which packages are you using?

stream_chat_flutter

### On what platforms did you experience the issue?

Android

### What version are you using?

9.3.0

### What happened?

After the dialog is opened, the background color is not transparent or translucent.

![Image](https://github.com/user-attachments/assets/bd80944c-205c-4668-8c79-f5ff2e16ec86)

### Steps to reproduce

```bash
1. create conversation
2. send image 
3. click image message
4. click on the top right menu
```

### Relevant log output

```shell

```

### Flutter analyze output

```shell

```

### Flutter doctor output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

DmytroMeleshko:
@Vincent8725 Have you found any workaround?

Issue #2096, opened 2025-02-04
StreamChannel Material color


ihijazi:
### Please select which package this feature is related to

stream_chat_flutter, stream_chat_flutter_core

### Is your feature request related to a problem?

Yes. In a way related to this: https://github.com/GetStream/stream-chat-flutter/issues/2095

### Describe the solution that you'd like

StreamChannel widget's Material to allow for color to be transperant.

### Describe alternatives that you have considered

Creating custom StreamChannel, but that create a lot of dependencies need to change and make use of it especially if StreamChat is not on top of the widgets tree.

Simple ask is to have color parameter inheritable/adjustable for specific use-cases where it doesn't have to follow the full app material theme.

https://github.com/GetStream/stream-chat-flutter/blob/e1ec460b9644446620c692ebd4c2aabe5b956317/packages/stream_chat_flutter_core/lib/src/stream_channel.dart#L471C1-L473C1

### Additional context

One example where this can be useful is within channel, on message actions, to have the background transperant rather than showing solid color. Desired outcome (which I had to customise) is something like this:

<img width="392" alt="Image" src="https://github.com/user-attachments/assets/2032b638-625b-42ac-b17f-00262540fd0e" />

_No response_

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

ihijazi:
.

xsahil03x:
Should be fixed in #2157

PR #2157, opened 2025-03-24
fix(core): unable to pass non initialized channels in StreamChannel

xsahil03x:
Resolves: #2080 , #2096, #2113

* Fixed an issue with not being able to use a non-initialized `Channel` in the `StreamChannel` widget.
* Fixed issues with message action modals background getting covered by Material background color.

### DIFF
--- a/packages/stream_chat_flutter_core/lib/src/stream_channel.dart
+++ b/packages/stream_chat_flutter_core/lib/src/stream_channel.dart
@@ -409,66 +409,57 @@ class StreamChannelState extends State<StreamChannel> {
   /// Reloads the channel with latest message
   Future<void> reloadChannel() => _queryAtMessage(limit: 30);
 
-  late List<Future<bool>> _futures;
+  Future<void> _maybeInitChannel() async {
+    // If the channel doesn't have an ID yet, it hasn't been created on the
+    // server so we don't need to initialize it.
+    if (channel.id == null) return;
 
-  Future<bool> get _loadChannelAtMessage async {
-    try {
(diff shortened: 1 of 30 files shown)


packages/stream_chat_flutter_core/lib/src/stream_channel.dart
```diff
@@ -23,6 +23,14 @@ typedef ErrorWidgetBuilder = Widget Function(
   StackTrace? stackTrace,
 );
 
+Color _getDefaultBackgroundColor(BuildContext context) {
+  final brightness = Theme.of(context).brightness;
+  return switch (brightness) {
+    Brightness.light => const Color(0xfff7f7f8),
+    Brightness.dark => const Color(0xff000000),
+  };
```
renefloor, line 31: Is there a reason not to make it transparent?

xsahil03x: # Before
simulator_screenshot_097BF6A1-EF79-4759-B678-17FBD322FBA5

# After
simulator_screenshot_56B50734-41E8-4865-857D-00CD86A6BBD0

`StreamChannel is used to wrap various modals in the app which requires it to be transparent.`

xsahil03x: Yes, without the material widget, the default loader and error text appears with a black background. Also the error text requires a directionality widget otherwise it appears with yellow lines.

Brazol (approved):
LGTM

PR #2168, opened 2025-04-01
feat(llc): add support for OwnUser.blockedUserIds

xsahil03x:
Resolves: FLU-86

Improves user blocking functionality by updating client state when blocking/unblocking users:
  - `client.blockUser` now updates `currentUser.blockedUserIds` list with newly blocked user IDs.
  - `client.unblockUser` now removes the unblocked user ID from `currentUser.blockedUserIds` list.
  - `client.queryBlockedUsers` now updates `currentUser.blockedUserIds` with the latest blocked users data.

### DIFF
--- a/packages/stream_chat/lib/src/client/client.dart
+++ b/packages/stream_chat/lib/src/client/client.dart
@@ -2,6 +2,7 @@ import 'dart:async';
 
 import 'package:dio/dio.dart';
 import 'package:logging/logging.dart';
+import 'package:meta/meta.dart';
 import 'package:rxdart/rxdart.dart';
 import 'package:stream_chat/src/client/channel.dart';
 import 'package:stream_chat/src/client/retry_policy.dart';
--- a/packages/stream_chat/lib/src/client/client.dart
+++ b/packages/stream_chat/lib/src/client/client.dart
@@ -1451,6 +1451,68 @@ class StreamChatClient {
         ...options,
       });
 
+  final _userBlockLock = Lock();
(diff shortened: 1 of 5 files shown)


packages/stream_chat/lib/src/client/client.dart
```diff
@@ -1995,6 +2045,10 @@ class ClientState {
     channels = channels..remove(channelCid);
   }
 
+  set blockedUserIds(List<String> blockedUserIds) {
+    currentUser = currentUser?.copyWith(blockedUserIds: blockedUserIds);
+  }
```
renefloor, line 2050: Should this be a public setter? I don't think it makes a lot of sense to set this from somewhere, because setting it on a client feels to me like it should be send to the backend.

xsahil03x: It's part of the `ClientState` so shouldn't be a problem

renefloor: ClientState is exposed to the public world, so adding it here makes it part of our public API. If we don't want or expect other devs to use this method we should just make it private. Everything we make public from this package is public api.

If it's public for testing we should just mark it with `visibleForTesting`

```suggestion
  @visibleForTesting
  set blockedUserIds(List<String> blockedUserIds) {
    currentUser = currentUser?.copyWith(blockedUserIds: blockedUserIds);
  }
```

renefloor (approved):

PR #2188, opened 2025-04-07
refactor!(llc, core): Improve query sorts and provide defaults.

xsahil03x:

This pull request aims to redesign sorting to be type-safe with IDE assistance. The new system uses `ComparableField` to handle comparison between different value types and introduces extension types like `UserSortKey` and `ChannelSortKey` for better discoverability. This ensures the same sorting logic works client-side and server-side while catching errors at compile-time instead of runtime.


packages/stream_chat/lib/src/core/models/comparable_field.dart
```diff
@@ -0,0 +1,46 @@
...
+      _ => 0 // All comparisons were equal or incomparable types
+    };
+  }
+}
+
+/// A mixin that provides a way to access comparable fields by string keys.
+///
+/// Classes that implement this mixin can be used in sorting operations
+/// where the sort key is determined at runtime.
+mixin ComparableFieldProvider {
```
renefloor, line 40: I would just make this an `abstract interface` class, because a mixin usually has functionality and not just an interface. A [mixin](https://dart.dev/language/mixins) you also use with `with`, for example `class Musician extends Performer with Musical`, but in our code it's always used with `implements`, so just as an interface.
```suggestion
abstract interface class ComparableFieldProvider {
```

packages/stream_chat/lib/src/core/api/requests.dart
renefloor, line 1: Can you extract all the sorting logic out of requests.dart? For example sort_option.dart. The file is getting quite big and this is now not only usable for requests, but also in other parts of the code.

xsahil03x: Done

renefloor (approved):

PR #2204, opened 2025-04-10
feat(llc, core, persistence): channel pinning and archiving

renefloor:
# Submit a pull request

## CLA

This adds channel pinning and archiving.
When doing an api call to pin/archive or unpin/unarchive the membership object on the channel is updated. In the UI we have to sort again with the pinned status in mind, so we cannot easily just put channels at the top anymore when they are changed.

Fixes FLU-20
Fixes FLU-21


packages/stream_chat/lib/src/core/models/channel_state.dart
```diff
@@ -84,6 +84,7 @@ class ChannelState implements ComparableFieldProvider {
       ChannelSortKey.updatedAt => channel?.updatedAt,
       ChannelSortKey.lastMessageAt => channel?.lastMessageAt,
       ChannelSortKey.memberCount => channel?.memberCount,
+      ChannelSortKey.pinnedAt => membership?.pinnedAt,
```
xsahil03x, line 87: We should also provide support for archivedAt

packages/stream_chat_flutter_core/lib/src/stream_channel_list_controller.dart
```diff
@@ -139,6 +139,7 @@ class StreamChannelListController extends PagedValueNotifier<int, Channel> {
         presence: presence,
         paginationParams: PaginationParams(limit: limit),
       )) {
+        _sortChannels(channels);
```
xsahil03x, line 142: We no longer need to do this as we already sort the value while setting it.

packages/stream_chat_flutter_core/lib/src/stream_channel_list_controller.dart
```diff
@@ -170,6 +171,7 @@ class StreamChannelListController extends PagedValueNotifier<int, Channel> {
       )) {
         final previousItems = previousValue.items;
         final newItems = previousItems + channels;
+        _sortChannels(newItems);
```
xsahil03x, line 174: We no longer need to do this as we already sort the value while setting it.

packages/stream_chat_flutter_core/lib/src/stream_channel_list_controller.dart
```diff
@@ -194,6 +196,12 @@ class StreamChannelListController extends PagedValueNotifier<int, Channel> {
     }
   }
 
+  /// Sorts the channels based on [channelStateSort] and sets the [channels].
+  void setAndSortChannels(List<Channel> channels) {
+    _sortChannels(channels);
+    this.channels = channels;
+  }
+
```
xsahil03x, line 204: This now can be removed

packages/stream_chat_flutter_core/lib/src/stream_channel_list_controller.dart
```diff
@@ -294,6 +304,18 @@ class StreamChannelListController extends PagedValueNotifier<int, Channel> {
...
+    if (channelStateSort case final sortOrder?) {
+      channels.sort((a, b) {
+        final stateA = a.state?.channelState;
+        final stateB = b.state?.channelState;
+        if (stateA == null || stateB == null) return 0;
+
+        return sortOrder.compare(stateA, stateB);
+      });
+    }
+  }
```
xsahil03x, line 317: This now can be removed

packages/stream_chat/lib/src/core/api/channel_api.dart
xsahil03x, line 1: I don't think we should provide separate Api's for pin and archive from `channel_api` as it represents the API's available in backend. We can provide updateMemberPartial from here and then can have separate api's for them in client.dart similar to how we use `partialUpdateMessage` for `pinning` and `unpinning` the message.

xsahil03x (approved):

PR #2230, opened 2025-04-28
feat(ui, core, localization): add draft message preview and controller

xsahil03x:
Resolves: FLU-98, FLU-99

This pull request adds UI, Core support for the draft messages feature. 

https://github.com/user-attachments/assets/f342be64-23d7-497b-b55a-45ee0f9c7a97


packages/stream_chat/lib/src/db/chat_persistence_client.dart
```diff
@@ -79,11 +79,9 @@ abstract class ChatPersistenceClient {
...
   });
 
-  /// Get stored [Draft] message by providing channel [cid].
-  Future<Draft?> getDraftMessageByCid(String cid);
-
-  /// Get stored [Draft] message by providing parent message [id].
-  Future<Draft?> getDraftMessageByParentId(String parentId);
+  /// Get stored [Draft] message by providing channel [cid] and a optional
+  /// [parentId] for thread messages.
+  Future<Draft?> getDraftMessageByCid(String cid, {String? parentId});
```
renefloor, line 84: This is a breaking change in the package, but technically adding an abstract method without default implementation is already a breaking change in the interface. I guess we never applied semantic versioning there and just assumed people always use the same version of stream_chat and stream_chat_persistence?

packages/stream_chat_flutter/lib/src/channel/stream_draft_message_preview_text.dart
```diff
@@ -0,0 +1,42 @@
...
+  Widget build(BuildContext context) {
+    final theme = StreamChatTheme.of(context);
+    final colorTheme = theme.colorTheme;
+
+    final previewTextSpan = TextSpan(
+      text: '${context.translations.draftLabel}: ',
+      style: textStyle?.copyWith(
+        fontWeight: FontWeight.bold,
+        color: colorTheme.accentPrimary,
+      ),
```
renefloor, line 29: Do we want to make this more themeable?

packages/stream_chat_flutter/lib/src/message_input/stream_message_input.dart
```diff
@@ -549,11 +551,31 @@ class StreamMessageInputState extends State<StreamMessageInput>
...
+        _draftStreamSubscription = draftStream?.distinct().listen((draft) {
+          // If the draft is removed, reset the controller.
+          if (draft == null) return _effectiveController.reset();
+
+          // Otherwise, update the controller with the draft message.
+          if (draft.message case final draftMessage) {
+            _effectiveController.message = draftMessage.toMessage();
+          }
+        });
+      }
```
renefloor, line 578: I would extract this to a separate method, the initState is already very long

packages/stream_chat_flutter/lib/src/channel/stream_draft_message_preview_text.dart
xsahil03x: Nah, it's a simple Text and can be easily replaced using the `subtitleBuilder` of `ChannelListTile`

packages/stream_chat_flutter/lib/src/message_input/stream_message_input.dart
xsahil03x: Done

renefloor (approved):

PR #2256, opened 2025-05-16
fix(samples): mark ITSAppUsesNonExemptEncryption to false

xsahil03x:

Added the `ITSAppUsesNonExemptEncryption` key with a value of `false`, indicating that the app does not use non-exempt encryption.

This is needed otherwise, AppStore complains about `Missing Compliance status in TestFlight`.

### DIFF
--- a/sample_app/ios/Runner/Info.plist
+++ b/sample_app/ios/Runner/Info.plist
@@ -70,5 +70,7 @@
 		<key>NSAllowsArbitraryLoads</key>
 		<true/>
 	</dict>
+	<key>ITSAppUsesNonExemptEncryption</key>
+	<false/>
 </dict>
 </plist>


Brazol (approved):

sample_app/ios/Runner/Info.plist
```diff
@@ -70,5 +70,7 @@
 		<key>NSAllowsArbitraryLoads</key>
 		<true/>
 	</dict>
+	<key>ITSAppUsesNonExemptEncryption</key>
+    <false/>
```
xsahil03x, line 74:
```suggestion
	<key>ITSAppUsesNonExemptEncryption</key>
    	<false/>
```

Brazol (approved):

PR #2258, opened 2025-05-19
fix(ui): fix MessageInput to use with non-initialized channel

xsahil03x:

**Fixed `StreamMessageInput` exception**: Updated the `StreamMessageInput` widget to handle cases where the `Channel` is not initialized. This change introduces a conditional logic using a `switch` statement to ensure the widget behaves correctly even when `channel.state` is `null`. (`packages/stream_chat_flutter/lib/src/message_input/stream_message_input.dart`)


Brazol (approved):
LGTM

PR #2269, opened 2025-05-30
feat(llc, core): message reminders

xsahil03x:
# Submit a pull request
Linear: FLU-143

## Docs PR
https://github.com/GetStream/docs-content/pull/380

## Screenshots / Videos

https://github.com/user-attachments/assets/eb439fba-d166-4257-b354-76f5f31a6322

## Summary by CodeRabbit

- **New Features**
  - Introduced message reminders, allowing users to bookmark messages or set reminders for later within channels.
  - Added a Reminders tab in the app to view, filter, edit, and delete message reminders.
  - Users can manage reminders directly from message actions, including setting, editing, or removing reminders.
  - Reminders are highlighted in message lists and provide visual indicators for scheduled or saved-for-later messages.
  - Added reminder dialogs for creating or editing reminder times with preset options.
  - Added a controller to manage paginated lists of message reminders with real-time event updates.
  - Added new reminder-related event types for creation, update, deletion, and due notifications.

- **Enhancements**
  - Reminders are synchronized in real-time and update automatically based on reminder events.
  - Reminders can be filtered by categories such as all, overdue, upcoming, scheduled, or saved for later.
  - Improved sorting options with configurable null ordering for reminders and other data.
  - Enhanced message rendering to integrate reminder UI and actions within chat messages.
  - Updated channel sorting behavior to place channels with null pinned dates last.

- **Documentation**
  - Updated changelogs to announce the new message reminders feature and reminder list controller.

### DIFF
--- a/packages/stream_chat_flutter_core/test/stream_message_reminder_list_controller_test.dart
+++ b/packages/stream_chat_flutter_core/test/stream_message_reminder_list_controller_test.dart
@@ -240,8 +240,8 @@ void main() {
 
       expect(result, isTrue);
       expect(
-        controller.value.asSuccess.items.first.remindAt?.millisecondsSinceEpoch,
-        equals(newRemindAt.millisecondsSinceEpoch),
+        controller.value.asSuccess.items.any((r) => r.remindAt == newRemindAt),
+        isTrue,
       );
       expect(controller.value.asSuccess.items.length, equals(reminders.length));
     });


xsahil03x:
@coderabbitai review

packages/stream_chat/lib/src/core/api/sort_order.dart
```diff
@@ -13,6 +13,15 @@ part 'sort_order.g.dart';
...
 typedef SortOrder<T extends ComparableFieldProvider> = List<SortOption<T>>;
 
+/// Defines how null values should be ordered in a sort operation.
+enum NullOrdering {
+  /// Null values are treated as less than any non-null value.
+  nullsFirst,
+
+  /// Null values are treated as greater than any non-null value.
+  nullsLast,
+}
```
renefloor, line 23: First and Last might be a bit confusing, because it's also depending on the direction of the sorting.

The default settings actually put it last right?
```dart
    this.direction = SortOption.DESC,
    this.nullOrdering = NullOrdering.nullsFirst,
```

Maybe `least` and `most` instead of first and last?

xsahil03x: I am following the naming convention used in postgres. This also matches our backend.

packages/stream_chat/lib/src/core/models/message_reminder.dart
```diff
@@ -0,0 +1,164 @@
...
+
+  /// Creates a copy of this [Draft] with specified attributes overridden.
+  MessageReminder copyWith({
+    String? channelCid,
+    ChannelModel? channel,
+    String? messageId,
+    Message? message,
+    String? userId,
+    User? user,
+    Object? remindAt = _nullConst,
```
renefloor, line 91: I wonder if there is a way to do this while we can keep the type safety, something like 
```dart
    DateTime? remindAt = _nullConst,
```

Maybe by doing something like this?

```dart
class _NullConstDateTime extends DateTime {
  _NullConstDateTime() : super(0);
}
```

xsahil03x: Freezed and similar packages use inheritence to overcome this, but i feel that will be a overkill. Maybe sometime later we move on to use [dart_mappable](https://pub.dev/packages/dart_mappable) to generate our models.

packages/stream_chat/lib/src/core/api/sort_order.dart
xsahil03x: [fix: fix sort option null order](https://github.com/GetStream/stream-chat-flutter/pull/2269/commits/ea548399730f8a486429bfe207f977449e0a1431)

packages/stream_chat/lib/src/core/api/sort_order.dart
```diff
@@ -104,13 +106,15 @@ class SortOption<T extends ComparableFieldProvider> {
...
   ///
   /// ```dart
   /// final sortOption = SortOption<ChannelState>("last_message_at");
   /// final sortedChannels = channels.sort(sortOption.compare);
   /// ```
-  int compare(T a, T b) => direction * comparator(a, b);
+  int compare(T a, T b) => comparator(a, b);
 
   /// Returns a comparator function for sorting objects of type T.
   @JsonKey(includeToJson: false, includeFromJson: false)
```
renefloor, line 120: This changes breaks it if you have your own comparator (line 118/122)

```dart
if (_comparator case final comparator?) return comparator;
```

Because now direction has no effect anymore and you have to apply this yourself.

Maybe we could change that into something like this:
```dart
if (_comparator case final comparator?) return (a, b) => direction*comparator(a,b);
```

That still wouldn't apply `nullOrdering`, but at least it wouldn't break current behaviour.

xsahil03x: Updated

PR #2289, opened 2025-06-26
fix(ui): improve scroll to bottom and floating date divider

xsahil03x:

This PR addresses two issues:

1.  **Scroll to Bottom:** The logic for determining if the last item is fully visible was flawed, causing the "scroll to bottom" button to appear incorrectly in some cases. This has been fixed by checking if the last item's leading edge is greater than or equal to 0.

2.  **Floating Date Divider:** The floating date divider was not correctly calculating the date for messages in case the message widget exceeds the viewport height. This has been fixed by adjusting the index calculations to account for these items and ensuring that the divider only considers actual message widgets.

Additionally, the `isThreadConversation` property in `FloatingDateDivider` has been deprecated as it's no longer used. The utility functions `getTopElementIndex` and `getBottomElementIndex` have also been updated to ignore items that are not actually rendering anything in the viewport.

## Screenshots / Videos

| Before | After |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/394e7774-547c-4aef-90e1-b1fa7fea5be7" /> | <video src="https://github.com/user-attachments/assets/9e89c6ec-75ff-4f90-9d06-56f8765a5e9a" /> |

### DIFF
--- a/packages/stream_chat_flutter/lib/src/message_list_view/floating_date_divider.dart
+++ b/packages/stream_chat_flutter/lib/src/message_list_view/floating_date_divider.dart
@@ -16,11 +16,13 @@ class FloatingDateDivider extends StatelessWidget {
     required this.reverse,
     required this.messages,
     required this.itemCount,
+    @Deprecated('No longer used, Will be removed in future versions.')
     this.isThreadConversation = false,
     this.dateDividerBuilder,
   });
 
   /// true if this is a thread conversation
+  @Deprecated('No longer used, Will be removed in future versions.')
   final bool isThreadConversation;
 
   // ignore: public_member_api_docs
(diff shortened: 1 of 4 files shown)


packages/stream_chat_flutter/lib/src/message_list_view/floating_date_divider.dart
```diff
@@ -47,28 +49,37 @@ class FloatingDateDivider extends StatelessWidget {
...
+        // Parent Message
+        if (index == itemCount - 1) return const Empty();
+        // Header Builder
+        if (index == itemCount - 2) return const Empty();
+        // Top Loader Builder
+        if (index == itemCount - 3) return const Empty();
+        // Bottom Loader Builder
+        if (index == 1) return const Empty();
+        // Footer Builder
+        if (index == 0) return const Empty();
```
renefloor, line 71: I would create a different method for it, this is adding quite some noise

Just like this:
```dart
if(!_shouldShowDivider(index)) return const Empty();
```

with:
```dart
bool _shouldShowDivider(int index){
  return index != null &&
    
    // We can only calculate the date divider if the element is a message
    // widget and not one of the special items.

    // Parent Message
    index != itemCount - 1 &&
    // Header Builder
    index != itemCount - 2 &&
    // Top Loader Builder
    index != itemCount - 3 &&
    // Bottom Loader Builder
    index != 1 &&
    // Footer Builder
    index != 0;
}

PR #2320, opened 2025-07-22
fix(llc, persistence): fix draft message persistence and retrieval issues

xsahil03x:
Fix Draft Message Persistence and Retrieval Issues:
1. **Persistence Issue**: When channel states were updated, drafts that were removed from the state weren't being deleted from the database, leading to orphaned draft records.

2. **Retrieval Issue**: Channel drafts were being incorrectly attached to ALL messages in a channel instead of only attaching thread drafts to their respective parent messages.


xsahil03x:
@coderabbitai review

PR #2384, opened 2025-09-19
fix(ui): ensure dialog confirmation for delete/flag actions

xsahil03x:
# Submit a pull request

Fixes: FLU-254

This PR updates the `_showDeleteMessageDialog` and `_showFlagMessageDialog` methods in `MessageWidget`.

Previously, if the user dismissed the confirmation dialog (e.g., by tapping outside), the `showDialog` method would return `null`. The existing checks (`confirmDelete == false` and `confirmFlag == false`) would evaluate to `true` in this scenario, causing the delete/flag action to proceed unintentionally.

The checks have been changed to `confirmDelete != true` and `confirmFlag != true` to ensure that the actions only proceed if the user explicitly confirms by tapping the "Yes" button (which returns `true`)

## Summary by CodeRabbit

- Bug Fixes
  - Prevented delete and flag actions from triggering when their confirmation dialogs are dismissed or canceled. Actions now execute only after explicit confirmation, reducing accidental deletions or flags.

- Documentation
  - Updated CHANGELOG with an Upcoming Beta section noting the fix for delete/flag dialogs requiring explicit confirmation.


xsahil03x:
@coderabbit review

renefloor (approved):

PR #2422, opened 2025-10-21
feat(ui): Add floating date divider builder.

AleksandrLarin-Ardas:
## CLA

- [x] Code changes are tested

## Summary by CodeRabbit

* **New Features**
  * StreamMessageListView now supports an optional customization for the floating date divider, letting apps supply a custom widget for the floating divider.
  * If no custom floating divider is provided, the control falls back to the existing date divider appearance to preserve backward compatibility.
  * The change is exposed through the component's public API so it can be configured by consumers.

### DIFF
--- a/packages/stream_chat_flutter/CHANGELOG.md
+++ b/packages/stream_chat_flutter/CHANGELOG.md
@@ -7,10 +7,6 @@
 
 ## 9.18.0
 
-✅ Added
-
--  Added `StreamMessageListView.floatingDateDividerBuilder` in order to specifically customize the floating date divider.
-
 🐞 Fixed
 
 - Fixed `StreamMessageListView` not marking thread messages as read when scrolled to the bottom of the list.


packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
```diff
@@ -852,7 +857,7 @@ class _StreamMessageListViewState extends State<StreamMessageListView> {
               reverse: widget.reverse,
               itemPositionListener: _itemPositionListener.itemPositions,
               messages: messages,
-              dateDividerBuilder: widget.dateDividerBuilder,
+              floatingDateDividerBuilder: widget.floatingDateDividerBuilder,
```
xsahil03x: Hey @AleksandrLarin-Ardas , this is going to be a breaking change, we don't have to rename the param in `FloatingDateDivider`. We can reuse the same param name.

Also, for backward compatability we should use the previous builder as a fallback.

```suggestion
              dateDividerBuilder: widget.floatingDateDividerBuilder ?? widget.dateDividerBuilder
```

packages/stream_chat_flutter/lib/src/message_list_view/floating_date_divider.dart
xsahil03x: These changes are not required, we can reuse the existing builder.

PR #2522, opened 2026-03-03
feat(ui): channel list item

renefloor:
# Submit a pull request

Linear: FLU-379

## CLA

This updates to the new channel list design.

Relevant PR: https://github.com/GetStream/stream-core-flutter/pull/61

## Screenshots / Videos

| Before | After |
| --- | --- |
| image | image |


packages/stream_chat_flutter/lib/src/scroll_view/channel_scroll_view/stream_channel_list_tile.dart
```diff
@@ -1,41 +1,40 @@
...
+    this.subtitleTrailing,
     this.trailing,
     this.onTap,
     this.onLongPress,
-    this.tileColor,
-    this.visualDensity = VisualDensity.compact,
-    this.contentPadding = const EdgeInsets.symmetric(horizontal: 8),
-    this.unreadIndicatorBuilder,
     this.sendingIndicatorBuilder,
-    this.selected = false,
```
renefloor, line 37: The `selected` is used in the `stream_chat_flutter/example` app, but I don't know if we should still support that here.

packages/stream_chat_flutter/lib/src/scroll_view/channel_scroll_view/stream_channel_list_tile.dart
```diff
@@ -62,199 +80,175 @@ class StreamChannelListTile extends StatelessWidget {
...
-
-    final channelPreviewTheme = StreamChannelPreviewTheme.of(context);
-    final streamChatTheme = StreamChatTheme.of(context);
-    final streamChat = StreamChat.of(context);
+    final colorScheme = context.streamColorScheme;
+    final textTheme = context.streamTextTheme;
 
-    final leading = this.leading ?? StreamChannelAvatar(channel: channel);
+    // TODO: Make this configurable
+    const showMuteIconInTitle = true;
```
renefloor: We have to see how/where we want configurations like this.

packages/stream_chat_flutter/lib/src/components/avatar/stream_channel_avatar.dart
```diff
@@ -82,13 +82,25 @@ class StreamChannelAvatar extends StatelessWidget {
...
+          if (channel.isDistinct && users.length == 2) {
+            final otherUser = users.firstWhere(
+              (u) => u.id != currentUserId,
+              orElse: () => users.first,
+            );
+            return StreamUserAvatar(
+              user: otherUser,
+              size: _avatarSizeForAvatarGroupSize(effectiveSize),
+              // TODO: make this configurable when the online state is shown.
+              showOnlineIndicator: otherUser.online,
```
renefloor, line 97: We have to see how/where we want configurations like this.

packages/stream_chat_flutter/test/src/message_modal/goldens/ci/stream_message_reactions_modal_dark.png
renefloor, line 1: @xsahil03x This is going to be re-made anway right?

packages/stream_chat_flutter/lib/src/scroll_view/channel_scroll_view/stream_channel_list_item.dart
```diff
@@ -108,27 +108,117 @@ class StreamChannelListItem extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    final channelState = channel.state!;
+    final builder = StreamComponentFactory.of(context).extension<StreamChannelListItemProps>();
```
xsahil03x: we should use `context.channelListItem` instead, similar to composer

packages/stream_chat_flutter/pubspec.yaml
```diff
@@ -80,6 +77,7 @@ dev_dependencies:
   path: ^1.8.3
   path_provider_platform_interface: ^2.0.0
   plugin_platform_interface: ^2.0.0
+  theme_extensions_builder_annotation: ^7.1.0
```
xsahil03x: do we need this? and this should be in `dependencies` instead and builder should be added here

renefloor: I just moved the channel list theme from core. Let's think about it when we refactor the stream chat theme. For now I think it's useful, but indeed I'll move it.

packages/stream_chat_flutter/lib/src/scroll_view/channel_scroll_view/stream_channel_list_item.dart
renefloor: Do you want 1 dart file for all builder extensions? I think we should keep it internal right?

xsahil03x: yeah, we can do it. also if we are planning to keep it internal then i am not against using the `.of` method

packages/stream_chat_flutter/lib/src/components/avatar/stream_channel_avatar.dart
renefloor: We might want a boolean in the `StreamOnlineIndicatorTheme` to show/hide offline states. Or maybe an enum.

```dart
enum StreamOnlineIndicatorBehaviour {
 alwaysShow,
 onlyShowOnline,
}
```

packages/stream_chat_flutter/lib/src/scroll_view/channel_scroll_view/stream_channel_list_item.dart
renefloor: Made an in between, just an extension for the long `StreamComponentFactory.of(context).extension`:
https://github.com/GetStream/stream-chat-flutter/pull/2522/commits/596f1301367f1b4cc43d16fda1ad3bdeee2cf938

xsahil03x (approved):

PR #2590, opened 2026-04-01
fix(ui): hide mark as unread for own messages

Brazol:
# Submit a pull request

Linear: FLU-463

[in-depth Slack thread](https://getstream.slack.com/archives/C0A3JU6C6G3/p1774948953492979)

## CLA

- [ ] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required).
- [ ] The code changes follow best practices

## Screenshots / Videos

| Before | After |
| --- | --- |
| img | img |

### DIFF
--- a/packages/stream_chat_flutter/lib/src/message_action/message_actions_builder.dart
+++ b/packages/stream_chat_flutter/lib/src/message_action/message_actions_builder.dart
@@ -148,7 +148,7 @@ class StreamMessageActionsBuilder {
       );
     }
 
-    if (canReceiveReadEvents) {
+    if (canReceiveReadEvents && !isSentByCurrentUser) {
       StreamContextMenuAction<MessageAction> markUnreadAction() {
         return StreamContextMenuAction(
           value: MarkUnread(message: message),
--- a/packages/stream_chat_flutter/lib/src/message_action/message_actions_builder.dart
+++ b/packages/stream_chat_flutter/lib/src/message_action/message_actions_builder.dart
@@ -164,7 +164,7 @@ class StreamMessageActionsBuilder {
       }
       // If the message is in the channel view, only other user messages can be
       // marked unread.
-      else if (!isSentByCurrentUser && (!isThreadMessage || canShowInChannel)) {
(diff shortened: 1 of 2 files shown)


packages/stream_chat_flutter/lib/src/message_action/message_actions_builder.dart
xsahil03x: We already kind of had this check here. Can you verify if this is an actual issue?

Brazol: right, it was reported by Alexey and it did show up on the actions menu when tested on own message that has a thread. So question is should it always be hidden or do we keep the current logic of showing it on the parent message

xsahil03x: afaik, IOS have the same [UX](https://github.com/GetStream/stream-chat-swift/blob/develop/Sources/StreamChatUI/MessageActionsPopup/ChatMessageActionsVC.swift#L128) right now. But i think its better to not show it even in own parent message. So basically this is correct and we should just remove it from the internal if check.

xsahil03x: Lets also update the tests to verify this is working correctly.

Brazol: done

packages/stream_chat_flutter/lib/src/message_action/message_actions_builder.dart
```diff
@@ -164,7 +164,7 @@ class StreamMessageActionsBuilder {
       }
       // If the message is in the channel view, only other user messages can be
       // marked unread.
```
xsahil03x, line 167: We should also update the inline doc now that we no longer check for currentUser in this if block. and also same for the parent message.

xsahil03x (approved):
Approved with a comment

PR #2594, opened 2026-04-02
fix(ui): hide message replies in thread view and other improvements

xsahil03x:
Fixes: FLU-431
https://github.com/user-attachments/assets/d20facee-07aa-4d99-94bb-778c26c09f94


packages/stream_chat_flutter/lib/src/message_widget/components/stream_message_header.dart
```diff
@@ -19,100 +19,112 @@ import 'package:stream_core_flutter/stream_core_flutter.dart';
...
-  required BuildContext context,
-  required Message message,
-  VoidCallback? onViewChannelTap,
-}) {
-  final translations = context.translations;
-  final icons = context.streamIcons;
-  final textTheme = context.streamTextTheme;
-  final colorScheme = context.streamColorScheme;
-  final crossAxisAlignment = StreamMessageLayout.crossAxisAlignmentOf(context);
+class StreamMessageHeader extends core.NullableStatelessWidget {
```
renefloor: Don't you want to rename this to annotation or something? Or does that conflict too much with the core component?

renefloor (approved):

packages/stream_chat_flutter/lib/src/message_widget/components/stream_message_header.dart
xsahil03x: Renamed

PR #2615, opened 2026-04-13
fix(ui): increase max width for ephemeral Giphy messages

xsahil03x:
## Screenshots / Videos

| Before | After |
| --- | --- |
| simulator_screenshot_85F6C8F6-AF68-4119-A9E8-A530C0113561 | simulator_screenshot_F7A3A666-8A5C-4934-89BA-AFC941827DBD |


renefloor (commented):
I wonder if we should fix this differently. Now it's working for english, but does it work well for all languages? Maybe the actions should go into 2 lines if it doesn't fit?

xsahil03x:
> I wonder if we should fix this differently. Now it's working for english, but does it work well for all languages? Maybe the actions should go into 2 lines if it doesn't fit?

Should be handled now, Using Wrap instead of Row now. Also removed the constraints so that max width can be based on the max of the children width.

renefloor (approved):

PR #2642, opened 2026-05-05
feat(UI): rename message input

renefloor:
# Submit a pull request

## CLA

This renames the StreamMessageInput to StreamMessageComposer and the StreamChatMessageComposer to StreamChatMessageInput. 

It also fixes the component builders and the attachment sync.

## Summary by CodeRabbit

* **New Features**
  * New message composer experience: StreamMessageComposer (full-featured) and StreamChatMessageInput (UI-only) with integrated voice recording and improved send controls.
  * New customizable input center and input field components for easier UI overrides.

* **Updates**
  * Examples, docs, and migration guides updated to the new composer naming and usage.
  * Attachment picker and placeholder behaviors refined; keyboard/quoted-message handling improved.


migrations/redesign/message_composer.md
```diff
@@ -420,6 +422,32 @@ The following public widgets are provided as building blocks for custom attachme
...
+| `DefaultStreamMessageComposerInput` widget | `DefaultStreamMessageComposerInputCenter` |
+| `MessageComposerInputProps` (the props class carrying text-field configuration) | `MessageComposerInputCenterProps` |
+| `messageComposerInput` builder key in `streamChatComponentBuilders` | `messageComposerInputCenter` |
+
+The name `StreamMessageComposerInput` is now used for the **input container** (the widget that assembles the header, leading, center, and trailing sub-components). Its corresponding props class is `MessageComposerInputProps` and its builder key in `streamChatComponentBuilders` is `messageComposerInput`.
+
+`inputHeader` / `StreamMessageComposerInputHeader` / `MessageComposerInputHeaderProps` / `messageComposerInputHeader` are **unchanged**.
+
+---
+
```
renefloor: Do we need a migration list compared to last beta?

xsahil03x: nah, it should be fine. just mention it in changelog.

packages/stream_chat_flutter/lib/src/components/message_composer/message_composer_input.dart
xsahil03x, line 1: do we need this widget? cant we compose them directly in the message composer? If not maybe we should call it message composer container instead?

packages/stream_chat_flutter/lib/src/message_input/stream_chat_message_input.dart
xsahil03x, line 1: The name of this widget does not go well with the other widgets we have here

packages/stream_chat_flutter/lib/src/components/message_composer/message_composer_input.dart
renefloor: Image

It's this part. The naming was agreed with the others and also makes sense. The Input has an input leading, input center and input trailing. If we would just have container it would be confusing if it's only this part of the whole thing. 

Having a different factory for it is also useful for if you don't want this layout, you can update the separate parts without needing a completely custom composer with all the logic.

renefloor: image

I think it's good if we have something like this in the docs, but a bit nicer

packages/stream_chat_flutter/lib/src/message_input/stream_chat_message_input.dart
renefloor: Also not meant to be used outside of the sdk, maybe we should make it internal. I think some of the other sdk's also named it input right?

packages/stream_chat_flutter/lib/src/components/message_composer/message_composer_input.dart
xsahil03x: yes, but the input center is the actual input right. the leading and trailing are respect to the input not center. If we call it center then what actually is the input?

packages/stream_chat_flutter/lib/src/message_input/stream_chat_message_input.dart
xsahil03x: I meant the chat prefix in the name but lets keep it internal then, maybe move to the message input directory to be clear.

packages/stream_chat_flutter/lib/src/components/message_composer/message_composer_input.dart
renefloor: Simulator Screenshot - iPhone 17 Pro Max - 2026-05-12 at 16 07 33

The whole red thing is the input. The blue box is the inputCenter, which is the actual input field. Yellow (send button) is inputTrailing and orange is the input header. So we have leading/center/trailing/header inside the input. The green one is the composer leading, so we have leading/input/trailing for the composer. The complete input is the thing with the rounded rectangle.

> the leading and trailing are respect to the input not center.

Not exactly sure what you mean with this, but I hope I explained it.

packages/stream_chat_flutter/lib/src/message_input/stream_chat_message_input.dart
renefloor: I moved it and removed it from the export. Have a direct src import now in the docs, that works fine for creating the screenshot

xsahil03x (approved):

PR #2651, opened 2026-05-17
perf(repo): cut StreamMessageListView UI-thread work under flood load

xsahil03x:
Closes [FLU-482](https://linear.app/stream/issue/FLU-482/reduce-unnecessary-message-list-rebuilds-in-flutter)

## Summary

Cross-cutting performance push targeting `StreamMessageListView` under high-throughput / livestream conditions (100+ msg/s). The series spans the core layer and the UI library — each layer's hot paths were profiled against a synthetic flood and an on-device DevTools CPU profile.

Each commit is self-contained and ships independent of the others; they are stacked by layer (SDK → core → UI) so a reviewer can read top-down or bottom-up.

## What changed, by layer

**`stream_chat` (SDK):**
- Added `.distinct()` on `readStream`, `currentUserReadStream`, and `unreadCountStream` so listeners only fire on real value changes. Previously the per-getter-access `.map` chains emitted identical values repeatedly, which made every downstream `StreamBuilder` resubscribe-and-rebuild at flood rate (~100×/sec for `unreadCountStream`). `currentUserReadStream` also tightens its upstream — switches on `mapNotNull(it?.id).distinct()` instead of re-emitting on every user-stream tick.

**`stream_chat_flutter_core`:**
- `BetterStreamBuilder` correctness fixes — handle null payloads and identity-equal emissions correctly so downstream rebuilds gate on real value changes (separate from the SDK-side `.distinct()` work above; this layer is the consumer).
- `MessageListCore` caches its `messagesStream` resolution so it doesn't re-evaluate the stream chain on every `setState`, and reduces per-emission filter+reverse work.

**`stream_chat_flutter` (UI):**
- **SPL `_keyToIndexMap` bounded to a sliding window** (matches Compose's `LazyLayoutNearestRangeState`: window 30, ±100 extras, bucket-anchored). Cache invalidates only on `itemCount` or window-bucket cross. **Profile: 6.00 % → 0.17 %** (–333 ms over 5.6 s).
- `HashMap` + `.clear()`-and-refill instead of allocating a fresh `<Object, int>{}` literal per cache miss — drops `LinkedHashMap`'s insertion-order metadata we don't use.
- New `ItemScrollController.isScrolling` + `isScrollingListenable` API (mirror of Compose's `LazyListState.isScrollInProgress`). `StreamMessageListView` reads this directly instead of maintaining its own `_isUserScrolling` flag fed by bubbling `ScrollNotification`s (which got stuck `true` on drag-without-fling).
- `_messageNewListener` consumes `channel.membership` directly instead of building the freshened `members` list per row and scanning with `firstWhereOrNull`. **Per-row, per-build savings**: one cached field read in place of an `O(N)` list allocation + scan + an extra `StreamChat.of(context)` lookup.
- Anchor preservation across data mutations via `itemKeyBuilder` + saved-anchor-key state, invalidated on explicit `jumpTo`/`scrollTo`.
- Fit-anchor fallback in `UnboundedRenderViewport` for short-content lists.
- Scroll-to-bottom FAB now reads from `_unreadState` (a `ValueListenable`) instead of a per-getter-call `Stream` chain — single subscription, single source of truth.
- `_buildScrollToBottom`'s membership check (`members.any(...)`) was a brittle proxy for "has unread"; replaced with `unreadCount > 0` (the actual source of truth) which also fixes a silent bug in large team channels where the current user was paginated out of `state.members` and the badge disappeared.
- `ItemScrollController.scrollTo` gained sensible defaults (`Duration(milliseconds: 250)`, `Curves.fastOutSlowIn` — closest built-in match to a critically-damped spring at ~99 % settling time). All four caller sites (auto-scroll-on-new-message, FAB tap, unread tap, quoted-message tap) now use them.
- `onQuotedMessageTap` else-branch fix: previously did `loadChannelAtMessage(id)` then set `initialIndex = 21` on State fields the SPL doesn't read post-mount — so out-of-window quoted-message taps silently failed to scroll. Now does the paginate, waits one frame for the rebuild, re-scans `this.messages` for the real index, and scrolls.
- `initialIndex` / `initialAlignment` are `late final` — `getInitialIndex` (`O(N)` over channel messages) runs once on first build's read, never recomputed. Dead reassignments in `didChangeDependencies` and `scrollToBottomDefaultTapAction` removed.

## Profile numbers

On-device CPU profile (Android profile build, ~5.6 s window, synthetic flood @ 100 msg/s, rendering the regular `StreamMessageWidget` — not a lightweight stand-in). Numbers reflect only what's in this branch; SDK ingest work (e.g. `ChannelClientState.updateMessage` / `_listenMessageNew`) is *not* included here.

| Frame | Before | After | Δ |
|---|---|---|---|
| `_PositionedListState.build` + `_rebuildKeyIndexMap` | 342 ms (6.0 %) | 9 ms (0.17 %) | **–333 ms / –97 %** |
| `ChannelClientState.members` chain (3 frames) | 286 ms (5.0 %) | 0 ms | **gone** |
| `_StreamMessageListViewState.buildMessage` chain (cumulative)* | ~1 961 ms | ~1 197 ms | **–764 ms / –39 %** |

\*Sum of five overlapping parent-chain frames (`_PositionedListState.build.<closure>`, `_buildSeparatedListElement`, `_buildItem`, `_StreamMessageListViewState.buildMessage`, the message-list-view item-builder closure). This row double-counts the savings already captured in the first two rows; the two non-overlapping rows alone contribute **~619 ms / ~11 %** of UI-thread time.

Estimated FPS impact during flood on a 60–90 Hz panel: **+5 – +8 fps** depending on baseline frame time, with materially lower jank %.

## New tests

Five new test files, all passing locally:

- `test/scrollable_positioned_list/is_scrolling_test.dart` — `isScrolling` API across drag, animated `scrollTo`, `jumpTo`, and the listenable contract.
- `test/scrollable_positioned_list/item_key_builder_test.dart` — anchor preservation, element reuse, separated lists, reanchor guard, mid-gesture preservation, null opt-out (13 tests across 6 groups).
- `test/scrollable_positioned_list/viewport_fit_anchor_test.dart` — short-content layout-time fit-anchor fallback.
- `test/src/message_list_view/mark_read_test.dart` — `markReadWhenAtTheBottom` trigger gates + FAB visibility.
- `test/src/message_list_view/message_list_view_benchmark_test.dart` — perf regression gate (cold build, batch insert, trickled stream) with generous budgets sized for CI variance.

`melos analyze` and `melos format` both clean across all packages.

## Test plan

- [ ] `melos analyze` — clean
- [ ] `melos format` — no changes
- [ ] `flutter test packages/stream_chat/` — passes
- [ ] `flutter test packages/stream_chat_flutter_core/` — passes
- [ ] `flutter test packages/stream_chat_flutter/test/scrollable_positioned_list/` — passes
- [ ] `flutter test packages/stream_chat_flutter/test/src/message_list_view/` — passes
- [ ] Manual: run the benchmark page in `--profile` mode, flood a 1 k-message channel at 100 msg/s on a 90 Hz device, verify HUD shows fps ≥ baseline + ~5 and jank% materially lower.
- [ ] Manual: tap a quoted message that's outside the loaded window — verify list scrolls to the message after pagination.
- [ ] Manual: tap scroll-to-bottom FAB while channel `isUpToDate = false` — verify reload completes and list scrolls to bottom.

## Summary by CodeRabbit

* **New Features**
  * Stable keyed identity for scrollable lists to preserve item state across rebuilds.
  * Scrolling controller now exposes "isScrolling" state and improved scroll-to behavior.

* **Bug Fixes**
  * Prevented thread view crashes when parent messages are missing.
  * Fixed unread-badge and mark-read race conditions; improved quoted-thread navigation resilience.
  * Hardening of position reporting to avoid transient layout errors.

* **Performance**
  * Read/unread streams dedupe identical successive values to avoid redundant UI updates.
  * Message list filtering and caching optimized.

* **Tests**
  * Added comprehensive widget tests for scrolling, keyed lists, anchoring, and read/mark-read behaviors.

* **Documentation**
  * Changelogs updated for the release.

[Review Change Stack](https://app.coderabbit.ai/change-stack/GetStream/stream-chat-flutter/pull/2651?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


packages/stream_chat_flutter/lib/scrollable_positioned_list/src/scrollable_positioned_list.dart
```diff
@@ -347,9 +471,103 @@ class _ScrollablePositionedListState extends State<ScrollablePositionedList>
...
+  int _findFirstVisibleItemIndex(Object? Function(int) keyBuilder) {
+    final savedKey = _lastKnownFirstItemKey!;
+    final savedIndex = _lastKnownFirstItemIndex;
+    // Fast path: same key still at the same index. Skips the O(N) scan
+    // in the common case where items haven't shifted.
+    if (savedIndex < widget.itemCount && keyBuilder(savedIndex) == savedKey) {
+      return savedIndex;
+    }
+    // Full key→index scan.
+    for (var i = 0; i < widget.itemCount; i++) {
```
Brazol, line 500: This fires each time the itemCount changes (so each time new message arrives?) and performs a full scan? Might be another performance bottleneck if I get it right.

Brazol (approved):

packages/stream_chat_flutter/lib/scrollable_positioned_list/src/scrollable_positioned_list.dart
xsahil03x: Good catch — yes, it scans on every new message. In practice it's small (didn't show up in profiling even under flood load), and for livestream-style channels where this could matter most, the new               
  `maximumMessageLimit` prop on StreamMessageListView caps the list size so the scan stays bounded.

PR #2654, opened 2026-05-18
chore(ui): drop the bundled IANA TLD list

xsahil03x:
Closes [FLU-490](https://linear.app/stream/issue/FLU-490/remove-hardcoded-tld-data-from-flutter-source)

## Summary

Deletes `packages/stream_chat_flutter/lib/src/message_input/tld.dart` — ~1,500 lines of bundled IANA TLDs that lived only to second-guess `_urlRegex` in `StreamMessageInput`.

## Why

- The URL regex already requires a TLD-shaped suffix (`\.[a-z]{2,}\b`); the allowlist check on top added nothing the server doesn't already validate.
- The bundled list went stale whenever IANA registered a new TLD.
- Other Stream SDKs do **not** bundle a TLD list:
  - **JS**: configurable URL regex, no TLD allowlist
  - **Android**: platform `Linkify` / `PatternsCompat.AUTOLINK_WEB_URL`
  - **Swift**: `NSDataDetector`

## What changes

- Relax `_urlRegex` TLD bracket from `\.[a-z]{2,4}\b` → `\.[a-z]{2,}\b`. Side effect: newer long TLDs (`.academy`, `.foundation`, etc.) now detect correctly — they were silently dropped before.
- Drop `.split('.').last.isValidTLD()` in the two filter call sites in `stream_message_input.dart`. The regex itself is enough.
- Delete `tld.dart`.

The file was internal (`src/`, never exported via the public barrel), so no behaviour visible to consumers changed beyond the regex relaxation.

## Test plan

- [x] `dart analyze` clean on `packages/stream_chat_flutter/lib`
- [x] No tests covered the removed `isValidTLD()` extension (confirmed via grep)
- [ ] Manual: paste a URL with a long TLD like `https://example.academy` into a `StreamMessageInput` and confirm an OG preview is triggered (previously silently filtered out)

## Summary by CodeRabbit

* **Refactor**
  * Improved link detection in the message input to recognize a broader range of domain endings and to rely on the configured preview filter for Open Graph previews.

* **Bug Fixes**
  * When link posting is disabled, messages containing detected links are now consistently blocked.

* **Chores**
  * Removed legacy TLD validation code.

[Review Change Stack](https://app.coderabbit.ai/change-stack/GetStream/stream-chat-flutter/pull/2654?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Brazol (approved):
LGTM

PR #2655, opened 2026-05-18
feat(core): bound channel state with opt-in message retention cap

xsahil03x:
## CLA

- [x] Code changes are tested

Adds an opt-in retention cap so long-lived sessions on busy channels stop growing `ChannelClientState.messages` without bound, and stop running the per-emission filter+reverse pass over an ever-growing list.

Three layers, all additive — every new prop defaults to off, so no consumer sees a change unless they opt in.

**LLC (`stream_chat`)**

- `ChannelClientState.pruneOldest(int)` drops the oldest messages and keeps the most recent ones. No-op when the channel isn't up to date.

**Core (`stream_chat_flutter_core`)**

- `StreamChannel.pruneOldest(int)` wraps the data-layer call and resets the top-pagination tracker so older messages can be re-fetched after a prune.
- `MessageListCore.maximumMessageLimit` (default `null` = no pruning) and `retentionTrimBuffer` (default `30`) — auto-prune fires when a new message arrives or the user paginates bottom and the count exceeds `limit + buffer`. **Top pagination, edits, reactions, deletions, jump-to-message, and threads do not trigger pruning.** Decision logic is encapsulated in `MessageRetentionGate` (`@visibleForTesting`).
- `defaultMessageFilter` now accepts an optional `currentUserId`.

**UI (`stream_chat_flutter`)**

- `StreamMessageListView.maximumMessageLimit` passes through to `MessageListCore`.

### Test plan

- [x] `stream_chat`: 7 unit tests for `ChannelClientState.pruneOldest` (basic trim, stream emission, no-op guards, field preservation, no spurious emission).
- [x] `stream_chat_flutter_core`: 30 widget + unit tests for `MessageListCore` retention wiring and `MessageRetentionGate` semantics — covers new-message-arrives, top pagination, edits on tail and non-tail, thread mode, channel/thread transitions, trim buffer respect.
- [x] Smoke-tested via the benchmark page (`sample_app/lib/benchmark`) — flood mode + manual scrolling under cap.


Brazol (approved):
LGTM

PR #2679, opened 2026-05-21
perf(persistence): Reduce the number of read message per channel from DB when paginating (part 1)

VelikovPetar:
# Submit a pull request
Linear: Part one of: FLU-485

## CLA

- Read only the messages matching the `PaginationParams` from DB when calling `MessageDao.getMessagesByCid` instead of reading all messages for the channel and applying pagination in memory.
- Read only the reactions matching the `userId` from DB when calling `ReactionDao.getReactionsByUserId` instead of reading all reactions for the message and filtering in memory.
- Read only the reactions matching the `userId` from DB when calling `PinnedMessageReactionDao.getReactionsByUserId` instead of reading all reactions for the message and filtering in memory.

## Screenshots / Videos
_NA_

## Testing
Apply the given patch and run the test suites:
- `get_messages_by_cid_parity.dart` - parity test check affirming no regressions in the new SQL first implementation against the old implementation
- `get_messages_by_cid_bench_test.dart` - bench test comparing the performance of the new vs old implementations

<details>

<summary>Patch</summary>

```
Subject: [PATCH] refactor(dao): Benchmark
---
Index: packages/stream_chat_persistence/test/src/dao/get_messages_by_cid_parity_test.dart
===================================================================
diff --git a/packages/stream_chat_persistence/test/src/dao/get_messages_by_cid_parity_test.dart b/packages/stream_chat_persistence/test/src/dao/get_messages_by_cid_parity_test.dart
new file mode 100644
--- /dev/null	(date 1779744805256)
+++ b/packages/stream_chat_persistence/test/src/dao/get_messages_by_cid_parity_test.dart	(date 1779744805256)
@@ -0,0 +1,283 @@
+import 'package:flutter_test/flutter_test.dart';
+import 'package:stream_chat/stream_chat.dart';
+import 'package:stream_chat_persistence/src/dao/dao.dart';
+import 'package:stream_chat_persistence/src/db/drift_chat_database.dart';
+
+import '../../stream_chat_persistence_client_test.dart';
+
+void main() {
+  late DriftChatDatabase database;
+  late MessageDao messageDao;
+
+  setUp(() {
+    database = testDatabaseProvider('testUserId');
+    messageDao = database.messageDao;
+  });
+
+  tearDown(() async {
+    await database.disconnect();
+  });
+
+  // Seeds a channel with `count` messages that exercise every hydration path
+  // `_messageFromJoinRow` touches: user attribution, latest/own reactions,
+  // quoted messages, polls, and a channel-level draft. Monotonic 1-second
+  // offsets on `createdAt` because Drift stores DateTime as integer Unix
+  // seconds — sub-second offsets collapse onto the same tick.
+  Future<void> seedRichMessages(String cid, int count) async {
+    final channels = [ChannelModel(cid: cid)];
+    final dbUser = User(id: 'testUserId'); // matches the DB's _userId
+    final otherUsers = List.generate(count, (i) => User(id: 'otherUser$i'));
+    final allUsers = [dbUser, ...otherUsers];
+    final baseTime = DateTime.now();
+
+    final poll = Poll(
+      id: 'poll0',
+      name: 'Pick one',
+      options: [
+        PollOption(id: 'opt0', text: 'A'),
+        PollOption(id: 'opt1', text: 'B'),
+      ],
+      // `createdById` must reference an existing user — `PollDao._pollFromJoinRow`
+      // uses `readTable(users)` (not `readTableOrNull`) on a LEFT JOIN, which
+      // throws if there's no matching row. Pre-existing PollDao quirk; out of
+      // scope for the pagination-pushdown work.
+      createdById: dbUser.id,
+    );
+
+    final messages = List.generate(
+      count,
+      (i) => Message(
+        id: 'msg$i',
+        type: 'regular',
+        user: allUsers[i % allUsers.length],
+        text: 'Hello $i',
+        createdAt: baseTime.add(Duration(seconds: i)),
+        updatedAt: baseTime.add(Duration(seconds: i)),
+        // Every 3rd message (i ≥ 3) quotes the message 2 positions earlier.
+        quotedMessageId: (i >= 3 && i % 3 == 0) ? 'msg${i - 2}' : null,
+        // Every 5th message attaches to a poll.
+        pollId: (i % 5 == 0) ? 'poll0' : null,
+      ),
+    );
+
+    // Reactions populate both `latestReactions` (any user) and
+    // `ownReactions` (matching the DB user). Mix on each message so the two
+    // lists end up non-identical, which surfaces any divergence in hydration.
+    final reactions = <Reaction>[
+      for (var i = 0; i < count; i++) ...[
+        if (i.isEven)
+          Reaction(
+            type: 'like',
+            messageId: 'msg$i',
+            user: dbUser,
+            createdAt: baseTime.add(Duration(seconds: i)),
+          ),
+        Reaction(
+          type: 'love',
+          messageId: 'msg$i',
+          user: otherUsers[i % otherUsers.length],
+          createdAt: baseTime.add(Duration(seconds: i)),
+        ),
+      ],
+    ];
+
+    final draft = Draft(
+      channelCid: cid,
+      createdAt: baseTime,
+      message: DraftMessage(
+        id: 'draft0',
+        text: 'Unsent reply',
+      ),
+    );
+
+    await database.userDao.updateUsers(allUsers);
+    await database.channelDao.updateChannels(channels);
+    await database.pollDao.updatePolls([poll]);
+    await messageDao.updateMessages(cid, messages);
+    await database.reactionDao.updateReactions(reactions);
+    await database.draftMessageDao.updateDraftMessages([draft]);
+  }
+
+  // Builds a structural fingerprint of a Message that captures every field
+  // the two implementations should agree on after hydration. Used in place
+  // of `==` because `Reaction` (and friends) don't extend `Equatable`, so
+  // identity-based equality on nested lists fails for instances built by
+  // separate calls.
+  String fingerprintMessage(Message m) {
+    String reactionFp(Reaction r) => '${r.type}@${r.user?.id ?? "-"}';
+    String reactionListFp(List<Reaction>? rs) =>
+        '[${(rs ?? const []).map(reactionFp).join(",")}]';
+
+    return [
+      'id=${m.id}',
+      'text=${m.text ?? ""}',
+      'user=${m.user?.id ?? "-"}',
+      'createdAt=${m.createdAt.toUtc().toIso8601String()}',
+      'latest=${reactionListFp(m.latestReactions)}',
+      'own=${reactionListFp(m.ownReactions)}',
+      'quoted=${m.quotedMessage?.id ?? "-"}',
+      'poll=${m.poll?.id ?? "-"}',
+      'draft=${m.draft?.message.id ?? "-"}',
+    ].join(' | ');
+  }
+
+  group('getMessagesByCid: full data parity (legacy vs SQL pushdown)', () {
+    const cid = 'test:Cid';
+    const n = 30;
+
+    Future<void> assertParity(String label, PaginationParams? p) async {
+      await seedRichMessages(cid, n);
+
+      final legacy = await messageDao.getMessagesByCidLegacy(
+        cid,
+        messagePagination: p,
+      );
+      final pushdown = await messageDao.getMessagesByCid(
+        cid,
+        messagePagination: p,
+      );
+
+      expect(
+        pushdown.length,
+        legacy.length,
+        reason: 'list lengths differ for "$label"',
+      );
+      expect(
+        pushdown.map(fingerprintMessage).toList(),
+        equals(legacy.map(fingerprintMessage).toList()),
+        reason: 'message data parity broken for "$label"',
+      );
+    }
+
+    test('no pagination', () => assertParity('no pagination', null));
+
+    test(
+      'limit only',
+      () => assertParity('limit: 10', const PaginationParams(limit: 10)),
+    );
+
+    test(
+      'lessThan + limit',
+      () => assertParity(
+        'lessThan: msg25, limit: 10',
+        const PaginationParams(limit: 10, lessThan: 'msg25'),
+      ),
+    );
+
+    test(
+      'greaterThan + limit',
+      () => assertParity(
+        'greaterThan: msg5, limit: 10',
+        const PaginationParams(limit: 10, greaterThan: 'msg5'),
+      ),
+    );
+
+    test(
+      'lessThan + greaterThan + limit',
+      () => assertParity(
+        'lessThan: msg25, greaterThan: msg5, limit: 10',
+        const PaginationParams(
+          limit: 10,
+          lessThan: 'msg25',
+          greaterThan: 'msg5',
+        ),
+      ),
+    );
+  });
+
+  group('getMessagesByCid: tied createdAt parity (legacy vs SQL pushdown)', () {
+    const cid = 'test:Cid';
+
+    // Seeds five messages so three of them (`msg_tieA`, `msg_tieB`, `msg_tieC`)
+    // share an identical `createdAt`. Insertion order matches id-lex order so
+    // legacy's SQL tie-break (rowid order on `ORDER BY createdAt ASC`) lines
+    // up with the pushdown's `(createdAt, id)` ordering — without that
+    // alignment, parity would fail on within-page row order, not on the
+    // cursor logic this group is meant to cover.
+    Future<void> seedTiedMessages() async {
+      final user = User(id: 'testUserId'); // matches DB's _userId
+      final tie = DateTime.now();
+      final earlier = tie.subtract(const Duration(seconds: 1));
+      final later = tie.add(const Duration(seconds: 1));
+
+      Message build(String id, DateTime t) => Message(
+            id: id,
+            type: 'regular',
+            user: user,
+            createdAt: t,
+            updatedAt: t,
+            text: id,
+          );
+
+      await database.userDao.updateUsers([user]);
+      await database.channelDao.updateChannels([ChannelModel(cid: cid)]);
+      await messageDao.updateMessages(cid, [
+        build('msg_pre', earlier),
+        build('msg_tieA', tie),
+        build('msg_tieB', tie),
+        build('msg_tieC', tie),
+        build('msg_post', later),
+      ]);
+    }
+
+    Future<void> assertParity(String label, PaginationParams? p) async {
+      await seedTiedMessages();
+
+      final legacy = await messageDao.getMessagesByCidLegacy(
+        cid,
+        messagePagination: p,
+      );
+      final pushdown = await messageDao.getMessagesByCid(
+        cid,
+        messagePagination: p,
+      );
+
+      expect(
+        pushdown.length,
+        legacy.length,
+        reason: 'list lengths differ for "$label"',
+      );
+      expect(
+        pushdown.map(fingerprintMessage).toList(),
+        equals(legacy.map(fingerprintMessage).toList()),
+        reason: 'message data parity broken for "$label"',
+      );
+    }
+
+    test(
+      'lessThan cursor at tie sibling',
+      () => assertParity(
+        'lessThan: msg_tieB',
+        const PaginationParams(limit: 100, lessThan: 'msg_tieB'),
+      ),
+    );
+
+    test(
+      'greaterThan cursor at tie sibling',
+      () => assertParity(
+        'greaterThan: msg_tieB',
+        const PaginationParams(limit: 100, greaterThan: 'msg_tieB'),
+      ),
+    );
+
+    test(
+      'lessThan + greaterThan straddling the tie group',
+      () => assertParity(
+        'lessThan: msg_tieC, greaterThan: msg_tieA',
+        const PaginationParams(
+          limit: 100,
+          lessThan: 'msg_tieC',
+          greaterThan: 'msg_tieA',
+        ),
+      ),
+    );
+
+    test(
+      'lessThan + limit that splits the tie group',
+      () => assertParity(
+        'lessThan: msg_tieC, limit: 2',
+        const PaginationParams(limit: 2, lessThan: 'msg_tieC'),
+      ),
+    );
+  });
+}
Index: packages/stream_chat_persistence/test/src/benchmark/get_messages_by_cid_bench_test.dart
===================================================================
diff --git a/packages/stream_chat_persistence/test/src/benchmark/get_messages_by_cid_bench_test.dart b/packages/stream_chat_persistence/test/src/benchmark/get_messages_by_cid_bench_test.dart
new file mode 100644
--- /dev/null	(date 1779743905552)
+++ b/packages/stream_chat_persistence/test/src/benchmark/get_messages_by_cid_bench_test.dart	(date 1779743905552)
@@ -0,0 +1,237 @@
+import 'package:drift/drift.dart';
+import 'package:drift/native.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:stream_chat/stream_chat.dart';
+import 'package:stream_chat_persistence/src/dao/dao.dart';
+import 'package:stream_chat_persistence/src/db/drift_chat_database.dart';
+
+/// Counts SELECT statements and rows returned through them. Used to
+/// instrument the legacy and SQL-pushdown implementations of
+/// `getMessagesByCid` for head-to-head comparison.
+class _CountingInterceptor extends QueryInterceptor {
+  int selectCount = 0;
+  int rowsReturned = 0;
+
+  void reset() {
+    selectCount = 0;
+    rowsReturned = 0;
+  }
+
+  @override
+  Future<List<Map<String, Object?>>> runSelect(
+    QueryExecutor executor,
+    String statement,
+    List<Object?> args,
+  ) async {
+    final result = await executor.runSelect(statement, args);
+    selectCount++;
+    rowsReturned += result.length;
+    return result;
+  }
+}
+
+typedef _BenchResult = ({
+  List<String> messageIds,
+  int selectCount,
+  int rowsReturned,
+  int medianMicros,
+});
+
+void main() {
+  late DriftChatDatabase database;
+  late MessageDao messageDao;
+  late _CountingInterceptor interceptor;
+
+  setUp(() {
+    interceptor = _CountingInterceptor();
+    final executor = NativeDatabase.memory().interceptWith(interceptor);
+    database = DriftChatDatabase('testUserId', executor);
+    messageDao = database.messageDao;
+  });
+
+  tearDown(() async {
+    await database.disconnect();
+  });
+
+  // Seeds `count` messages with strictly monotonic `createdAt`. Drift stores
+  // DateTime as integer Unix seconds by default, so the offset must be at
+  // least 1 second per row to avoid tie collapses.
+  Future<void> seedMessages(String cid, int count) async {
+    final channels = [ChannelModel(cid: cid)];
+    final users = List.generate(count, (i) => User(id: 'user$i'));
+    final baseTime = DateTime.now();
+    final messages = List.generate(
+      count,
+      (i) => Message(
+        id: 'msg$i',
+        type: 'regular',
+        user: users[i],
+        text: 'Hello $i',
+        createdAt: baseTime.add(Duration(seconds: i)),
+        updatedAt: baseTime.add(Duration(seconds: i)),
+      ),
+    );
+    await database.userDao.updateUsers(users);
+    await database.channelDao.updateChannels(channels);
+    await messageDao.updateMessages(cid, messages);
+  }
+
+  Future<_BenchResult> runBench(
+    Future<List<Message>> Function() fn, {
+    int warmups = 2,
+    int iterations = 10,
+  }) async {
+    // Warmup runs prime the JIT / SQLite query cache. Their interceptor
+    // counts are discarded by the reset before each timed iteration below.
+    for (var i = 0; i < warmups; i++) {
+      await fn();
+    }
+
+    final timings = <int>[];
+    List<Message>? lastResult;
+    for (var i = 0; i < iterations; i++) {
+      interceptor.reset();
+      final sw = Stopwatch()..start();
+      lastResult = await fn();
+      sw.stop();
+      timings.add(sw.elapsedMicroseconds);
+    }
+    timings.sort();
+    final median = timings[timings.length ~/ 2];
+
+    return (
+      messageIds: lastResult!.map((m) => m.id).toList(),
+      selectCount: interceptor.selectCount,
+      rowsReturned: interceptor.rowsReturned,
+      medianMicros: median,
+    );
+  }
+
+  void printTable(
+    String scenario,
+    int n,
+    _BenchResult legacy,
+    _BenchResult pushdown,
+  ) {
+    String improvement(int oldV, int newV) =>
+        newV == 0 ? '—' : '${(oldV / newV).toStringAsFixed(2)}×';
+
+    String pad(Object v, [int width = 10]) => v.toString().padRight(width);
+
+    final parity = pushdown.messageIds.toString() ==
+            legacy.messageIds.toString()
+        ? 'OK'
+        : 'MISMATCH';
+
+    // ignore: avoid_print
+    print('''
+
+Scenario: $scenario  (N=$n, P=${pushdown.messageIds.length})
+                 ${pad('OLD')}${pad('NEW')}Improvement (old/new)
+SELECT calls     ${pad(legacy.selectCount)}${pad(pushdown.selectCount)}${improvement(legacy.selectCount, pushdown.selectCount)}
+Rows fetched     ${pad(legacy.rowsReturned)}${pad(pushdown.rowsReturned)}${improvement(legacy.rowsReturned, pushdown.rowsReturned)}
+Time (us, med)   ${pad(legacy.medianMicros)}${pad(pushdown.medianMicros)}${improvement(legacy.medianMicros, pushdown.medianMicros)}
+Result parity    $parity
+''');
+  }
+
+  group('getMessagesByCid: legacy vs SQL pushdown', () {
+    const cid = 'test:Cid';
+    const n = 200;
+
+    Future<void> runScenario(String label, PaginationParams? p) async {
+      await seedMessages(cid, n);
+
+      final legacy = await runBench(
+        () => messageDao.getMessagesByCidLegacy(cid, messagePagination: p),
+      );
+      final pushdown = await runBench(
+        () => messageDao.getMessagesByCid(cid, messagePagination: p),
+      );
+
+      expect(
+        pushdown.messageIds,
+        equals(legacy.messageIds),
+        reason: 'parity broken for "$label"',
+      );
+      expect(
+        pushdown.selectCount,
+        lessThanOrEqualTo(legacy.selectCount),
+        reason: 'SQL pushdown issued more SELECTs than legacy for "$label"',
+      );
+      expect(
+        pushdown.rowsReturned,
+        lessThanOrEqualTo(legacy.rowsReturned),
+        reason: 'SQL pushdown materialized more rows than legacy for "$label"',
+      );
+
+      printTable(label, n, legacy, pushdown);
+    }
+
+    test('no pagination', () => runScenario('no pagination', null));
+
+    test(
+      'limit: 30',
+      () => runScenario('limit: 30', const PaginationParams(limit: 30)),
+    );
+
+    test(
+      'lessThan + limit (scroll up)',
+      () => runScenario(
+        'lessThan: msg175, limit: 30',
+        const PaginationParams(limit: 30, lessThan: 'msg175'),
+      ),
+    );
+
+    test(
+      'greaterThan + limit (scroll down)',
+      () => runScenario(
+        'greaterThan: msg25, limit: 30',
+        const PaginationParams(limit: 30, greaterThan: 'msg25'),
+      ),
+    );
+
+    test(
+      'lessThan + greaterThan + limit',
+      () => runScenario(
+        'lessThan: msg175, greaterThan: msg25, limit: 30',
+        const PaginationParams(
+          limit: 30,
+          lessThan: 'msg175',
+          greaterThan: 'msg25',
+        ),
+      ),
+    );
+  });
+
+  group('getMessagesByCid: stress', () {
+    const cid = 'test:Cid';
+
+    test(
+      '1000 messages, limit: 30',
+      () async {
+        await seedMessages(cid, 1000);
+
+        final legacy = await runBench(
+          () => messageDao.getMessagesByCidLegacy(
+            cid,
+            messagePagination: const PaginationParams(limit: 30),
+          ),
+        );
+        final pushdown = await runBench(
+          () => messageDao.getMessagesByCid(
+            cid,
+            messagePagination: const PaginationParams(limit: 30),
+          ),
+        );
+
+        expect(pushdown.messageIds, equals(legacy.messageIds));
+        expect(pushdown.selectCount, lessThanOrEqualTo(legacy.selectCount));
+        expect(pushdown.rowsReturned, lessThanOrEqualTo(legacy.rowsReturned));
+
+        printTable('stress: 1000 messages, limit 30', 1000, legacy, pushdown);
+      },
+      // skip: 'Stress test — slow, runnable on demand by removing this skip',
+    );
+  });
+}
Index: packages/stream_chat_persistence/lib/src/dao/message_dao.dart
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/packages/stream_chat_persistence/lib/src/dao/message_dao.dart b/packages/stream_chat_persistence/lib/src/dao/message_dao.dart
--- a/packages/stream_chat_persistence/lib/src/dao/message_dao.dart	(revision 2e719f4c1eae81b2e66be248df64f480c501f522)
+++ b/packages/stream_chat_persistence/lib/src/dao/message_dao.dart	(date 1779744559662)
@@ -1,4 +1,7 @@
+import 'dart:math';
+
 import 'package:drift/drift.dart';
+import 'package:flutter/foundation.dart';
 import 'package:stream_chat/stream_chat.dart';
 import 'package:stream_chat_persistence/src/db/drift_chat_database.dart';
 import 'package:stream_chat_persistence/src/entity/messages.dart';
@@ -161,6 +164,78 @@
     }
     return msgList;
   }
+
+  /// Pre-SQL-pushdown reference implementation of [getMessagesByCid]. Fetches
+  /// every cached message for the channel, hydrates each row, then trims the
+  /// result in Dart. Kept only as the head-to-head baseline for the
+  /// `get_messages_by_cid_bench_test.dart` benchmark — remove once we no
+  /// longer need behavioral parity proof.
+  @visibleForTesting
+  Future<List<Message>> getMessagesByCidLegacy(
+    String cid, {
+    bool fetchDraft = true,
+    PaginationParams? messagePagination,
+  }) async {
+    final query = select(messages).join([
+      leftOuterJoin(_users, messages.userId.equalsExp(_users.id)),
+      leftOuterJoin(
+        _pinnedByUsers,
+        messages.pinnedByUserId.equalsExp(_pinnedByUsers.id),
+      ),
+    ])
+      ..where(messages.channelCid.equals(cid))
+      ..where(messages.parentId.isNull() | messages.showInChannel.equals(true))
+      ..orderBy([OrderingTerm.asc(messages.createdAt)]);
+
+    final result = await query.get();
+    if (result.isEmpty) return [];
+
+    final msgList = await Future.wait(
+      result.map(
+            (row) => _messageFromJoinRow(
+          row,
+          fetchDraft: fetchDraft,
+        ),
+      ),
+    );
+
+    if (msgList.isNotEmpty) {
+      var lessThanResolved = false;
+      if (messagePagination?.lessThan != null) {
+        final lessThanIndex = msgList.indexWhere(
+              (m) => m.id == messagePagination!.lessThan,
+        );
+        if (lessThanIndex != -1) {
+          msgList.removeRange(lessThanIndex, msgList.length);
+          lessThanResolved = true;
+        }
+      }
+      var greaterThanResolved = false;
+      if (messagePagination?.greaterThan != null) {
+        final greaterThanIndex = msgList.indexWhere(
+              (m) => m.id == messagePagination!.greaterThan,
+        );
+        if (greaterThanIndex != -1) {
+          // Exclude the cursor row itself — mirrors the pushdown's strict
+          // `id > cursor` predicate (vs. legacy's inclusive `removeRange(0, idx)`).
+          msgList.removeRange(0, greaterThanIndex + 1);
+          greaterThanResolved = true;
+        }
+      }
+      if (messagePagination?.limit != null) {
+        final limit = messagePagination!.limit;
+        // Forward pagination — resolved `greaterThan` cursor and no
+        // `lessThan` cursor — takes the first N strictly after the cursor.
+        // Otherwise return the last N (closest to a `lessThan` cursor or the
+        // channel tail).
+        final isForwardPagination = greaterThanResolved && !lessThanResolved;
+        return isForwardPagination
+            ? msgList.take(limit).toList()
+            : msgList.skip(max(0, msgList.length - limit)).toList();
+      }
+    }
+    return msgList;
+  }
 
   /// Returns all the messages of a channel by matching
   /// [Messages.channelCid] with [parentId]

```

</details>

## Summary by CodeRabbit

* **Bug Fixes**
  * Reactions and pinned reactions now filter by user ID in the database; cursor pagination semantics clarified (which cursors are inclusive vs exclusive).

* **Performance**
  * Message pagination applies cursor filters and limits at the database level for more efficient, correctly ordered queries.

* **Tests**
  * Expanded pagination and reaction-by-user tests; timestamps made deterministic for stable, repeatable results.

* **Documentation**
  * Added an "Upcoming Changes" section to the package changelog.

[Review Change Stack](https://app.coderabbit.ai/change-stack/GetStream/stream-chat-flutter/pull/2679?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


packages/stream_chat_persistence/lib/src/dao/message_dao.dart
```diff
@@ -180,44 +187,28 @@ class MessageDao extends DatabaseAccessor<DriftChatDatabase>
...
-        }
-      }
-      if (messagePagination?.limit != null) {
-        return msgList
-            .skip(max(0, msgList.length - messagePagination!.limit))
-            .toList();
-      }
+    if (messagePagination != null) {
+      query.limit(messagePagination.limit);
     }
```
renefloor, line 250: Shouldn't we only limit this when either `lessThanCutoff` or `greaterThanCutoff` is not null? I don't know if it can happen that the target message is not in the cache, but in that case limiting the amount of messages might give issues?

packages/stream_chat_persistence/test/src/dao/message_dao_test.dart
```diff
@@ -25,14 +25,20 @@ void main() {
   }) async {
     final channels = [ChannelModel(cid: cid)];
     final users = List.generate(count, (index) => User(id: 'testUserId$index'));
+    // Strictly monotonic `createdAt` per message so SQL-side pagination
+    // filters (`WHERE createdAt < cutoff`, `ORDER BY createdAt ASC`) can't be
+    // confused by ties. Drift stores `DateTime` as integer Unix seconds by
+    // default, so the offset must be at least 1 second per row — otherwise
+    // sub-second offsets all round-trip onto the same second.
```
renefloor, line 32: I wonder if we should migrate to storing in milliseconds instead of a datetime.

packages/stream_chat_persistence/lib/src/dao/message_dao.dart
VelikovPetar: This actually matches the previous behaviour of the method. The `limit` was always applied, regardless of whether we have a `lessThan`/`greaterThan` ID supplied, and regardless of whether the corresponding message was found in the cache. 
```
if (msgList.isNotEmpty) {
      if (messagePagination?.lessThan != null) {
        final lessThanIndex = msgList.indexWhere(
          (m) => m.id == messagePagination!.lessThan,
        );
        if (lessThanIndex != -1) {
          msgList.removeRange(lessThanIndex, msgList.length);
        }
      }
      if (messagePagination?.greaterThan != null) {
        final greaterThanIndex = msgList.indexWhere(
          (m) => m.id == messagePagination!.greaterThan,
        );
        if (greaterThanIndex != -1) {
          msgList.removeRange(0, greaterThanIndex);
        }
      }
      if (messagePagination?.limit != null) {
        return msgList
            .skip(max(0, msgList.length - messagePagination!.limit))
            .toList();
      }
    }
```
But now that you mentioned it, we actually have several smaller issues that neither the 'old' nor the 'new' implementation cover:
1. Limit is used wrongly in the `greaterThan` path: We always take the last items from the fetch, which means that we skip the first messages which are after the `greaterThan` ID. IMO, we should retrieve the `limit` amount of messages right after the cursor
2. We are missing handling for `greaterThanOrEqual`/`lessThanOrEqual` in the method
3. `greaterThan` case returns the cursor (the pivot message) as well, when it shouldn't 

Should we fix these things in this PR? Or do you think it is better to handle them in a follow up?

packages/stream_chat_persistence/test/src/dao/message_dao_test.dart
VelikovPetar: This is actually something I already noted down to check after we do the offline storage optimisations. 
In my opinion we should do this, as currently it seems like we might be losing some precision on dates (probably not that much of a big deal, but IMO it would be better to be precise as possible). 

I think we should evaluate this separately, because I assume there would be other consequences of this change (maybe more storage consumed?)

renefloor: Yeah, better to do this separately

packages/stream_chat_persistence/lib/src/dao/message_dao.dart
renefloor: I think they can be fixed in this PR. You already change the method quite a bit, so would be good to make it correct.

VelikovPetar: Ok, I will address this!

VelikovPetar: Adressed here: https://github.com/GetStream/stream-chat-flutter/pull/2679/commits/b580ccdc473fc06e537b622915aa7600e4ece87d

packages/stream_chat_persistence/test/src/dao/message_dao_test.dart
xsahil03x: @VelikovPetar , I recently did something to solve this in v10. maybe you can check it.
https://github.com/GetStream/stream-chat-flutter/blob/v10.0.0/packages/stream_chat_persistence/lib/src/db/drift_chat_database.dart#L62-L64

VelikovPetar: Nice, good to know that this is already handled! 
We should consider if it makes sense to also back-port this for the `v9` branch (once it branches out of `master`)

renefloor (approved):

PR #2665, opened 2026-05-19
chore(samples): replace Sentry with Firebase Crashlytics

xsahil03x:
## Summary
- Drop `sentry_flutter` from the sample app: we no longer have access to the Sentry account, so the DSN, init call, log-handler capture, and `runZonedGuarded` wrapper around `main()` are removed.
- Wire **Firebase Crashlytics** in its place. Firebase Core is already initialized for messaging, so the only pubspec/melos additions are:
  - `firebase_crashlytics: ^4.0.0` in `sample_app/pubspec.yaml` and `melos.yaml`
- Dart wiring in `main.dart` (modern post-Flutter-3.3 pattern, no `runZonedGuarded`):
  - `FlutterError.onError` → `FirebaseCrashlytics.instance.recordFlutterFatalError` (release) / `dumpErrorToConsole` (debug)
  - `PlatformDispatcher.instance.onError` → `recordError(..., fatal: true)` (release) / `debugPrint` (debug) — catches every uncaught async error, replacing the old zone-based capture
  - Crashlytics collection disabled in debug builds via `setCrashlyticsCollectionEnabled(!kDebugMode)`
- `_sampleAppLogHandler` in `app.dart` now forwards Stream chat client errors to Crashlytics, attaching the log message as `reason`.

## Platform setup
`flutterfire configure --project=stream-chat-internal --platforms=android,ios,macos,web` was run as part of this PR. The CLI:
- added the `com.google.firebase.crashlytics` Gradle plugin to `android/settings.gradle` and `android/app/build.gradle` (inside the auto-managed FlutterFire Configuration block);
- added the `flutterfire upload-crashlytics-symbols` run-script phase to the iOS and macOS Runner Xcode targets so dSYMs upload automatically on release builds;
- regenerated `google-services.json`, `GoogleService-Info.plist`, `firebase_options.dart`, and `firebase.json` against the current `stream-chat-internal` Firebase project state.

The remaining `Pods/Sentry/*` directory and Sentry framework references in `ios/Runner.xcodeproj/project.pbxproj` are auto-generated by CocoaPods and will disappear on the next `pod install`.

## Test plan
- [ ] `flutter pub get` resolves cleanly in `sample_app/` (verified locally)
- [ ] `dart analyze` passes (verified locally — "No issues found!")
- [ ] `flutter run` (debug) on Android + iOS — app boots, no Crashlytics reports expected with collection disabled
[...]

PR #2653, opened 2026-05-18
perf(llc): faster channel state ingest

xsahil03x:
Closes [FLU-486](https://linear.app/stream/issue/FLU-486/stop-re-sorting-messages-on-every-update)

## Summary

Reduces CPU and allocation cost in `ChannelClientState` on the message-ingest hot path. Targets [Pain Point #6](https://www.notion.so/) from the Flutter tech-debt notion doc, plus a few adjacent cleanups.

Each commit is one focused change so they can be reviewed (and reverted) independently.

| Commit | Change |
|---|---|
| `feat(llc): add SortedListX and ListX list extensions` | Foundational `List` extensions: `merge` (two-pointer + append/prepend fast paths when `compare` provided), `sortedInsert`, `sortedUpsert`, `sortedUpsertAt`, `updateIf`. Signatures mirror `stream_core`'s upcoming `list_extensions` for a clean future swap. |
| `perf(llc): faster channel state ingest` | The actual Pain Point #6 fix. `updateMessage` uses one `indexWhere` + strict-newer short-circuit + `sortedUpsertAt`; `updateChannelState` inlines merges with identity-skip so read/membership-only updates allocate nothing; `_updatePinnedMessages` has a fast-path for the no-pins case; reaction/update handlers dispatch by `parentId` first via `_findMessage`. Legacy `Iterable.merge` extension removed. |
| `fix(llc): preserve null transitions in Channel.currentUserReadStream` | `mapNotNull` swallowed the user-logged-out transition. Switched to `map` so `null` propagates. |
| `refactor(llc): drop the per-update walk for expired CDN attachment URLs` | `_checkExpiredAttachmentMessages` ran on every `updateChannelState` walking every message and parsing every CDN URL. Removed — matches JS SDK; v10's `StreamImageCDN.cacheKey` already keeps the image cache valid across signed-URL rotations. |

## Base branch

This PR is based on `fix/message-list-view-performance-issues` (the UI perf PR), because that branch added `.distinct()` to `Channel.readStream` / `Channel.currentUserReadStream` / `Channel.unreadCountStream` which this work refines. Once the UI PR merges to `master`, this PR's base will auto-shift.

## Test plan

- [x] `flutter test` across `stream_chat` — 1,204 tests pass
- [x] `dart analyze` clean
- [x] No external callers reference the removed symbols (`_checkExpiredAttachmentMessages`, `_updatedMessagesIds`, `Iterable.merge`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

PR #2650, opened 2026-05-15
ci(repo): Resolve local setup related issues

testableapple:
## Summary by CodeRabbit

* **Chores**
  * Stop tracking ephemeral Flutter build files to reduce noise in version control.
  * Switch iOS development builds to automatic code signing and set the development signing identity to streamline local builds.
  * Update Ruby runtimes used by project tooling for Android and iOS workflows to newer versions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/GetStream/stream-chat-flutter/pull/2650?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #2682, opened 2026-05-25
chore(repo): merge master into v10.0.0

xsahil03x:
## Summary

Brings master's perf / security / CI work into the v10 design-refresh branch.

Master is 17 commits ahead of v10.0.0 — this PR carries them into v10 so the upcoming v10 release ships with the latest LLC perf rewrites, the in-flight `queryChannels` dedup, the jose security floor, and the relaxed URL detection. On top of those, this PR ports the same perf approach into v10's design-refresh code (single-message fast path in `_mergeMessagesIntoExisting`, Jiffy removal from message-list predicates), rewrites the date formatter as a clock-driven extension, and overhauls the message-list pagination / scroll / highlight plumbing.

## What landed

**LLC (`stream_chat`)**
- `Client.queryDrafts` now forwards `filter` (#2647).
- `Client.queryChannels` coalesces concurrent identical queries via the new `InFlightCache<K, V>` (#2652).
- `list_extensions.dart` added — `IterableMergeX.merge` (keyed-map merge, unsorted output) plus `SortedListX.mergeSorted` (two-pointer merge requiring a sorted receiver). Duplicate-keyed inputs tolerated (#2660). v10's `IterableMergeExtension` callers route through `merge`; the perf paths use `mergeSorted`. Renaming `SortedListX.merge` (added in 9.24.0) → `mergeSorted` is **breaking** for any direct caller of the former.
- `Channel._mergeMessagesIntoExisting` gains a single-message fast path (`lastIndexWhere` + `sortedUpsertAt`) — skips the full keyed merge for the steady-state new-message case.
- `ChannelClientState._checkExpiredAttachmentMessages` removed (#2653); v10's `StreamImageCDN.cacheKey` already keeps the image cache valid across signed-URL rotations.
- `ChannelClientState.updateChannelState` now identity-short-circuits when `updatedState.messages` is null or the same reference, so downstream `.distinct()` listeners can skip rebuilds.
- Reaction listeners dispatch via `_findMessage` (parentId-aware) while keeping v10's `addMyReaction` / `deleteMyReaction` semantics.
- `jose` floor bumped to 0.3.5+1 for CVE-2026-34240 (#2658).

- `BetterStreamBuilder` correctness fixes: mounted guard, error reporting via `FlutterError.reportError`, identity-equal emission gating (#2651).
- `MessageListCore` caches its `messagesStream` / `_initialMessages` as fields instead of recomputing in `build()` (#2651). `defaultMessageFilter` takes an optional `currentUserId`.
- `StreamChatCore` debounces connectivity events to 3 s (#2652).

**`stream_chat_flutter`**

_Date formatting_
- `date_formatter.dart` rewritten — replaces Jiffy predicates with a `DateTimeComparisonUtils` extension (`isToday`, `isYesterday`, `isWithinLastMinute`, `isWithinLastWeek`, `isInSameYear`, single `isSame(unit:)` entry point with a `DateTimeUnit` enum). Uses `package:clock` for testable time. Cuts ~1.5s on the trace per the latest DevTools profile.
- `formatRecentDateTime` output reworked to a five-state cascade matching the Figma design: `Just now` / `Today at H:mm` / `Yesterday at H:mm` / `Weekday at H:mm` / `MMM do at H:mm`.

_Message-list view internals_
- `mlv_utils.dart`: `_isGroupBoundary`, `_resolveSpacingRules`, and `computeStackPosition` drop Jiffy in favor of the new extension.
- `scrollable_positioned_list/`: master version taken in full. Bounded `_keyToIndexMap`, `isScrolling` / `isScrollingListenable`, `itemKeyBuilder` anchor preservation, fit-anchor fallback in `UnboundedRenderViewport`, sensible defaults on `scrollTo` (#2651).
- `tld.dart` removed (#2654); `StreamMessageComposer` relaxed its URL regex from `[a-z]{2,4}` to `[a-z]{2,}` and dropped the `isValidTLD` filter at both call sites.
- `StreamMessageListView` and `separated_reorderable_list_view`: v10's design-refresh version retained on top of the perf changes.

_Scroll, highlight, pagination plumbing_
- `_moveToAndHighlight` renamed to `_scrollToMessage` and tightened: `messageId` is required (was nullable), dropped the unused `messages` / `initialScrollIndex` / `scrollTo` parameters, added `alignment` (default 0.5) and `highlight` (default true) knobs, guards on `controller.isAttached` before scrolling. In thread mode, bails when the target reply isn't loaded (no thread-around-reply loader is exposed today — matches Android/RN; iOS/SwiftUI have `loadPageAroundReplyId` and remain the proper-fix template).
- Initial-highlight selection discriminates channel vs thread — threads read `_ThreadHighlightScope` only, the channel reads `streamChannel.initialMessageId` only. Previously the channel-scoped id leaked into thread mode and could mis-route `_scrollToMessage` into the channel pagination path. Matches Android's `parentMessageId`-discriminated focus and SwiftUI's per-VM highlight environment. `_ThreadHighlightScope.of` switched to `dependOnInheritedWidgetOfExactType` for O(1) lookup.
- Pagination `LoadingIndicator` reworked: `StatefulWidget` that caches the `queryTopMessages` / `queryBottomMessages` stream (matches `MessageListCore`'s pattern — the getters on `StreamChannelState` return a fresh BehaviorSubject view per access), reads `StreamChannel.of(context)` instead of taking it as a prop, drops the unused `streamTheme` / `isThreadConversation` fields. In-flight spinner uses `StreamLoadingSpinner` and the existing `widget.builders.paginationLoadingIndicator` override; the error tile is a tappable `StreamScrollViewLoadMoreError.list` that retries via `_paginateData`. Initial-load `defaultErrorBuilder` upgraded from plain `Text` to `StreamScrollViewErrorWidget` with retry → `streamChannel.reloadChannel()`. `_paginateData` drops its unused `channel` parameter.
- Inner `ScrollablePositionedList.separated` gets a stable `Key('mlv-${cid}-${parentId}')` — same-channel/same-thread rebuilds keep the SPL's scroll position and item-position cache; channel swap or thread enter/leave forces a clean remount.
- `MessageListView` default builders switched from `if (X != null) return X!(...)` to `if (X case final builder?) return builder(...)`, dropping the bang-assertions.

_Other_
- Committed two previously-untracked generated part files (`stream_chat_theme.g.theme.dart`, `thread_list_tile_theme.g.theme.dart`) — CI analyze couldn't compile without them.
- Suppressed a deprecation on `SizeTransition.axisAlignment` — its replacement `alignment` was added after Flutter 3.41, beyond the SDK floor (3.38.1).
- Added `auto_scroll_test.dart` and `stream_chat_theme_test.dart` (10 tests covering message-list scroll behavior and theme ancestor lookup).

**`sample_app`**
- Completed the Sentry → Firebase Crashlytics migration (#2665) on v10's redesigned sample app: removed `sentry_flutter` from `pubspec.yaml` and swapped the `Sentry.captureException` call in `auth_controller.dart` for `FirebaseCrashlytics.instance.recordError(error, stack, reason: record.message)`. `main.dart` was already on Crashlytics; the log handler in `auth_controller` was the last sentry consumer.

**CI / repo**
- Path/draft gating job (`gate`) added to `legacy_version_analyze`, `check_db_entities`, and `stream_flutter_workflow` (#2669).
- Flutter 3.44 fixes (#2667), pana / build cleanups (#2656), local-setup CI fixes (#2650).
- SPM disabled at the CI level instead of via pubspec flag (#2675, #2672).
- `melos.yaml`: kept v10's higher floors; added `firebase_crashlytics` (#2665); dropped `sentry_flutter`.
- `release-pr` skill added (#2657).
- Reformatted eight files to match Dart 3.12's `dart format` output. Dart 3.10 had a bug ([dart-lang/dart_style#1785](https://github.com/dart-lang/dart_style/issues/1785)) that left a stray `\n  ;` on enum declarations with parameters when `trailing_commas: preserve` is set; CI's Dart 3.12 (Flutter 3.44) correctly collapses them.

**Docs / tests**
- `docs_screenshots/voice_recording_test.dart`: replaced `StreamChatThemeData().copyWith(...)` with the factory constructor — the generator-emitted `copyWith` returns the base `ThemeExtension<StreamChatThemeData>` type, which isn't assignable to a `StreamChatThemeData?` parameter.

## Notes / follow-ups

- **`sample_app/`**: v10's redesigned sample app retained as a unit (gradle, manifest, main.dart, pbxproj — auto-merge would have left it inconsistent with v10's app.dart / pubspec). #2665's Sentry → Crashlytics work is now fully ported (see above).
- **`mark_read_test.dart`**: dropped a dead `when(() => channel.on(any(), any(), any(), any()))` stub that leaked `any()` matchers into the next `setUp` (since `MockChannel.on()` is a concrete override, not a mocktail stub).
- **Locale**: `formatRecentDateTime` hardcodes English `'at'` separator — needs a `dateAtTimeText({date, time})` Translations method for locale support. Tracked.
- **Deferred perf**: `StreamTimestamp` / `StreamDateDivider` still call Jiffy (~684ms / 11.8% in the latest profile) — should move to a cached `intl.DateFormat`.
- **Goldens**: 49 pixel-diff failures in reactions / polls / avatars / icons remain — pre-existing Flutter-SDK rendering drift, unrelated to this PR. Regenerate with `melos run update:goldens` if needed.
- **`Channel.query()` truncate-on-reload**: after a `loadChannelAtMessage(Y)` jump, calling `reloadChannel()` merges the latest 30 with the around-Y window instead of replacing — older messages (including the one that contained the quoted tap) end up unreachable without paginating back. Fix needs to land in v9 first (same LLC code), then port to v10. Tracked.
- **Thread-around-reply loader**: `_scrollToMessage` bails in thread mode when the target reply isn't loaded. Backend supports `id_around` on `GET /messages/{parent_id}/replies` and the LLC already exposes it via `Channel.getReplies(parentId, options: PaginationParams(idAround: ...))`; iOS UIKit (`MessageController.loadPageAroundReplyId`) and SwiftUI wire it through, Android and React Native do not. Wire a `loadPageAroundReplyId`-style helper into `StreamChannel` / the thread controller as a follow-up.

## Test plan

- [x] `dart test packages/stream_chat` — 1381/1381 pass
- [x] `flutter test packages/stream_chat_flutter_core` — 239/239 pass
- [x] `flutter test packages/stream_chat_flutter` — touched-area tests all pass (`date_formatter_test`, `message_list_view_test`, `mark_read_test`, `auto_scroll_test`, `timestamp_test`, `stream_chat_test`, `stream_channel_test`)
- [x] CI: `analyze`, `format`, `test`, `build (android)` green
- [ ] CI: `build (ios)` — Sentry/Crashlytics archive failure was the prior blocker; should clear now that `sentry_flutter` is gone (first run since the migration)
- [ ] Manual: `sample_app` boots against v10 on at least one platform


packages/stream_chat_flutter/lib/scrollable_positioned_list/src/positioned_list.dart
```diff
@@ -184,80 +228,189 @@ class _PositionedListState extends State<PositionedList> {
...
-                    ? _buildItem(index + widget.positionedIndex + 1)
+                    ? _buildItem(index + widget.positionedIndex)
                     : _buildSeparatedListElement(
-                        index + widget.positionedIndex * 2 + 1,
+                        index + widget.positionedIndex * 2,
                       ),
-                childCount: widget.separatorBuilder == null
-                    ? widget.itemCount - widget.positionedIndex - 1
-                    : 2 * (widget.itemCount - widget.positionedIndex - 1),
+                childCount: widget.itemCount != 0 ? 1 : 0,
```
renefloor, line 382: I wonder if we really need to wrap this item in a list. Is that to preserve the keyed tree?

packages/stream_chat_flutter/lib/src/message_input/stream_message_composer.dart
```diff
@@ -855,7 +854,8 @@ class DefaultStreamMessageComposerState extends State<DefaultStreamMessageCompos
             ),
             SizeTransition(
               sizeFactor: _pickerAnimation,
-              alignment: AlignmentDirectional.topStart,
+              // ignore: deprecated_member_use
+              axisAlignment: -1,
```
renefloor, line 858: Shouldn't we just use `alignment` in v10? axisAlignment was deprecated in flutter `v3.41`.

xsahil03x: Its only available in the latest flutter version

packages/stream_chat_flutter/lib/scrollable_positioned_list/src/positioned_list.dart
```diff
@@ -30,6 +32,7 @@ class PositionedList extends StatefulWidget {
     required this.itemCount,
     required this.itemBuilder,
     this.separatorBuilder,
+    this.itemKeyBuilder,
```
renefloor, line 35: Does it make sense to make this required?

renefloor (approved):

PR #2683, opened 2026-05-25
fix(ui): auto-scroll on optimistic local messages

xsahil03x:
Closes [FLU-499](https://linear.app/stream/issue/FLU-499/auto-scroll-on-optimistic-local-messages) (QA item #97 from [Flutter QA — New Design](https://www.notion.so/stream-wiki/Flutter-QA-New-Design-3296a5d7f9f680c4b592c4b677dd5a78)).

## Summary

- `StreamMessageListView`'s auto-scroll-to-bottom listener now subscribes to `channel.state.newMessageStream` (or `newThreadMessageStream(parentId)` in thread mode), exposed via a new `NewMessageStreamX` extension on `ChannelClientState` (in `mlv_utils.dart`, not exported). Diffs `messagesStream` / `threadsStream` for "a new message landed at the bottom" — covering both server-confirmed `message.new` events AND optimistic local sends. The previous `channel.on(EventType.messageNew)` path missed local sends because no WS event fires for them.
- Gated on `isUpToDate`: no auto-scroll while the channel is loaded around a historic message. The `false → true` flip is a wholesale replacement and is also suppressed.
- Detection uses `(newLast.id != lastSeen.id) && newLast.createdAt.isAfter(lastSeen.createdAt)` — handles tail deletion, optimistic→server confirm (same id, no double-scroll), and pruning + append in the same emission.

## Test plan

- [x] `flutter test test/src/message_list_view/auto_scroll_test.dart` — 7/7 pass (new file).
- [x] `flutter test test/src/message_list_view/` — full MLV suite passes.
- [x] `flutter analyze lib/src/message_list_view/` — clean.
- [x] Manual on Pixel 4 (profile mode): send while at bottom, send while scrolled up, receive while at bottom, receive while scrolled up.
- [x] Manual: send → server confirm doesn't double-scroll.
- [ ] Manual: send → fails → tap retry — no jump, message stays in place.
- [ ] Manual: open a thread, send in thread — thread auto-scrolls.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed auto-scroll behavior when sending outgoing messages before server confirmation is received, ensuring the message list stays positioned at the latest message.

[Review Change Stack](https://app.coderabbit.ai/change-stack/GetStream/stream-chat-flutter/pull/2683?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/stream_chat_flutter/CHANGELOG.md
+++ b/packages/stream_chat_flutter/CHANGELOG.md
@@ -1,3 +1,10 @@
+## Upcoming Changes
+
+🐞 Fixed
+
+- Fixed `StreamMessageListView` not auto-scrolling to the bottom on the user's own outgoing message
+  until the server confirmed it.
+
 ## 9.24.0
 
 ✅ Added
--- a/packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
+++ b/packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
@@ -416,7 +416,7 @@ class _StreamMessageListViewState extends State<StreamMessageListView> {
   MessageListController get _messageListController =>
       widget.messageListController ?? _defaultController;
 
-  StreamSubscription? _messageNewListener;
+  StreamSubscription<Message>? _messageNewListener;
   StreamSubscription? _userReadListener;
 
   @override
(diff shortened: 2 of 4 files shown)


xsahil03x:
Triage of CodeRabbit findings:

1. **`lengthGrew && lastChanged` drops events on append-with-trim** ([inline](https://github.com/GetStream/stream-chat-flutter/pull/2683#discussion_r3305022024)) — ✅ fixed. The gate is now `lastChanged && newLast.createdAt.isAfter(lastSeen.createdAt)`, so retention pruning + append in the same emission still fires (length stays flat but the tail id changes to a newer createdAt). CodeRabbit already auto-marked this as addressed.

2. **Rebind listener when `parentMessage` changes** (flagged across two reviews) — ✅ fixed. Extracted `_subscribeMessageNewListener()` and added a `didUpdateWidget` that re-runs it when `oldWidget.parentMessage?.id != widget.parentMessage?.id`. The previous code only rebound on `StreamChannel` change, so a thread parent switch on the same channel would leak the old subscription.

3. **`.reversed.toList()` inconsistent with `mark_read_test.dart`** — declining. The reverse is load-bearing here, not stylistic. `generateConversation(...)` returns newest-first, but production `state.messages` is sorted oldest-first (`channel.dart:3619` — `_sortByCreatedAt: a.createdAt.compareTo(b.createdAt)` is ascending). The auto-scroll listener uses `messages.last` as "newest"; without the reverse it would be "oldest" and the tests would fire on the wrong message. `mark_read_test.dart` doesn't reverse because its assertions don't depend on the bottom-most-message identity — it only checks that `channel.markRead()` was called. Different test contracts.

renefloor (approved):

PR #2686, opened 2026-05-26
fix(llc)!: prevent external mutation of ClientState collections

xsahil03x:
Closes [FLU-500](https://linear.app/stream/issue/FLU-500/make-clientstatechannels-and-related-collection-getters-unmodifiable)

## Summary

The collection getters on `ClientState` — `channels`, `users`, `activeLiveLocations` — returned live internal references. Callers could mutate the cache directly and silently bypass `channelsStream` / `usersStream` / `activeLiveLocationsStream`. Wrap them in unmodifiable maps/lists at write time so both the synchronous getter and the stream emit immutable values, and lock down the cache-mutation surface with `@internal`.

## What changed

**Read-side**
- `state.channels`, `state.users`, `state.activeLiveLocations` now return unmodifiable maps/lists. Calling `.clear()`, `.remove(...)`, `[k] = v`, etc. throws `UnsupportedError`.
- The wrap happens once at write time inside the setter / cache-mutation method, so the getter is zero-allocation. Subscribers to `channelsStream` / `usersStream` / `activeLiveLocationsStream` also receive unmodifiable values.
- Initial seeds are unmodifiable too — subscribers connecting *before* any write still see immutable values.

**Write-side hardening**
- All write calls in `ClientState` go through `safeAdd(...)` (the existing close-safe extension). Emissions after close are no-ops instead of throwing.
- `removeChannel` early-returns when the channel isn't cached and emits a freshly-built map, so `.distinct()` subscribers correctly observe the change. Previously `channels = channels..remove(cid)` re-emitted the same reference and missed the change for `.distinct()`.
- `dispose()` closes `_channelsController` before iterating the cached channels and disposing them. Each `Channel.dispose()` calls `removeChannel(cid)` which `safeAdd`-no-ops on the closed controller, so teardown produces **zero** subscriber emissions (vs. N progressively-shrinking maps previously).

**`@internal` audit**
The cache-mutation API surface on `ClientState` is now marked `@internal`. App flows (`connectUser`, `client.queryChannels`, `client.startLiveLocationSharing`, etc.) populate the cache through these methods and are unaffected.

- Setters: `channels=`, `currentUser=`, `activeLiveLocations=`, `totalUnreadCount=`, `blockedUserIds=`
- Methods: `updateUsers`, `updateUser`, `addChannels`, `removeChannel`

`ChannelClientState` is left out of this PR — the same pattern can be applied later when threads/typing events become a priority.

## Extracted wrapper module (post-review)

The unmodifiable-wrap discipline lives in two `@internal` wrapper classes (in `lib/src/core/util/immutable_collection_subjects.dart`):

- `ImmutableMapBehaviorSubject<K, V>`
- `ImmutableListBehaviorSubject<E>`

Both wrap a `BehaviorSubject<UnmodifiableMapView/ListView>`, extend `StreamView<T>`, and implement `Sink<T>` — so they behave like a regular broadcast stream + sink. Callers pass plain `Map<K, V>` / `List<E>` to `add` / `safeAdd`; the unmodifiable wrap happens inside, structurally impossible to forget or bypass via the wrapper's public surface.

The file also exports two public typedefs to keep the "Immutable" naming consistent:

- `ImmutableMap<K, V>` = `UnmodifiableMapView<K, V>`
- `ImmutableList<E>` = `UnmodifiableListView<E>`

The three collection controllers in `ClientState` now use the wrappers instead of the inline `BehaviorSubject<...>.seeded(.unmodifiable(...))` pattern.

## Tests

8 tests under `ClientState mutation guards`:

- `state.channels` returns an unmodifiable view
- `state.users` returns an unmodifiable view
- `state.activeLiveLocations` returns an unmodifiable view
- `removeChannel` emits a fresh map so `.distinct()` subscribers see the change
- initial seeded values are unmodifiable (before any write)
- `channelsStream` emits unmodifiable maps
- `usersStream` emits unmodifiable maps
- `activeLiveLocationsStream` emits unmodifiable lists

Plus 18 unit tests for the wrappers in `immutable_collection_subjects_test.dart` — covering both factory constructors, `add`/`safeAdd` wrap behavior, `add` throws after close, `safeAdd` no-op after close, subscribers receive unmodifiable values, and IS-A `Stream`/`Sink`.

Full LLC suite passes (1407/1407) on the rebased branch.

## Breaking

`fix(llc)!` — code that was mutating these collections directly (`state.channels.remove(...)`, `state.users.clear()`, etc.) will now throw `UnsupportedError` at runtime. Migration: go through the existing `addChannels` / `removeChannel` / `updateUser` / `updateUsers` methods (now `@internal`, but callable by SDK code; apps populate the cache via API methods that hit these internally).

### DIFF
--- a/packages/stream_chat/lib/src/client/client.dart
+++ b/packages/stream_chat/lib/src/client/client.dart
@@ -42,6 +42,7 @@ import 'package:stream_chat/src/core/models/reaction.dart';
 import 'package:stream_chat/src/core/models/thread.dart';
 import 'package:stream_chat/src/core/models/user.dart';
 import 'package:stream_chat/src/core/util/event_controller.dart';
+import 'package:stream_chat/src/core/util/extension.dart';
 import 'package:stream_chat/src/core/util/in_flight_cache.dart';
 import 'package:stream_chat/src/core/util/list_extensions.dart';
 import 'package:stream_chat/src/core/util/utils.dart';
@@ -2320,7 +2321,7 @@ class ClientState {
       _client.on(EventType.channelHidden).listen((event) async {
         final eventChannel = event.channel!;
         await _client.chatPersistenceClient?.deleteChannels([eventChannel.cid]);
-        channels.remove(eventChannel.cid)?.dispose();
+        channels[eventChannel.cid]?.dispose();
       }),
     );
   }
@@ -2376,7 +2377,7 @@ class ClientState {
             if (isCurrentUser && event.channel != null) {
               final eventChannel = event.channel!;
               await _client.chatPersistenceClient?.deleteChannels([eventChannel.cid]);
-              channels.remove(eventChannel.cid)?.dispose();
+              channels[eventChannel.cid]?.dispose();
             }
           }),
     );
@@ -2392,7 +2393,7 @@ class ClientState {
           .listen((Event event) async {
             final eventChannel = event.channel!;
             await _client.chatPersistenceClient?.deleteChannels([eventChannel.cid]);
-            channels.remove(eventChannel.cid)?.dispose();
+            channels[eventChannel.cid]?.dispose();
           }),
     );
   }
@@ -2516,22 +2517,25 @@ class ClientState {
 
   /// Sets the user currently interacting with the client
   /// note: this fully overrides the [currentUser]
+  @internal
   set currentUser(OwnUser? user) {
     _computeUnreadCounts(user);
-    _currentUserController.add(user);
+    _currentUserController.safeAdd(user);
   }
 
   /// Update all the [users] with the provided [userList]
+  @internal
   void updateUsers(List<User?> userList) {
     final newUsers = {
       ...users,
       for (final user in userList)
         if (user != null) user.id: user,
     };
-    _usersController.add(newUsers);
+    _usersController.safeAdd(.unmodifiable(newUsers));
   }
 
   /// Update the passed [user] in state
+  @internal
   void updateUser(User? user) => updateUsers([user]);
 
   /// The current user
@@ -2547,20 +2551,17 @@ class ClientState {
   Stream<Map<String, User>> get usersStream => _usersController.stream;
 
   /// The current active live locations shared by the user.
-  List<Location> get activeLiveLocations {
-    return _activeLiveLocationsController.value;
-  }
+  List<Location> get activeLiveLocations => _activeLiveLocationsController.value;
 
   /// The current active live locations shared by the user as a stream.
-  Stream<List<Location>> get activeLiveLocationsStream {
-    return _activeLiveLocationsController.stream;
-  }
+  Stream<List<Location>> get activeLiveLocationsStream => _activeLiveLocationsController.stream;
 
   /// Sets the active live locations.
+  @internal
   set activeLiveLocations(List<Location> locations) {
     // For safe-keeping, we filter out any inactive locations before update.
-    final activeLocations = [...locations.where((it) => it.isActive)];
-    _activeLiveLocationsController.add(activeLocations);
+    final activeLocations = locations.where((it) => it.isActive);
+    _activeLiveLocationsController.safeAdd(.unmodifiable(activeLocations));
   }
 
   /// The current unread channels count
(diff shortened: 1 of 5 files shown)


packages/stream_chat/lib/src/client/client.dart
```diff
@@ -2320,7 +2321,7 @@ class ClientState {
       _client.on(EventType.channelHidden).listen((event) async {
         final eventChannel = event.channel!;
         await _client.chatPersistenceClient?.deleteChannels([eventChannel.cid]);
-        channels.remove(eventChannel.cid)?.dispose();
+        channels[eventChannel.cid]?.dispose();
```
renefloor, line 2325: Are these still being removed from the list somehow, or is that not needed?

packages/stream_chat/lib/src/client/client.dart
```diff
@@ -2587,52 +2588,57 @@ class ClientState {
...
     }
 
     if (user?.unreadThreads case final count?) {
-      _unreadThreadsController.add(count);
+      _unreadThreadsController.safeAdd(count);
     }
   }
 
-  final _channelsController = BehaviorSubject<Map<String, Channel>>.seeded({});
+  final _channelsController = BehaviorSubject<Map<String, Channel>>.seeded(.unmodifiable(const {}));
```
renefloor: I wonder how we can make sure that we never add a modifiable list/map to the `BehaviorSubject`. Maybe we can extend `BehaviorSubject`

Something like this:
```dart
class UnmodifiableBehaviorSubject<T> extends BehaviorSubject<T> {

  factory UnmodifiableBehaviorSubject({
    void Function()? onListen,
    void Function()? onCancel,
    bool sync = false,
  }) {
    return UnmodifiableBehaviorSubject(
      onListen: onListen,
      onCancel: onCancel,
      sync: sync,
    );
  }

  factory UnmodifiableBehaviorSubject.seeded(
    T seedValue, {
    void Function()? onListen,
    void Function()? onCancel,
    bool sync = false,
  }) {
    _assertNotModifiable(seedValue);
    return UnmodifiableBehaviorSubject.seeded(
      seedValue,
      onListen: onListen,
      onCancel: onCancel,
      sync: sync,
    );
  }

  @override
  void add(T event) {
    _assertNotModifiable(event);
    super.add(event);
  }

  static void _assertNotModifiable(dynamic value) {
    if (value is List) {
      assert(value is UnmodifiableListView, 'List is modifiable');
    }
    if (value is Map) {
      assert(value is UnmodifiableMapView, 'Map is modifiable');
    }
    if (value is Set) {
      assert(value is UnmodifiableSetView, 'Set is modifiable');
    }
  }
}
```

renefloor: Hmm, the constructor is wrong here, solution would probably be more similar to this:
```dart
class UnmodifiableBehaviorSubject<T> {
  final BehaviorSubject<T> _subject;

  UnmodifiableBehaviorSubject._(this._subject);

  factory UnmodifiableBehaviorSubject({
    void Function()? onListen,
    void Function()? onCancel,
    bool sync = false,
  }) {
    return UnmodifiableBehaviorSubject._(
      BehaviorSubject<T>(
        onListen: onListen,
        onCancel: onCancel,
        sync: sync,
      ),
    );
  }

  factory UnmodifiableBehaviorSubject.seeded(
    T seedValue, {
    void Function()? onListen,
    void Function()? onCancel,
    bool sync = false,
  }) {
    _assertNotModifiable(seedValue);
    return UnmodifiableBehaviorSubject._(
      BehaviorSubject<T>.seeded(
        seedValue,
        onListen: onListen,
        onCancel: onCancel,
        sync: sync,
      ),
    );
  }

  /// The current value, or throws if no value has been emitted yet.
  T get value => _subject.value;

  /// The current value, or `null` if no value has been emitted yet.
  T? get valueOrNull => _subject.valueOrNull;

  /// Whether a value has been emitted.
  bool get hasValue => _subject.hasValue;

  /// The stream of values.
  Stream<T> get stream => _subject.stream;

  void safeAdd(T event) {
    _assertNotModifiable(event);
    _subject.safeAdd(event);
  }

  void close() => _subject.close();

  static void _assertNotModifiable(dynamic value) {
    if (value is List) {
      assert(value is UnmodifiableListView, 'List is modifiable');
    }
    if (value is Map) {
      assert(value is UnmodifiableMapView, 'Map is modifiable');
    }
    if (value is Set) {
      assert(value is UnmodifiableSetView, 'Set is modifiable');
    }
  }
}
```

renefloor: Btw, it's not really possible to check the type of `List.unmodifiable`, so we have to change the callers to using `UnmodifiableListView` instead of `.unmodifiable`.

See POC: https://github.com/GetStream/stream-chat-flutter/pull/2692

packages/stream_chat/lib/src/client/client.dart
xsahil03x: Yep — `Channel.dispose()` calls `client.state.removeChannel('$cid')` internally (`channel.dart:2344`), so disposing the channel removes it from the cache and emits on `channelsStream`. The old code's explicit `channels.remove(cid)` was effectively redundant with that — it just bypassed the stream emission via direct map mutation.

packages/stream_chat/lib/src/client/client.dart
xsahil03x: Landed in 8363ab1a9 as `ImmutableMapBehaviorSubject<K, V>` and `ImmutableListBehaviorSubject<E>` in `lib/src/core/util/immutable_collection_subjects.dart`.

Shape we ended up with (composition wrapper, not assert-based):

```dart
class ImmutableMapBehaviorSubject<K, V> extends StreamView<Map<K, V>>
    implements ValueStream<Map<K, V>>, Sink<Map<K, V>> {

  factory ImmutableMapBehaviorSubject({onListen, onCancel, sync = false}) => ...;
  factory ImmutableMapBehaviorSubject.seeded(Map<K, V> seed, {...}) => ...;

  final BehaviorSubject<UnmodifiableMapView<K, V>> _subject;

  // ValueStream + Sink getters/methods forward to _subject.
  // add() wraps with UnmodifiableMapView before forwarding.
  // safeAdd() calls _subject.safeAdd which uses the existing extension.
}
```

Two tradeoffs we resolved along the way:

1. **Runtime asserts vs. type-level guarantee** — instead of asserting that callers passed `UnmodifiableMapView`, the wrapper accepts plain `Map<K, V>` and does the wrap itself inside `add` / `safeAdd`. The wrap is structurally impossible to forget; no debug-only check needed. The underlying `BehaviorSubject` is typed `<UnmodifiableMapView<K, V>>`, so the compiler also enforces the wrap at every internal write site.

2. **Why composition, not `extension type` or extending `Subject`** — both were tempting for "looks like a regular Subject," but:
   - `extension type` redeclarations are static-only, so `wrapper.safeAdd(plainMap)` resolved through the existing `StreamControllerX.safeAdd` extension *bypasses* our `add` redeclaration and either throws `_TypeError` at runtime (when the rep type is `BehaviorSubject<UnmodifiableMapView>`) or silently emits an unwrapped map (when rep is `BehaviorSubject<Map>`).
   - Extending `Subject<T>` requires reimplementing `BehaviorSubject`'s `_Wrapper`, `Rx.defer` seeding, and value/error tracking (~80 LOC of stdlib coupling) because of `StreamController<T>` invariance — we can't pass `BehaviorSubject<UnmodifiableMapView<K, V>>` to `super(StreamController<Map<K, V>>, ...)`.
   - Composition is bypass-safe (private `_subject`), virtually dispatched, and works with the existing `safeAdd` extension internally without any of that overhead.

Also marked the wrappers `@internal` since they're SDK plumbing, not part of the public surface.

Tests: 16 unit tests for the wrappers plus the existing 8 `ClientState mutation guards`, 1405/1405 LLC tests pass.

packages/stream_chat/lib/src/core/util/immutable_collection_subjects.dart
```diff
@@ -0,0 +1,211 @@
...
+/// // Writes throw `UnsupportedError`.
+/// controller.value['general'] = otherChannel;
+/// ```
+///
+/// See also:
+///
+///  * [ImmutableListBehaviorSubject], the list counterpart.
+@internal
+class ImmutableMapBehaviorSubject<K, V> extends StreamView<Map<K, V>>
+    implements ValueStream<Map<K, V>>, Sink<Map<K, V>> {
```
renefloor: I think we should be careful with implementing an interface we don't manage. If rxdart adds a method to the interface at some point this immediately breaks the sdk for us.

I don't think we have any reason to implement the rxdart interface right? It's only for internal use?

xsahil03x: Good call — fixed in c552ecd2c.

Dropped `implements ValueStream<T>` from both wrappers, plus the 7 forwarding getters that existed only to satisfy that interface (`valueOrNull`, `hasValue`, `error`, `errorOrNull`, `hasError`, `stackTrace`, `lastEventOrNull`). None were used by `client.dart`.

Kept `extends StreamView<T>` and `implements Sink<T>` since both are `dart:async` — stable stdlib interfaces, very unlikely to grow new abstract members. Same principle applies in theory, but the risk profile is fundamentally different from a third-party `^0.28.0` dep.

renefloor (approved):

PR #2704, opened 2026-06-02
fix(llc, core, ui): isolate channel state from sub-route wraps and smooth thread-open scroll

xsahil03x:
## Summary

Small cluster of related fixes around channel-state isolation and thread-page UX.

- `Channel.getMessagesById` is now a pure fetch — no longer merges the response into `ChannelState.messages`. Resolving a thread parent for an in-channel reply previously injected the parent into the loaded channel window (typically at index 0, since parents predate the loaded window).
- `StreamChannel.getMessage` cache lookup now also scans `state.threads.values` and `state.pinnedMessages` before falling back to the network.
- Added `StreamChannel.value` constructor. Wraps a channel for context access without running channel-page positioning. Use for sub-route or overlay wraps (thread page, info screens, long-press modal, attachment viewer).
- `MessageListCore` no longer reloads the parent channel from its dispose path when the disposing instance is in thread mode.
- `ScrollablePositionedList._startScroll` waits one frame for layout when `itemPositions` is empty. Previously a `scrollTo` invoked right after mount took the dual-controller teleport path even for nearby targets, producing a multi-screen fake-scroll where a short animation was expected. Matches Compose's `LazyListState.animateScrollToItem` (which suspends until first composition).

## Test plan

- [ ] In a channel with a thread that has an in-channel reply, tap "View Thread" on the reply — thread opens at the target without an animated scroll-up, channel window unchanged on pop.
- [ ] Tap a quoted (old) message in the channel, then long-press any message — channel doesn't silently reload to latest behind the modal.
- [ ] Open a thread, pop back — channel's loaded window unchanged.
- [ ] `streamChannel.getMessage(threadReplyId)` for a reply already loaded in `state.threads` resolves from cache (no network call).

## Summary by CodeRabbit

* **New Features**
  * Added an option to wrap pre-initialized channels while skipping automatic positioning/loading.

* **Bug Fixes**
  * Fetching messages by ID no longer mutates the loaded message window.
  * Fixed premature long-distance scroll behavior immediately after widget mount.
  * Improved message lookup to check multiple in-memory caches before a network fetch.
  * Channel reload now replaces previously loaded messages instead of merging.
  * Prevented channel reloads during disposal for thread views.

### DIFF
--- a/packages/stream_chat_flutter_core/lib/src/message_list_core.dart
+++ b/packages/stream_chat_flutter_core/lib/src/message_list_core.dart
@@ -293,6 +293,10 @@ class MessageListCoreState extends State<MessageListCore> {
     // If the channel is up to date, we don't need to reload it.
     if (_upToDate) return;
 
+    // Thread conversations share the parent channel; reloading from a
+    // disposing thread would clobber the channel page state.
+    if (_isThreadConversation) return;
+
     try {
       return await _streamChannel?.reloadChannel();
     } catch (_) {
@@ -305,7 +309,7 @@ class MessageListCoreState extends State<MessageListCore> {
   @override
   void dispose() {
     _teardownController(widget.messageListController);
-    if (!_isThreadConversation) _reloadChannelIfNeeded();
+    _reloadChannelIfNeeded();
     super.dispose();
   }
 }
--- a/packages/stream_chat_flutter_core/lib/src/message_list_core.dart
+++ b/packages/stream_chat_flutter_core/lib/src/message_list_core.dart
@@ -305,7 +305,7 @@ class MessageListCoreState extends State<MessageListCore> {
   @override
   void dispose() {
     _teardownController(widget.messageListController);
-    _reloadChannelIfNeeded();
+    if (!_isThreadConversation) _reloadChannelIfNeeded();
     super.dispose();
   }
 }
--- a/packages/stream_chat_flutter_core/lib/src/stream_channel.dart
+++ b/packages/stream_chat_flutter_core/lib/src/stream_channel.dart
@@ -45,7 +45,30 @@ class StreamChannel extends StatefulWidget {
     this.initialMessageId,
     this.errorBuilder = _defaultErrorBuilder,
     this.loadingBuilder = _defaultLoadingBuilder,
-  });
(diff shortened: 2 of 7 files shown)


packages/stream_chat_flutter_core/lib/src/stream_channel.dart
```diff
@@ -45,7 +45,33 @@ class StreamChannel extends StatefulWidget {
...
+  ///    on mount.
+  const StreamChannel.value({
+    super.key,
+    required this.child,
+    required this.channel,
+  }) : showLoading = false,
+       initialMessageId = null,
+       errorBuilder = _defaultErrorBuilder,
+       loadingBuilder = _defaultLoadingBuilder,
+       _shouldPosition = false;
```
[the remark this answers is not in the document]
xsahil03x: The assert isn't needed — `_maybeInitChannel` calls `channel.watch()` if `state` is null before checking `_shouldPosition`, so descendants always see a watched channel regardless of which constructor was used. The asymmetry between the constructors is only about *positioning*, not about whether the channel is watched. Dropped the misleading "must already be initialized" line from the dartdoc to match.

packages/stream_chat_flutter_core/lib/src/message_list_core.dart
```diff
@@ -305,7 +305,7 @@ class MessageListCoreState extends State<MessageListCore> {
   @override
   void dispose() {
     _teardownController(widget.messageListController);
-    _reloadChannelIfNeeded();
+    if (!_isThreadConversation) _reloadChannelIfNeeded();
```
VelikovPetar: Perhaps it would make sense to move the `if (!_isThreadConversation)` check inside the `_reloadChannelIfNeeded()`?

VelikovPetar (approved):
Left a very small question, otherwise looks good!

packages/stream_chat_flutter_core/lib/src/message_list_core.dart
xsahil03x: Good call — done in 04e9c1018. Moved the `_isThreadConversation` guard inside `_reloadChannelIfNeeded()` so callers don't need to know the skip conditions.

PR #2709, opened 2026-06-02
feat(llc, core, persistence): Add support for `PredefinedFilters` on `QueryChannels`

VelikovPetar:
# Submit a pull request

Linear: [FLU-357](https://linear.app/stream/issue/FLU-357/predefined-filters)

## CLA

Adds support for server-defined **predefined filters** on `queryChannels` across the LLC, persistence, and core layers.

A predefined filter is a named filter/sort preset that lives on the server. The client sends `predefined_filter` (plus optional `filter_values` / `sort_values` to interpolate placeholders) instead of an inline filter, and the server echoes back the materialized `filter` and `sort` it applied.

**`stream_chat` (LLC)**
- `ChannelApi.queryChannels` and `StreamChatClient.queryChannels` accept `predefinedFilter`, `filterValues`, `sortValues`.
- New `StreamChatClient.queryChannelsWithResult` returns a `QueryChannelsResult` exposing both the live channels and the server-resolved `PredefinedFilter` (name + materialized filter + sort).
- New `PredefinedFilter` model and `predefined_filter` field on `QueryChannelsResponse`.
- `SortOption<T>.fromJson` added so sort specs can round-trip through persistence.
- New `ChatPersistenceClient` hooks: `getChannelStatesByPredefinedFilter` and `updateChannelQueriesByPredefinedFilter` (default no-op so existing implementations keep compiling).

**`stream_chat_persistence`**
- New `channel_query_metadata` table that stores the server-resolved `filter` + `sort` keyed by the predefined-filter query hash, so offline reads reconstruct the exact resolved spec the server applied.
- New `FilterConverter` and `ChannelStateSortOrderConverter` Drift converters.
- `StreamChatPersistenceClient` implements the new hooks; `ChannelQueryDao` gains `updateChannelQueriesByPredefinedFilter` and `getChannelsAndSpecByPredefinedFilter`.

- `StreamChannelListController` accepts `predefinedFilter`, `filterValues`, `sortValues`. When set, they take precedence over `filter` / `channelStateSort`.
- The controller tracks a `_resolvedChannelStateSort` that is overwritten with the server-resolved sort on every successful query, so event-driven inserts keep matching the server's ordering even when the caller only specifies a `predefinedFilter`.

**Sample app**
- `ChannelList` now uses the `stream_chat_flutter_sample_app` predefined filter (interpolated with the current `user_id`) to demonstrate the flow end-to-end.

### How to test
- Run the sample app and confirm the channel list renders, paginates, and reacts to events as before.
- Toggle offline (airplane mode) after an online query and reopen — channels for the predefined-filter query should be served from persistence, ordered using the server-resolved sort.

### Test coverage
- `predefined_filter_test.dart` — model.
- `predefined_filter_defaults_test.dart` — fallback-sort selection.
- `client_test.dart` — online/offline `queryChannels` paths and `queryChannelsWithResult`.
- `channel_api_test.dart` — request payload assertions.
- `filter_converter_test.dart` / `sort_order_converter_test.dart` — Drift converters.
- `channel_query_dao_test.dart` — DAO read/write of cids + metadata.
- `stream_chat_persistence_client_test.dart` — persistence client integration.
- `stream_channel_list_controller_test.dart` — controller path including resolved-sort tracking.

## Summary by CodeRabbit

* **New Features**
  * Added server-side predefined channel query filters via `predefinedFilter`, with `filterValues`/`sortValues`, including server-resolved filter/sort materialization.
  * Introduced `queryChannelsWithResult` returning `QueryChannelsResult` (channels plus the resolved predefined filter details).
  * Updated `StreamChannelListController` to accept predefined filters and apply the resolved sort automatically.
* **Deprecations**
  * Deprecated channel-query persistence read/write methods in favor of unified query metadata save/read APIs.
* **Other**
  * Improved online/offline channel querying with better persistence and caching of resolved query parameters.


packages/stream_chat/lib/src/client/client.dart
```diff
@@ -643,12 +645,89 @@ class StreamChatClient {
...
+  }
+
+  /// Requests channels with a given query, yielding a [QueryChannelsResult]
+  /// that carries both the live channel list and the server-resolved
+  /// [PredefinedFilter] spec (when one is associated with the query).
+  ///
+  /// Yields the offline-cached result first (when available), followed by
+  /// the online result. Concurrent identical online queries are coalesced
+  /// via [_queryChannelsCache].
+  Stream<QueryChannelsResult> queryChannelsWithResult({
```
VelikovPetar, line 718: I wasn't sure about the naming here, I tried to make it as explicit as possible. (it wasn't really possible to use the existing `queryChannels` without a breaking change, as we must change the return type).

packages/stream_chat/lib/src/client/query_channels_result.dart
```diff
@@ -0,0 +1,22 @@
+import 'package:stream_chat/src/client/channel.dart';
+import 'package:stream_chat/src/core/models/predefined_filter.dart';
+
+/// The result of a `queryChannelsWithResult` call on [StreamChatClient].
+///
+/// Carries the live [Channel] instances matching the query alongside the
+/// server-resolved [PredefinedFilter] spec (when one is associated with the
+/// query).
+class QueryChannelsResult {
```
VelikovPetar, line 9: I introduced this new model to serve as a container for the result of the `StreamChatClient.queryChannelsWithResult`.

packages/stream_chat/lib/src/client/client.dart
```diff
@@ -718,14 +806,48 @@ class StreamChatClient {
...
-      if (offlineChannels.isEmpty) rethrow;
+      if (offlineResult == null || offlineResult.channels.isEmpty) rethrow;
     }
   }
 
   /// Requests channels with a given query from the API.
   Future<List<Channel>> queryChannelsOnline({
     Filter? filter,
     SortOrder<ChannelState>? sort,
+    String? predefinedFilter,
```
VelikovPetar, line 810: Note: At the moment I didn't expose public `queryChannelsWithResultOnline`/`queryChannelsWithResultOffline` -> only the new `queryChannelsWithResults` is public. I didn't want to pollute the public API, with methods which most likely will not be used. Let me know if you think we should expose them as well.

packages/stream_chat/lib/src/core/api/channel_api.dart
```diff
@@ -71,6 +79,9 @@ class ChannelApi {
           // passed options
           if (sort != null) 'sort': sort,
           if (filter != null) 'filter_conditions': filter,
+          if (predefinedFilter != null) 'predefined_filter': predefinedFilter,
```
VelikovPetar, line 82: I didn't add any guards preventing the users to send both `predefined_filter` and `filter/sort`. The BE will prioritise the predefined filter if available. 
Another option would be to:
1. Fail early (throw error when building the `StreamChannelListController`, or when calling `client.queryChannels`)
2. Override `filter/sort` with `null` when `predefinedFilter` exists

But overall, I a think that letting the server handle this scenario would be the best, I don't want to introduce front-end guards which might become invalid if the server changes the default behaviour. 

Let me know if you think we should introduce some front-end checks anyway.

packages/stream_chat/lib/src/core/util/predefined_filter_defaults.dart
```diff
@@ -0,0 +1,51 @@
...
+/// Defaults used in the predefined-filter query flow.
+extension PredefinedFilterDefaults on Filter {
+  /// Fallback sort applied when the server doesn't echo back a sort spec
+  /// for a predefined-filter query, so the persisted spec is never null.
+  ///
+  /// Picks the sort field that matches the filter's time predicate (if any),
+  /// falling back to `last_updated` otherwise — this mirrors the server's
+  /// default sort for channel queries.
+  SortOrder<ChannelState> get predefinedFilterFallbackSort {
+    if (_touchesField('last_message_at')) {
```
VelikovPetar: Note: This follows the BE default sort, if no sort is defined in the predefined filter. See [this](https://getstream.slack.com/archives/CE5N802GP/p1778226798199859) discussion.

packages/stream_chat/lib/src/db/chat_persistence_client.dart
```diff
@@ -146,6 +147,47 @@ abstract class ChatPersistenceClient {
...
+  /// per channel and [paginationParams] to paginate results.
+  ///
+  /// The returned [QueryChannelsResponse] carries the persisted resolved
+  /// `predefinedFilter` spec (filter + sort) so offline callers can apply the
+  /// same filter/order the server applied on the last online query.
+  ///
+  /// Default implementation returns an empty response; persistence
+  /// implementations that support predefined-filter caching should override
+  /// this.
+  Future<QueryChannelsResponse> getChannelStatesByPredefinedFilter({
```
VelikovPetar: Do you think it is OK to return `QueryChannelsResponse` here - or should we introduce a new specific model for this?

packages/stream_chat/lib/src/core/api/channel_api.dart
xsahil03x: Yeah, this should be fine.

packages/stream_chat_flutter_core/lib/src/stream_channel_list_controller.dart
```diff
@@ -131,14 +170,19 @@ class StreamChannelListController extends PagedValueNotifier<int, Channel> {
       _kDefaultBackendPaginationLimit,
     );
     try {
-      await for (final channels in client.queryChannels(
+      await for (final result in client.queryChannelsWithResult(
         filter: filter,
         channelStateSort: channelStateSort,
```
xsahil03x: We should pass the `_resolvedChannelStateSort` here right?

packages/stream_chat_flutter_core/lib/src/stream_channel_list_controller.dart
```diff
@@ -160,14 +204,18 @@ class StreamChannelListController extends PagedValueNotifier<int, Channel> {
     final previousValue = value.asSuccess;
 
     try {
-      await for (final channels in client.queryChannels(
+      await for (final result in client.queryChannelsWithResult(
         filter: filter,
         channelStateSort: channelStateSort,
```
xsahil03x: same here: https://github.com/GetStream/stream-chat-flutter/pull/2709/changes#r3404015685

packages/stream_chat_flutter_core/lib/src/stream_channel_list_controller.dart
```diff
@@ -184,6 +232,11 @@ class StreamChannelListController extends PagedValueNotifier<int, Channel> {
     }
   }
 
+  void _resolveSort(QueryChannelsResult result) {
+    final resolved = result.predefinedFilter?.sort;
+    if (resolved != null) _resolvedChannelStateSort = resolved;
```
xsahil03x: Should we handle the null case here or always rely on the value returned from backend?

packages/stream_chat_flutter_core/lib/src/stream_channel_list_controller.dart
VelikovPetar: It should not really matter, because the `_resolvedChannelStateSort` is only 'relevant' for locally sorting the channels retrieved via the predefined filter way. Passing it here should not make a difference, because in the 'standard' path, `_resolvedChannelStateSort  == channelStateSort`, and in the 'predefined' path, the passed `channelStateSort` is ignored -> only the sort parsed from the `response.predefinedFilter` is taken into consideration. 
But then again, you are right, it is probably more correct to pass the `_resolvedChannelStateSort` here -> it makes the whole thing more future-proof, in case we introduce new ways of dynamic sort resolution.

packages/stream_chat_flutter_core/lib/src/stream_channel_list_controller.dart
VelikovPetar: Good point, I only added the fallback sort when `predefinedFilter.sort == null` when persisting the sort in the DB. I will fix this, we should also apply the same fallback here (only when `predefinedFilter != null && predefinedFilter.sort == null`). 

Edit: A bit more info, if we get a `PredefinedFilter` with `null` sort, it means that there is no sort template provided for that predefined filter. So the BE will use a fallback sorter, calculated as `predefinedFilterFallbackSort`

VelikovPetar: Addressed here: https://github.com/GetStream/stream-chat-flutter/pull/2709/commits/784e39cff2cd5fe71e0b2470fddec550a33e6991

Note: I also refactored the fallbackSort calculation a bit, to avoid exposing the algorithm as a public API.

packages/stream_chat_flutter_core/lib/src/stream_channel_list_controller.dart
VelikovPetar: Addressed here: https://github.com/GetStream/stream-chat-flutter/pull/2709/commits/784e39cff2cd5fe71e0b2470fddec550a33e6991

packages/stream_chat_flutter_core/lib/src/stream_channel_list_controller.dart
VelikovPetar: Addressed here: https://github.com/GetStream/stream-chat-flutter/pull/2709/commits/784e39cff2cd5fe71e0b2470fddec550a33e6991

packages/stream_chat_persistence/lib/src/stream_chat_persistence_client.dart
```diff
@@ -299,6 +299,30 @@ class StreamChatPersistenceClient extends ChatPersistenceClient {
...
+    return _queryChannelStatesImpl(
+      filter: filter,
+      sort: sort,
+      predefinedFilter: predefinedFilter,
+      filterValues: filterValues,
+      sortValues: sortValues,
+      messageLimit: messageLimit,
+      paginationParams: paginationParams,
+    );
+  }
```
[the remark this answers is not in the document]
VelikovPetar: Added simple docs on the method: https://github.com/GetStream/stream-chat-flutter/pull/2709/commits/e2ecdcbd4ccf761dc5639483d8c14646de998a51 
Kept it simple because the thorough docs are already written on the base ChatPersistenceClient.

packages/stream_chat_persistence/lib/src/stream_chat_persistence_client.dart
```diff
@@ -308,40 +332,127 @@ class StreamChatPersistenceClient extends ChatPersistenceClient {
...
+      filter: filter,
+      sort: sort,
+      predefinedFilter: predefinedFilter,
+      resolvedFilter: resolvedFilter,
+      resolvedSort: resolvedSort,
+      filterValues: filterValues,
+      sortValues: sortValues,
+      clearQueryCache: clearQueryCache,
+    );
+  }
```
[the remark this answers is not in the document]
VelikovPetar: Added simple docs on the method: https://github.com/GetStream/stream-chat-flutter/pull/2709/commits/e2ecdcbd4ccf761dc5639483d8c14646de998a51 
Kept it simple because the thorough docs are already written on the base ChatPersistenceClient.

packages/stream_chat_flutter_core/lib/src/stream_channel_list_controller.dart
```diff
@@ -184,6 +232,14 @@ class StreamChannelListController extends PagedValueNotifier<int, Channel> {
...
   }
 
+  void _resolveSort(QueryChannelsResult result) {
+    final predefinedFilter = result.predefinedFilter;
+    // Update the active sort only when predefinedFilter is present,
+    // otherwise use the initially set sort.
+    if (predefinedFilter == null) return;
+    _resolvedChannelStateSort = predefinedFilter.effectiveSort;
+  }
+
```
xsahil03x, line 243: We should also update the refresh method here.

```dart
  @override
  Future<void> refresh({bool resetValue = true}) {
    if (resetValue) _resolvedChannelStateSort = channelStateSort;
    return super.refresh(resetValue: resetValue);
  }
```

packages/stream_chat/lib/src/client/client.dart
```diff
@@ -656,12 +657,89 @@ class StreamChatClient {
...
+      watch: watch,
+      presence: presence,
+      memberLimit: memberLimit,
+      messageLimit: messageLimit,
+      paginationParams: paginationParams,
+      waitForConnect: waitForConnect,
+    )) {
+      yield result.channels;
+    }
+  }
```
xsahil03x: We can simplify it a bit using `.map` directly.

```suggestion
  Stream<List<Channel>> queryChannels({
    Filter? filter,
    SortOrder<ChannelState>? channelStateSort,
    String? predefinedFilter,
    Map<String, Object?>? filterValues,
    Map<String, Object?>? sortValues,
    bool state = true,
    bool watch = true,
    bool presence = false,
    int? memberLimit,
    int? messageLimit,
    PaginationParams paginationParams = const PaginationParams(),
    bool waitForConnect = true,
  }) => queryChannelsWithResult(
    filter: filter,
    channelStateSort: channelStateSort,
    predefinedFilter: predefinedFilter,
    filterValues: filterValues,
    sortValues: sortValues,
    state: state,
    watch: watch,
    presence: presence,
    memberLimit: memberLimit,
    messageLimit: messageLimit,
    paginationParams: paginationParams,
    waitForConnect: waitForConnect,
  ).map((result) => result.channels);
```

packages/stream_chat/lib/src/core/models/predefined_filter.dart
```diff
@@ -0,0 +1,84 @@
...
+  /// Sort to apply locally, matching what the server applies for this
+  /// predefined filter — the echoed [sort], or a default derived from
+  /// [filter] when [sort] is null.
+  SortOrder<ChannelState> get effectiveSort => sort ?? _defaultSortFor(filter);
+
+  static Filter _filterFromJson(Map<String, dynamic> json) => Filter.raw(value: json);
+}
+
+SortOrder<ChannelState> _defaultSortFor(Filter filter) {
+  if (_touchesField(filter, 'last_message_at')) {
```
xsahil03x:
```suggestion
  if (_touchesField(filter, ChannelSortKey.lastMessageAt)) {
```

packages/stream_chat_flutter_core/lib/src/stream_channel_list_controller.dart
xsahil03x: >Edit: A bit more info, if we get a PredefinedFilter with null sort, it means that there is no sort template provided for that predefined filter. So the BE will use a fallback sorter, calculated as predefinedFilterFallbackSort

Not sure, but in case there's no sort template provided for the predefined filter, then the backend will use a fallback sorter, so shouldn't we also set it if we don't receive it in the result? Currently we are handling the case only when its not null.

Related: https://github.com/GetStream/stream-chat-flutter/pull/2709/changes#r3414280459

packages/stream_chat/lib/src/client/client.dart
```diff
@@ -799,40 +926,95 @@ class StreamChatClient {
...
+    final cachedCids = channels.map((c) => c.channel!.cid).toList();
+    // Clear the query cache if we are refreshing.
+    final clearQueryCache = (paginationParams.offset ?? 0) == 0;
+
+    Filter? resolvedFilter;
+    SortOrder<ChannelState>? resolvedSort;
+    if (predefinedFilter != null) {
+      resolvedFilter = res.predefinedFilter?.filter ?? const Filter.empty();
+      resolvedSort =
+          res.predefinedFilter?.effectiveSort ?? const [SortOption<ChannelState>.desc(ChannelSortKey.lastUpdated)];
```
xsahil03x: Should we use the backend default sort here instead?

packages/stream_chat_flutter_core/lib/src/stream_channel_list_controller.dart
xsahil03x, line 13: I think we also need to update this and have the default one based on if we are passing predefined filters or not. wdyt?

VelikovPetar: Let me check how we can improve this. In the predefined filters case, we need to already have the `QueryChannels` response, so that we can derive the default `sort` based on the filter fields (if `sort` is not already returned by the response)

VelikovPetar: What do you think about something like:
```dart
 StreamChannelListController({
    required this.client,
    StreamChannelListEventHandler? eventHandler,
    this.filter,                                                                                                                                                                   
    this.predefinedFilter,                         
    this.filterValues,                                                                                                                                                             
    this.sortValues,                                                                                                                                                             
    this.presence = true,
    this.limit = defaultChannelPagedLimit,                                                                                                                                         
    this.messageLimit,                    
    this.memberLimit,                                                                                                                                                              
  }) : _eventHandler = eventHandler ?? StreamChannelListEventHandler(),                                                                                                          
       _resolvedChannelStateSort =                                     
           predefinedFilter == null ? channelStateSort : null,                                                                                                                     
       super(const PagedValue.loading());                     
```
I was also considering something like:
```dart
StreamChannelListController({                                                                                                                                                    
    required this.client,
    StreamChannelListEventHandler? eventHandler,                                                                                                                                   
    this.filter,                                                                                                                                                                 
    SortOrder<ChannelState>? channelStateSort,   // ← no inline default, nullable named param
    this.predefinedFilter,
    this.filterValues,
    this.sortValues,
    this.presence = true,
    this.limit = defaultChannelPagedLimit,
    this.messageLimit,
    this.memberLimit,
  }) : channelStateSort = channelStateSort ??
           (predefinedFilter == null ? defaultChannelListSort : null),
       _resolvedChannelStateSort = channelStateSort ??
           (predefinedFilter == null ? defaultChannelListSort : null),
       _eventHandler = eventHandler ?? StreamChannelListEventHandler(),
       super(const PagedValue.loading());
```
However this could be a breaking change, if a customer is instantiating the `StreamChannelListController ` as:
```dart
StreamChannelListController(
    client: client,
    channelStateSort: null,   // explicit "I don't want a sort"
  )
```
In which case they will now fallback to the `defaultChannelListSort`, instead of the explicit `null`.

xsahil03x (approved):

packages/stream_chat_flutter_core/lib/src/stream_channel_list_controller.dart
xsahil03x: I checked backend and they skip the `channelStateSort` if `predefined-filters` is specified. So i think we are good here.

PR #2720, opened 2026-06-03
ci(repo): skip sample app build when targeting master

xsahil03x:
## Summary

PR builds already cover pre-merge validation, and `distribute_internal` builds (and ships) both platforms on every master push — so running the sample app `build` matrix again on master push is duplicate work.

The build job now runs only on `pull_request` events. Since the workflow only triggers `push` on master, this leaves PR builds untouched and skips the redundant master-push build.

```yaml
if: needs.gate.outputs.should_run == 'true' && github.event_name == 'pull_request'
```

| Trigger | Outcome |
|---|---|
| Push to master | skip |
| PR → master | run |
| PR → v10.0.0 | run |

## Test plan

- [ ] After merge, the next master push leaves `build (android)` and `build (ios)` skipped.
- [ ] PRs (any target) still trigger both build jobs.

### DIFF
--- a/.github/workflows/stream_flutter_workflow.yml
+++ b/.github/workflows/stream_flutter_workflow.yml
@@ -167,7 +167,7 @@ jobs:
     name: build (${{ matrix.platform }})
     needs: gate
     runs-on: ${{ matrix.os }}
-    if: needs.gate.outputs.should_run == 'true'
+    if: needs.gate.outputs.should_run == 'true' && github.event_name == 'pull_request'
     timeout-minutes: 30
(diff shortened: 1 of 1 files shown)


renefloor:
Do we really need to have this workflow at all on push to master? If anything is really broken the builds would fail immediately. In my experience we usually find out about master being broken when a workflow doesn't go green on a PR. Maybe if we want to have this run on master we should push a notification to slack when the analyze/test fails on master.

renefloor (approved):

PR #2743, opened 2026-06-10
fix(ui): wire onUserAvatarTap on StreamMessageItem

### LINKED ISSUE #2741: onUserAvatarTap not wired up

stream_chat_flutter

iOS

v10.0.1

### What happened?

When using the new v10. Setting `onUserAvatarTap` on StreamMessageItem does nothing. From looking at the code, there is no onTap handler wired to the `leadingWidget` which hosts the avatar.

```dart
componentBuilders: sc.StreamComponentBuilders(
  extensions: streamChatComponentBuilders(
    messageItem: (context, props) {
      return DefaultStreamMessageItem(
        props: props.copyWith(
          onUserAvatarTap: (user) {
            // This never fires
            showUserInfoDialog(context, user);
          },
        ),
      );
    },
  }
```

### Steps to reproduce

```bash
Tap avatar next to message
```

_No response_

### Relevant log output

```shell

```

### Flutter analyze output

```shell

```

### Flutter doctor output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

xsahil03x:
## Summary

- Fixes #2741: `onUserAvatarTap` set on `StreamMessageItem` was a no-op because `DefaultStreamMessageItem` never wired a tap handler onto the leading avatar.
- Added `onTap: VoidCallback?` to `StreamMessageLeadingProps`. `DefaultStreamMessageLeading` now wraps the avatar in a `GestureDetector(behavior: opaque)` when `onTap` is set.
- `DefaultStreamMessageItem` now passes a closure into the leading that invokes `onUserAvatarTap` with `message.user` when both are non-null. Visibility wrapping is unchanged, so `gone` / `hidden` semantics still apply.
- `onTap` is intentionally a `VoidCallback?` on the slot rather than `void Function(User)?` — the leading is a generic slot and custom builders may render non-avatar content (timestamps, status badges, etc.). Avatar-specific semantics belong on `DefaultStreamMessageItem`, which owns `onUserAvatarTap`.

## Test plan

- [x] `dart analyze --fatal-infos` on `stream_chat_flutter` — no issues
- [x] `melos run format` — no files changed
- [x] New widget test in `test/src/message_widget/stream_message_leading_test.dart` covering tap-fires-callback and null-callback-no-op cases — both pass
- [ ] Manual repro from the issue: register a `messageItem` builder with `onUserAvatarTap`, tap the avatar, confirm the callback fires

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `onUserAvatarTap` callback to properly trigger when users tap message author avatars (issue `#2741`)

### DIFF
--- a/packages/stream_chat_flutter/lib/src/message_widget/stream_message_item.dart
+++ b/packages/stream_chat_flutter/lib/src/message_widget/stream_message_item.dart
@@ -429,7 +429,13 @@ class DefaultStreamMessageItem extends StatelessWidget {
     final effectiveRepliesVisibility = resolve((theme) => theme?.repliesVisibility);
 
     final leadingWidget = effectiveAvatarVisibility.apply(
-      StreamMessageLeading(message: message),
+      StreamMessageLeading(
+        message: message,
+        onTap: switch ((props.onUserAvatarTap, message.user)) {
+          (final onTap?, final user?) => () => onTap(user),
+          _ => null,
+        },
+      ),
     );
 
     final headerWidget = effectiveAnnotationVisibility.apply(
--- a/packages/stream_chat_flutter/CHANGELOG.md
+++ b/packages/stream_chat_flutter/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Unreleased
+
+🐞 Fixed
+
+- `StreamMessageItem.onUserAvatarTap` now fires when the author avatar is tapped. ([#2741](https://github.com/GetStream/stream-chat-flutter/issues/2741))
+
 ## 10.0.1
 
 🐞 Fixed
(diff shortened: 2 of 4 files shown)


VelikovPetar (approved):

packages/stream_chat_flutter/lib/src/message_widget/stream_message_item.dart
```diff
@@ -429,7 +429,13 @@ class DefaultStreamMessageItem extends StatelessWidget {
     final effectiveRepliesVisibility = resolve((theme) => theme?.repliesVisibility);
 
     final leadingWidget = effectiveAvatarVisibility.apply(
```
szechyjs, line 431: I would have wrapped the entire `leadingWidget` in the `GestureDetector` as now if `messageLeading` builder is used, it too needs to implement the onTap functionality.

PR #2747, opened 2026-06-12
feat(llc, core, persistence, ui): Add support for enhanced mentions

VelikovPetar:
# Submit a pull request

Linear: [FLU-343](https://linear.app/stream/issue/FLU-343/enhanced-mentions)

Related `stream-core-flutter` PR: https://github.com/GetStream/stream-core-flutter/pull/120

## CLA

Adds end-to-end support for **enhanced mentions** across the SDK — `@channel`, `@here`, role mentions, and user-group mentions — along with the backend APIs (roles, user groups, granular push preferences) that back them. Mentions are parsed on incoming messages, persisted offline, rendered as styled tappable spans in message text, and selectable from the composer's `@` autocomplete.

### `stream_chat`

**Added**

- `StreamChatClient.searchRoles()`.
- User-group APIs on `StreamChatClient`: `listUserGroups`, `searchUserGroups`, `getUserGroup`, `createUserGroup`, `updateUserGroup`, `deleteUserGroup`, `addUserGroupMembers`, `removeUserGroupMembers`.
- Enhanced-mention fields on `Message`: `mentionedChannel`, `mentionedHere`, `mentionedRoles`, `mentionedGroupIds`, `mentionedGroups`.
- `ChannelCapability.notifyChannel`, `notifyHere`, `notifyRole`, `notifyGroup`, with matching `Channel.canNotify*` getters.
- `ChannelConfig.pushLevel`, `pushNotifications`, and `chatPreferences` for per-channel-type push configuration.
- `PushLevel`, `ChatPreferenceLevel`, and `ChatPreferences` for granular per-category push preferences.
- `chatPreferences` field on `PushPreferenceInput`, `PushPreference`, and `ChannelPushPreference`.

### `stream_chat_flutter_core`

**Added**

- `mentionedChannel`, `mentionedHere`, `mentionedRoles`, `addMentionedRole`, `mentionedUserGroups`, and `addMentionedUserGroup` on `StreamMessageComposerController`.

### `stream_chat_flutter`

**Added**

- Per-type tile builders on `StreamMessageComposer` and `StreamMentionAutocompleteOptions` (`mentionChannelTileBuilder`, `mentionHereTileBuilder`, `mentionRoleTileBuilder`, `mentionUserGroupTileBuilder`, `mentionUserTileBuilder`), plus matching `onMention*Tap` callbacks on `StreamMentionAutocompleteOptions`.
- `StreamMessageListView.onMentionTap` and `StreamMessageItem.onMentionTap` — receives a typed `Mention` (`UserMention`, `ChannelMention`, `HereMention`, `RoleMention`, or `GroupMention`).

**Deprecated**

- `userMentionsTileBuilder` on `StreamMessageComposer` and `mentionsTileBuilder` on `StreamMentionAutocompleteOptions` in favor of `mentionUserTileBuilder`.
- `StreamMessageListView.onUserMentionTap` and `StreamMessageItem.onUserMentionTap` in favor of `onMentionTap`.
- `UserListX.search`.

### `stream_chat_persistence`

**Added**

- Persistence for the enhanced-mention fields on messages.

### `stream_chat_localizations`

**Added**

- `notifyChannelText`, `notifyHereText`, and `notifyRoleText(String role)` translations across all supported locales.

## Screenshots / Videos

https://github.com/user-attachments/assets/2024cb89-1114-45ba-9464-5377820eb46a

## Release Notes

* **New Features**
  * Added user groups API with search, CRUD, and member management capabilities
  * Added role search functionality
  * Introduced `@channel`, `@here`, role, and user-group mentions in messages with autocomplete support
  * Added per-channel push notification configuration and preferences
  * Added message mention tap callbacks with typed mention support
  * New mention tile customization and localization strings

* **Bug Fixes & Improvements**
  * Enhanced mention autocomplete with improved filtering and suggestion ordering

* **Documentation**
  * Updated changelogs and added localization strings across all supported languages


docs/docs_screenshots/pubspec.yaml
```diff
@@ -21,8 +21,8 @@ dependencies:
   stream_chat_flutter: ^10.0.1
   stream_core_flutter:
     git:
-      url: https://github.com/GetStream/stream-core-flutter
-      ref: 43f90a46d8092ca2c1e24c36af06f05fc0c633a6
+      url: https://github.com/GetStream/stream-core-flutter.git
```
VelikovPetar, line 24: Will need to update this to a specific version before merging this PR.

packages/stream_chat_flutter_core/lib/src/stream_message_composer_controller.dart
```diff
@@ -349,6 +349,54 @@ class StreamMessageComposerController extends ValueNotifier<Message> {
...
+  ///
+  /// No-op if a group with the same id is already mentioned.
+  void addMentionedUserGroup(UserGroup group) {
+    final groups = mentionedUserGroups;
+    if (groups.any((it) => it.id == group.id)) return;
+    final updated = [...groups, group];
+    message = message.copyWith(
+      mentionedGroups: updated,
+      mentionedGroupIds: updated.map((it) => it.id).toList(),
+    );
```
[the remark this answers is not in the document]
VelikovPetar: While in practice not possible, I will address this to be safe.

packages/stream_chat_flutter/lib/src/autocomplete/stream_mention_autocomplete_options.dart
```diff
@@ -149,27 +230,80 @@ class _StreamMentionAutocompleteOptionsState extends State<StreamMentionAutocomp
...
+      final allMembersCached = memberCount != null && memberCount == channelState.members.length;
 
-    final result = await _queryMembers(query);
-    return result.map((it) => it.user).whereType<User>().toList(growable: false);
+      // When every channel member is already cached locally we can match
+      // against the in-memory member + watcher set instead of hitting
+      // queryMembers.
+      if (allMembersCached) {
+        return searchUsersForMention(membersAndWatchers, query);
+      }
```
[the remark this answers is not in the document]
VelikovPetar: We didn't limit the local results even before the change, I would not enforce that now

packages/stream_chat_persistence/test/src/mapper/channel_mapper_test.dart
```diff
@@ -122,6 +146,15 @@ void main() {
...
     );
+    expect(channelEntity.config['push_level'], model.config.pushLevel);
+    expect(channelEntity.config['push_notifications'], model.config.pushNotifications);
+    expect(
+      channelEntity.config['chat_preferences'],
+      {
+        'default_preference': ChatPreferenceLevel.all,
+        'direct_mentions': ChatPreferenceLevel.none,
+      },
+    );
```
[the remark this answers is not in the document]
VelikovPetar: I don't think we should really change this: The `ChatPreferenceLevel` is an `extension type` on `String`, so here we are actually already verifying against a String.

packages/stream_chat/lib/src/core/api/responses.dart
```diff
@@ -788,6 +801,87 @@ class GetActiveLiveLocationsResponse extends _BaseResponse {
...
+
+/// Model response for [StreamChatClient.searchRoles] api call
+@JsonSerializable(createToJson: false)
+class SearchRolesResponse extends _BaseResponse {
+  /// List of roles returned by the api call
+  late List<Role> roles;
+
+  /// Create a new instance from a json
+  static SearchRolesResponse fromJson(Map<String, dynamic> json) => _$SearchRolesResponseFromJson(json);
+}
```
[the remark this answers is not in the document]
VelikovPetar: Will fix!

packages/stream_chat/lib/src/core/api/user_groups_api.dart
```diff
@@ -0,0 +1,162 @@
...
+    String? teamId,
+  }) async {
+    final response = await _client.put(
+      '/usergroups/$id',
+      data: {
+        if (name != null) 'name': name,
+        if (description != null) 'description': description,
+        if (teamId != null) 'team_id': teamId,
+      },
+    );
```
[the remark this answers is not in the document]
VelikovPetar: I wouldn't apply this. If the BE changes this, we might end up enforcing something which is incorrect. It would be better to let the server handle this error.

packages/stream_chat/lib/stream_chat.dart
```diff
@@ -39,6 +39,7 @@ export 'src/core/models/channel_config.dart';
 export 'src/core/models/channel_model.dart';
 export 'src/core/models/channel_mute.dart';
 export 'src/core/models/channel_state.dart';
+export 'src/core/models/chat_preferences.dart';
```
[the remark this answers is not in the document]
VelikovPetar: This follows the exact pattern already established - we should follow it

packages/stream_chat/lib/src/core/models/channel_config.dart
```diff
@@ -21,6 +23,9 @@ class ChannelConfig {
     this.replies = false,
     this.search = false,
     this.polls = false,
+    this.pushLevel,
+    this.pushNotifications = true,
```
xsahil03x, line 27: this should be `false` by default right?

packages/stream_chat/lib/src/core/models/role_type.dart
xsahil03x, line 1: I think we can have it in the same `role.dart` file

packages/stream_chat/test/src/core/models/message_test.dart
```diff
@@ -105,27 +127,53 @@ void main() {
       expect(json['text'], equals('Hello world'));
     });
 
-    test('should remove mentioned users not found in text', () {
+    test('should remove mentions not found in text', () {
```
xsahil03x: Do we need to cover any other cases or combination of mentions here?

packages/stream_chat_flutter/lib/src/mention/mention.dart
```diff
@@ -0,0 +1,92 @@
+import 'package:stream_chat_flutter/stream_chat_flutter.dart';
+
+/// A typed mention surfaced by the read-side render pipeline.
+///
+/// One of the five built-in subclasses: [UserMention], [ChannelMention],
+/// [HereMention], [RoleMention], [GroupMention].
+abstract class Mention {
```
xsahil03x: Lets name it `StreamMention`

packages/stream_chat_flutter/lib/src/message_input/stream_message_composer.dart
```diff
@@ -165,7 +176,12 @@ class MessageComposerProps {
     this.enableVoiceRecording = false,
     this.sendVoiceRecordingAutomatically = false,
     this.voiceRecordingFeedback = const AudioRecorderFeedback(),
-    this.userMentionsTileBuilder,
+    @Deprecated('Use mentionUserTileBuilder instead') this.userMentionsTileBuilder,
+    this.mentionChannelTileBuilder,
+    this.mentionHereTileBuilder,
+    this.mentionRoleTileBuilder,
+    this.mentionUserGroupTileBuilder,
+    this.mentionUserTileBuilder,
```
xsahil03x: Do we need per tile builders here? I would create a simple one with Mention as the param, similar to the `onMentionTap` in `MessageListView`

packages/stream_chat_flutter/lib/src/message_widget/stream_message_item.dart
```diff
@@ -1092,3 +1121,21 @@ class _StreamMessageItemDefaults extends core.StreamMessageItemThemeData {
...
+  return switch (type) {
+    core.MentionType.user =>
+      message.mentionedUsers.firstWhereOrNull((u) => u.id == id)?.let((user) => UserMention(user: user)),
+    core.MentionType.channel => const ChannelMention(),
+    core.MentionType.here => const HereMention(),
+    core.MentionType.role =>
+      message.mentionedRoles?.firstWhereOrNull((r) => r == id)?.let((role) => RoleMention(role: role)),
+    core.MentionType.group =>
+      message.mentionedGroups?.firstWhereOrNull((g) => g.id == id)?.let((g) => GroupMention(userGroup: g)),
+    _ => null,
```
xsahil03x: We can use `dot shorthands` here.

```suggestion
.user => message.mentionedUsers.firstWhereOrNull((u) => u.id == id)?.let((user) => UserMention(user: user)),
.channel => const ChannelMention(),
.here => const HereMention(),
.role => message.mentionedRoles?.firstWhereOrNull((r) => r == id)?.let((role) => RoleMention(role: role)),
.group => message.mentionedGroups?.firstWhereOrNull((g) => g.id == id)?.let((g) => GroupMention(userGroup: g)),
_ => null,
```

packages/stream_chat_flutter/pubspec.yaml
```diff
@@ -58,17 +58,9 @@ dependencies:
...
   stream_chat_flutter_core: ^10.0.1
   stream_core_flutter:
-    # The ignore below silences `invalid_dependency` because we occasionally
-    # pin stream_core_flutter to a git ref to iterate on it alongside this
-    # SDK between its releases.
-    #
-    # **Note:** Before publishing stream_chat_flutter, this MUST be swapped
-    # back to a pub version constraint — git deps are not allowed on pub.dev
-    # and will block the release.
-    # ignore: invalid_dependency
```
xsahil03x, line 68: Lets keep the comment and ignore here.

packages/stream_chat_flutter/lib/src/autocomplete/stream_mention_autocomplete_options.dart
```diff
@@ -149,27 +230,80 @@ class _StreamMentionAutocompleteOptionsState extends State<StreamMentionAutocomp
...
-    }
+  Future<List<_MentionOption>> queryMentions(String query) async {
+    final broadcasts = _broadcastMentions(query);
+    final rolesFuture = _fetchRoles(query);
+    final groupsFuture = _fetchUserGroups(query);
+    final usersFuture = _fetchUsers(query);
+
+    final roles = await rolesFuture;
+    final groups = await groupsFuture;
+    final users = await usersFuture;
```
xsahil03x: We can do it parallelly using `.wait`.


```suggestion
    final (roles, groups, users) = await (rolesFuture, groupsFuture, usersFuture).wait;
```

packages/stream_chat/lib/src/core/models/role_type.dart
VelikovPetar: Of course, I will move it!

packages/stream_chat_flutter/lib/src/mention/mention.dart
VelikovPetar: Makes sense, especially with the rename `MentionType -> StreamMentionType` in `core`

packages/stream_chat_flutter/lib/src/message_input/stream_message_composer.dart
VelikovPetar: I took this approach with the idea that if a customer would like to customise only one tile (for example just the `channel` mention tile) => it will be easier to override a specific builder, instead of overriding the general builder, in which they will have to also take care of invoking the default builder as fallback. (this is possible, but IMO a bit more complex for customers). But if you think it is the better approach, I would be happy to change it.

packages/stream_chat/lib/src/core/models/channel_config.dart
VelikovPetar: This flag is actually set to `true` by default for all built-in channel types (`messaging`, `livestream`, `gaming`, `commerce`, `team`), and `false` for custom channel types. But it is anyway a mandatory property of the JSON, so it will always be returned by the server - in real live scenarios this default should never really be assigned.

packages/stream_chat/test/src/core/models/message_test.dart
VelikovPetar: The test covers the case where we have multiple mentions of the same type in a single message => and one mention of each type is preserved, while another mention (of the same type) is removed. Do you think we should add tests for some other scenarios? (for example separate tests for each mention type...)

packages/stream_chat_flutter/lib/src/mention/mention.dart
VelikovPetar: What do you think about its children: Should we rename them as well? (`RoleMention` -> `StreamRoleMention`)

xsahil03x: yes, lets also rename the childrens

packages/stream_chat_flutter/lib/src/message_input/stream_message_composer.dart
xsahil03x: It feels okay, if we had this only in the command autocomplete components but having them here kinds of pollutes the message composer api. 

>it will be easier to override a specific builder, instead of overriding the general builder, in which they will have to also take care of invoking the default builder as fallback.

About this, i think we can solve it easily by extracting the mention item in a separate widget and using the props and factory component pattern in it. Then we can simply use `.fromProps` if we want the default impl as a fallback

packages/stream_chat/test/src/core/models/message_test.dart
xsahil03x: Good to have separate tests but i am also fine with this.

packages/stream_chat_flutter/lib/src/mention/mention_tile.dart
```diff
@@ -0,0 +1,277 @@
...
+
+/// Renders a single mention suggestion row.
+///
+/// Customise rendering globally by registering a [StreamComponentBuilder]
+/// for [StreamMentionTileProps] through [streamChatComponentBuilders]. When
+/// no builder is registered the [DefaultStreamMentionTile] fallback is used.
+///
+/// For per-instance overrides, set `mentionTileBuilder` on
+/// [StreamMentionAutocompleteOptions] or [StreamMessageComposer].
+class StreamMentionTile extends StatelessWidget {
```
xsahil03x: Let's name it `Item`, we had to rename our ChannelListTile and other components to item in the design refresh.

packages/stream_chat_flutter/lib/src/mention/mention_tile.dart
```diff
@@ -0,0 +1,277 @@
...
+    if (mention is StreamRoleMention) {
+      return _MentionRoleTile(role: mention.role, onTap: onTap);
+    }
+    if (mention is StreamGroupMention) {
+      return _MentionUserGroupTile(userGroup: mention.userGroup, onTap: onTap);
+    }
+    if (mention is StreamUserMention) {
+      return _MentionUserTile(user: mention.user, onTap: onTap);
+    }
+    return _MentionFallbackTile(display: mention.display, onTap: onTap);
```
xsahil03x: :nit: we can use switch statement here.

xsahil03x (approved):
left some smaller comments but LGTM ✅

PR #2748, opened 2026-06-12
feat(ui): channel page and floating UI

renefloor:
# Submit a pull request

Linear: FLU-502

## CLA

This introduces the floating style on all sample app pages and moves the channel page and thread page to the sdk.
The `StreamMessageListViewConfiguration` is also added to the global chat config so it's easier to use in the channel and thread page.

Core PR: https://github.com/GetStream/stream-core-flutter/pull/146

## Screenshots / Videos

| Regular | Floating |
| --- | --- |
| image | image |
| image | image |
| image  | image |
| image | image |
| image | image |

## Summary by CodeRabbit

- **New Features**
  - Added ready-to-use channel and thread pages with reply, edit, navigation, and composer support.
  - Added floating or docked composer layouts with customizable themes.
  - Added global and per-view message-list configuration, padding controls, and reaction callbacks.
  - Added improved floating app-bar and avatar presentation.
  - Added optional location sharing in the sample app.

- **Bug Fixes**
  - Thread replies are hidden when replies are disabled.
  - Long deleted-message labels now wrap correctly.
  - Thread scroll-to-bottom controls reflect effective configuration.
  - Photo galleries respect bottom safe-area spacing.


packages/stream_chat_flutter/lib/src/message_input/attachment_picker/options/stream_gallery_picker.dart
```diff
@@ -75,7 +75,7 @@ class _StreamGalleryPickerState extends State<StreamGalleryPicker> {
     return FutureBuilder<PermissionState>(
       future: requestPermission,
       builder: (context, snapshot) {
-        if (!snapshot.hasData) return const Empty();
+        if (!snapshot.hasData) return const SizedBox.expand();
```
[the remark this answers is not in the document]
renefloor: This was introduced, because with the `Empty` made the background not fill from left to right when the permission dialog was shown, so the expand is intentional. Does this give any other issue? 
with box example

docs/docs_screenshots/test/theming/goldens/macos/theming_red.png
renefloor, line 1: This is from the HSL -> HCT color change

packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
```diff
@@ -273,8 +275,10 @@ class StreamMessageListView extends StatefulWidget {
   /// [StreamMessageListViewConfiguration.markReadWhenAtTheBottom], scroll
   /// physics, and other non-builder, non-theme settings.
   ///
-  /// Defaults to [StreamMessageListViewConfiguration] with all defaults.
-  final StreamMessageListViewConfiguration config;
+  /// When null, falls back to
+  /// [StreamChatConfigurationData.messageListViewConfiguration] from the
+  /// nearest [StreamChatConfiguration] ancestor.
+  final StreamMessageListViewConfiguration? config;
```
xsahil03x, line 279: This is kind of breaking

packages/stream_chat_flutter/lib/src/theme/message_composer_theme.dart
```diff
@@ -0,0 +1,130 @@
...
+/// When null on [StreamMessageComposerThemeData], the ambient [StreamAppStyle]
+/// is used as a fallback — [StreamAppStyle.floating] maps to [floating] and
+/// [StreamAppStyle.regular] maps to [docked].
+///
+/// See also:
+///
+///  * [StreamMessageComposerThemeData.location], which carries this
+///    value.
+///  * [StreamAppStyle], the global app-wide style that acts as fallback.
+enum StreamComposerLocation {
```
xsahil03x: We should name it `StreamMessageComposerBehavior` with values `regular` and `floating` to match the appbar and navbar.

packages/stream_chat_flutter/lib/src/channel/channel_header.dart
```diff
@@ -177,10 +186,15 @@ class StreamChannelHeader extends StatelessWidget implements PreferredSizeWidget
...
+            // typography, divider).
+            //
+            // [style] is merged into the theme data as well as passed to the
+            // bar, so every slot below — the default avatar, the default back
+            // button, and anything the caller supplied — resolves the same
+            // chain the bar itself does: [style], then the per-header theme,
+            // then the ambient app bar theme, then the app style.
             child: StreamAppBarTheme(
-              data: headerTheme,
+              data: headerTheme.merge(StreamAppBarThemeData(style: style)),
```
xsahil03x: do we need to pass the style in the data? we are already passing it in the StreamAppBar below

packages/stream_chat_flutter/lib/src/channel/channel_list_header.dart
```diff
@@ -158,10 +159,15 @@ class StreamChannelListHeader extends StatelessWidget implements PreferredSizeWi
...
+            // theme drives all default styling.
+            //
+            // [style] is merged into the theme data as well as passed to the
+            // bar, so every slot below — the default avatar and anything the
+            // caller supplied — resolves the same chain the bar itself does:
+            // [style], then the per-header theme, then the ambient app bar
+            // theme, then the app style.
             child: StreamAppBarTheme(
-              data: headerTheme,
+              data: headerTheme.merge(StreamAppBarThemeData(style: style)),
```
xsahil03x: do we need to pass the style in the data? we are already passing it in the StreamAppBar below

packages/stream_chat_flutter/lib/src/message_action/message_actions_builder.dart
```diff
@@ -143,7 +144,7 @@ class StreamMessageActionsBuilder {
     // Thread reply action is only available for parent messages that are not in a
     // thread view, as replying in a thread that is already being viewed doesn't make sense.
     // Additionally, the channel needs to support sending replies.
-    if (canSendReply && !isThreadMessage && !isInThreadView) {
+    if (canSendReply && repliesEnabled && !isThreadMessage && !isInThreadView) {
```
xsahil03x: Is `canSendReply` was not enough in this case?

packages/stream_chat_flutter/lib/src/message_input/stream_message_composer.dart
```diff
@@ -930,69 +989,98 @@ class DefaultStreamMessageComposerState extends State<DefaultStreamMessageCompos
...
+          canPop: !_isPickerVisible,
+          onPopInvokedWithResult: (didPop, _) {
+            if (!didPop) _hidePicker();
+          },
+          child: Column(
+            mainAxisSize: MainAxisSize.min,
+            // Reversed paint order so the pill (and its shadow) paints on top
+            // of the picker panel. VerticalDirection.up keeps the pill visually
+            // above the picker while making it the last-painted child.
+            verticalDirection: VerticalDirection.up,
```
xsahil03x: Not sure if this was needed here, we can just switch the position of picker and composer like we had earlier. Its also better to read in the correct order.

packages/stream_chat_flutter/lib/src/message_input/stream_message_composer.dart
```diff
@@ -1007,29 +1095,31 @@ class DefaultStreamMessageComposerState extends State<DefaultStreamMessageCompos
...
+            optionsBuilder: widget.props.attachmentPickerOptionsBuilder,
+            onError: _onPickerError,
+            onPollCreated: _onPollCreated,
+            onCommandSelected: _onCommandSelected,
+            commandValidator: _effectiveController.validateCommand,
+          );
+
+    // No background here: when floating, the panel sits on the composer-wide
+    // backdrop painted by _FloatingComposerBackdropPainter.
+    return child;
```
xsahil03x: Let's also revert it, the early return style is read better

packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
```diff
@@ -800,22 +845,32 @@ class _StreamMessageListViewState extends State<StreamMessageListView> {
...
+              stream: streamChannel!.channel.state!.isUpToDateStream,
+              initialData: streamChannel!.channel.state!.isUpToDate,
+              builder: (context, snapshot) => ValueListenableBuilder<bool>(
+                valueListenable: _showScrollToBottom,
+                child: _buildScrollToBottom(),
+                builder: (context, value, child) {
+                  if (!snapshot || value) return child!;
+                  return const Empty();
+                },
+              ),
```
xsahil03x, line 867: Is the value of `isUpToDate` not correct in case of thread conversations?

packages/stream_chat_flutter/lib/src/misc/back_button.dart
```diff
@@ -39,6 +41,12 @@ class StreamBackButton extends StatelessWidget {
   )
   final String? channelId;
 
+  /// Controls the back button's visual/layout behavior (floating vs regular).
+  ///
+  /// When null, falls back to [StreamAppBarStyle.appBarBehavior] from the
+  /// ambient [StreamAppBarTheme], then to the ambient [StreamAppStyle].
+  final StreamAppBarBehavior? appBarBehavior;
```
xsahil03x: we should pass `isFloating` directly similar to buttons in core rather than passing the `appBarBehavior`

packages/stream_chat_flutter/lib/src/misc/thread_header.dart
```diff
@@ -93,8 +100,12 @@ class StreamThreadHeader extends StatelessWidget implements PreferredSizeWidget
       alternativeWidget: fallbackSubtitle,
     );
 
+    // [style] is merged into the theme data as well as passed to the bar, so
+    // every slot below — the default back button and anything the caller
+    // supplied — resolves the same chain the bar itself does: [style], then the
+    // per-header theme, then the ambient app bar theme, then the app style.
     return StreamAppBarTheme(
-      data: headerTheme,
+      data: headerTheme.merge(StreamAppBarThemeData(style: style)),
```
xsahil03x: do we need to pass the style in the data? we are already passing it in the StreamAppBar below

sample_app/lib/config/sample_app_config.dart
```diff
@@ -3,11 +3,25 @@ import 'package:stream_chat_flutter/stream_chat_flutter.dart';
 import 'package:stream_chat_localizations/stream_chat_localizations.dart';
 import 'package:streaming_shared_preferences/streaming_shared_preferences.dart';
 
+// ---------------------------------------------------------------------------
+// AppStyle enum
+// ---------------------------------------------------------------------------
+
+/// The visual style for the app's UI chrome (app bar, composer, bottom bar).
+enum SampleAppStyle {
```
xsahil03x: We can use `StreamAppStyle` from the core here directly

packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
renefloor: Why? If you supplied it previously it's still going to work with what you supply. If you didn't supply anything it's still falling back to the default. Or do you think anybody will use it as getter?

packages/stream_chat_flutter/lib/src/theme/message_composer_theme.dart
renefloor: I based this on iOS, which names it location. But I'm fine with behavior to align with the other similar properties.

sample_app/lib/config/sample_app_config.dart
renefloor: Yeah I agree, I made this before core was an enum.

packages/stream_chat_flutter/lib/src/channel/channel_header.dart
renefloor: StreamAppBar's own style param only styles the bar itself. The header's default slot children — _DefaultChannelAvatar and StreamBackButton — are built outside the bar and resolve floating-ness from context via isFloatingAppBar(context) → StreamAppBarTheme.of(context).style?.behavior. Without the merge, StreamChannelHeader(style: ...behavior: floating) would render a floating bar with a regular (shadowless) avatar and a ghost instead of outline back button. This was exactly CodeRabbit's earlier finding (comment 3683952384), fixed by merging into the theme instead of threading isFloating through every slot.
 
 We can instead remove the `style` from the `StreamAppBar` below.

packages/stream_chat_flutter/lib/src/message_action/message_actions_builder.dart
renefloor: They're two different gates. canSendReply is the per-user own-capability (ChannelCapability.sendReply); repliesEnabled is channel.config.replies, the channel-type feature flag. The two can disagree — a channel type with threads turned off can still hand out the send-reply capability, which is what left the "Thread reply" action visible. There's a test pinning this (33eeab5): capability granted, replies: false → action must not appear (it fails if the guard is removed).

Compared with other SDK's:
```
iOS — capability wins, config is only a fallback

Sources/StreamChatUI/MessageActionsPopup/ChatMessageActionsVC.swift:111:
// If a channel is not set, we fallback to using channelConfig only.
let canSendReply = channel?.canSendReply ?? channelConfig.repliesEnabled
...
if canSendReply && !message.isPartOfThread && !isInsideThread {
    actions.append(threadReplyActionItem())
}
So when the channel is available (the normal case), config.replies is ignored entirely for this action — canSendReply is ownCapabilities.contains(.sendReply) (Channel.swift:698). iOS does use the config flag elsewhere though: ChatMessageLayoutOptionsResolver.swift:89 gates the thread-reply footer on channel.config.repliesEnabled.

Android UI-Components (XML) — ANDs both, exactly like this PR

MessageListView.kt:377:
val isThreadEnabled = style.threadsEnabled && channel.config.isThreadEnabled
fed into CapabilitiesHelper.kt:92:
threadsEnabled && !isInThread && !message.isThreadReply() && message.isSynced() &&
    ownCapabilities.contains(ChannelCapabilities.SEND_REPLY)
And Config.isThreadEnabled is mapped straight from the API's replies field (DomainMapping.kt:672). This is the same gate as canSendReply && repliesEnabled, plus a UI-level style toggle.

Android Compose — capability + a UI visibility flag, no channel config

MessageActionsOptionsVisibility.kt:44 passes threadsEnabled = isThreadReplyVisible (a developer-facing visibility flag), so config.replies never enters the Compose path.

---
What this means for the review comment: Sahil's instinct matches iOS and Compose; the PR matches Android XML. None of the three treats the capability as strictly sufficient and uses the config flag nowhere — both platforms consult config.replies for thread UI somewhere (iOS's footer, Android's FootnoteDecorator), which suggests the flag is treated as meaningful client-side rather than fully folded into capabilities server-side.

One parity gap worth mentioning separately: our thread-reply indicator (stream_message_item.dart:503) is gated only on replyCount > 0, not on repliesEnabled — iOS and Android both gate their equivalent on the config flag. So if we keep the guard, it arguably belongs there too; if we drop it, we're consistent with iOS's action list but diverge from Android.
```

packages/stream_chat_flutter/lib/src/message_input/stream_message_composer.dart
renefloor: It was needed for the draw order. The shadows of the composer was clipped by the background of the attachment picker. However, we now have 1 gradient behind both the attachment picker and the composer, so it can indeed be reversed. This was one of the first implementations: 
image

packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
renefloor: streamChannel.channel.state.isUpToDate tracks the channel's message window (false when you've jumped to an older message and the latest page isn't loaded). A thread's reply list is a separate list with no equivalent flag. So opening a thread after jumping to an old channel message left isUpToDate == false, and the thread's scroll-to-bottom button never appeared even though the thread list was perfectly scrollable. In thread mode the button is gated on _showScrollToBottom alone.

The branch is arguably ugly — an alternative is a single BetterStreamBuilder whose gate is _isThreadConversation || isUpToDate, which collapses the if/else into one widget tree.

sample_app/lib/config/sample_app_config.dart
renefloor: fixed in https://github.com/GetStream/stream-chat-flutter/pull/2748/commits/217be1a4ee7b52d00595913ab8b0e80468a5ac44

packages/stream_chat_flutter/lib/src/misc/back_button.dart
renefloor: fixed in https://github.com/GetStream/stream-chat-flutter/pull/2748/commits/217be1a4ee7b52d00595913ab8b0e80468a5ac44

packages/stream_chat_flutter/lib/src/message_input/stream_message_composer.dart
renefloor: fixed in https://github.com/GetStream/stream-chat-flutter/pull/2748/commits/217be1a4ee7b52d00595913ab8b0e80468a5ac44

packages/stream_chat_flutter/lib/src/message_input/stream_message_composer.dart
renefloor: fixed in https://github.com/GetStream/stream-chat-flutter/pull/2748/commits/217be1a4ee7b52d00595913ab8b0e80468a5ac44

packages/stream_chat_flutter/lib/src/theme/message_composer_theme.dart
renefloor: fixed in https://github.com/GetStream/stream-chat-flutter/pull/2748/commits/217be1a4ee7b52d00595913ab8b0e80468a5ac44

packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
xsahil03x: >anybody will use it as getter

it was non-nullable before so yeah kind of breaking

packages/stream_chat_flutter/lib/src/message_action/message_actions_builder.dart
xsahil03x: I checked with the backend source and it said,
```
⏺ Verified against the backend — and the source contradicts renefloor's central claim. The send-reply capability is itself gated on the replies config:

  lib/combined/internal/permissionutil/channel_capabilities.go:73
  func SendReply(session *ChannelSession) bool {
        if session.Channel.Frozen && !session.Check(permissions.UseFrozenChannel).Allowed {
                return false
        }       
        return session.Channel.Config.Replies && (session.Check(permissions.CreateReply).Allowed                                                                                                  
                || session.Check(permissions.CreateMessage).Allowed)
  }

  And Config.Replies is exactly the client's channel.config.replies:
  lib/combined/types/channel_config.go:51 → Replies bool json:"replies".

  So the server only ever puts send-reply in own_capabilities when config.replies == true. There is no path where "a channel type with threads turned off still hands out the send-reply capability" — the && 
  Config.Replies prevents it (the only other exit is the frozen check, also false). Contrast UpdateThread at :172, which is pure permission with no config gate — so the backend does couple send-reply to the config
  on purpose.

  What this means:
  - Your instinct is correct. canSendReply already implies repliesEnabled, so canSendReply && repliesEnabled is redundant — the second term is always true whenever the first is.
  - No realistic disagreement exists. own_capabilities and config arrive in the same channel response, and the server computes the capability from the config — so they can't drift apart. renefloor's pinning test
  (capability granted + replies: false) is a hand-built state the backend never emits.
  - Cross-SDK: this matches iOS's primary gate (canSendReply); Android XML's extra AND is belt-and-suspenders against that same impossible state.

  So dropping the repliesEnabled guard is safe and actually more faithful to the backend contract. renefloor's SDK-parity survey is useful, but its premise — that the capability and the config are independent
  gates that "can disagree" — doesn't hold against the source; the backend already folded the config into the capability.

  One fair caveat to acknowledge back to him: his separate point still stands — the thread-reply indicator (stream_message_item.dart:503, gated only on replyCount > 0) is a real parity gap independent of this, and
  worth its own look.

```

renefloor: changed: https://github.com/GetStream/stream-chat-flutter/pull/2748/changes/d07bcf5c4154ed4084d4e179fbc3bffc171cba1d

packages/stream_chat_flutter/example/ios/Runner.xcodeproj/xcshareddata/xcschemes/Runner.xcscheme.backup
renefloor, line 1: Can you remove this file?

xsahil03x (approved):

PR #2750, opened 2026-06-12
perf(persistence): Move thread messages pagination to SQL

VelikovPetar:
# Submit a pull request

Linear: [FLU-487](https://linear.app/stream/issue/FLU-487)

Github Issue: #

## CLA

Back-port of #2688 to v9. Refactors `MessageDao.getThreadMessagesByParentId` to filter and paginate thread replies at the SQL level instead of loading the full thread into memory, and adds support for all four `PaginationParams` cursor variants (`lessThan` / `lessThanOrEqual` / `greaterThan` / `greaterThanOrEqual`) with a `(createdAt, id)` tiebreaker.


packages/stream_chat_persistence/lib/src/dao/message_dao.dart
```diff
@@ -202,44 +202,89 @@ class MessageDao extends DatabaseAccessor<DriftChatDatabase>
...
+      query.where(
+        messages.createdAt.isBiggerThanValue(c.createdAt) |
+            (messages.createdAt.equals(c.createdAt) &
+                messages.id.isBiggerOrEqualValue(c.id)),
+      );
+    }
+
+    if (options != null) {
+      query.limit(options.limit);
+    }
```
renefloor: Is this intended?

- limit: 0 behavior changed. Old code: if (limit != null && limit > 0) return mutable.take(limit)... ; return mutable; — so limit: 0 was treated as "unlimited" and returned everything. New code: if (options != null) query.limit(options.limit); — LIMIT 0 in SQL returns zero rows. Since PaginationParams.limit defaults to 10 and no call site in this repo passes 0 explicitly, this is unlikely to matter in practice, but it is a silent behavior change from the pre-port implementation.

VelikovPetar: Hmm good catch - I can't say that it was intentional. 
It is a bit tricky to say what should be correct here:
- Does calling the method with `options: null` means return all thread replies?
- Does calling the method with `options: PaginationParams(limit: 0)` also means return all messages? 

While I prefer the first approach a bit more, perhaps we should revert this to the pre-refactor behaviour. While the default SDK behaviour should not be affected by this, it could potentially affect a customer calling the `persistenceClient.getReplies` method directly.

(this would also require a fix on `master` as well, because there we have already merged this change - https://github.com/GetStream/stream-chat-flutter/pull/2688)

renefloor (approved):

packages/stream_chat_persistence/lib/src/dao/message_dao.dart
VelikovPetar: addressed here: https://github.com/GetStream/stream-chat-flutter/pull/2750/commits/113fa2acd998a3bfca4ea27ef5994936b3f6531d

PR #2759, opened 2026-06-17
fix(llc, core): pause WebSocket reconnect while app is backgrounded

xsahil03x:
## Summary

- Suspends the WebSocket's auto-retry loop while the app is backgrounded, so an OS-initiated socket close inside the `backgroundKeepAlive` window no longer kicks off a retry storm (and no longer surfaces an unhandled `WebSocketChannelException` when DNS / network is restricted in background).
- New `StreamChatClient.pauseReconnect` / `resumeReconnect` API; wired from `StreamChatCore._ChatLifecycleManager` on the existing background / foreground transitions. Foreground path resumes retries and calls `maybeReconnect` as before.
- Reordered `WebSocket._reconnect` so the dead-channel cleanup (`_stopMonitoringEvents`, `_closeWebSocketChannel`) always runs, even when retry is paused — prevents the health-check timer from firing on a closed sink during the paused window.
- Attached a no-op error handler to `WebSocketChannel.sink.done` in `_initWebSocketChannel`; the duplicate error delivery there was the source of the unhandled `WebSocketChannelException`, since the stream's `onError` already handles it.

## Test plan

- [x] `dart test packages/stream_chat/test/src/ws/websocket_test.dart` — new test exercises a paused-close while pumping past `healthCheckInterval` to assert no sink writes happen on the dead channel, plus the expected status transitions.
- [x] `flutter test packages/stream_chat_flutter_core/test/stream_chat_core_test.dart` — new lifecycle test verifies `pauseReconnect` on background and `resumeReconnect` on foreground.
- [x] `dart analyze --fatal-infos` on changed files — clean.
- [x] Manual verification on Android: confirmed via app logs that backgrounding now produces `Pausing reconnection`, an OS-driven close inside the keep-alive window logs `Reconnect skipped: paused` with no retries until foreground resumes the connection cleanly.

## Release Notes

* **New Features**
  * Added `pauseReconnect()` and `resumeReconnect()` APIs to control WebSocket auto-reconnection
  * Automatically pauses/resumes reconnection on app background/foreground transitions, including the background keep-alive window

* **Bug Fixes**
  * Prevented unhandled errors during reconnect failures by handling duplicate completion signals
  * Avoided reconnect attempts when reconnection is paused during a pending delayed reconnect

* **Tests**
  * Updated WebSocket sink mocking and added reconnection-focused tests for pause/resume and lifecycle behavior

### DIFF
--- a/packages/stream_chat/lib/src/ws/websocket.dart
+++ b/packages/stream_chat/lib/src/ws/websocket.dart
@@ -285,9 +285,8 @@ class WebSocket with TimerHelper {
   void _reconnect({bool refreshToken = false}) async {
     if (_reconnectRequestInProgress) return;
 
-    // Tear down the dead channel and its timers up front so the health-check
-    // timer can't fire on a closed sink while we wait — both when retrying and
-    // when reconnect is paused.
+    // Must run before the pause check so the health-check timer can't fire
+    // on a closed sink while we wait.
     _stopMonitoringEvents();
     _closeWebSocketChannel();
 
--- a/packages/stream_chat/lib/src/ws/websocket.dart
+++ b/packages/stream_chat/lib/src/ws/websocket.dart
@@ -134,6 +134,10 @@ class WebSocket with TimerHelper {
       _closeWebSocketChannel();
     }
     _webSocketChannel = webSocketChannelProvider?.call(uri) ?? WebSocketChannel.connect(uri);
+    // Connection failures (e.g. DNS errors during background) are delivered
+    // via the stream's onError below; ignore the duplicate signal on sink.done
+    // so it doesn't surface as an unhandled future error.
+    _webSocketChannel?.sink.done.ignore();
     _subscribeToWebSocketChannel();
   }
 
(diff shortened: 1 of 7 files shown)


packages/stream_chat/lib/src/ws/websocket.dart
```diff
@@ -255,22 +259,52 @@ class WebSocket with TimerHelper {
...
+    if (!_reconnectEnabled) return;
+    _logger?.info('Pausing reconnection');
+    _reconnectEnabled = false;
+  }
+
+  /// Re-enables automatic reconnection after a previous [pauseReconnect].
+  void resumeReconnect() {
+    if (_reconnectEnabled) return;
+    _logger?.info('Resuming reconnection');
+    _reconnectEnabled = true;
```
renefloor, line 282: Should we also trigger a reconnect here?

xsahil03x: No, it should be fine, we already handle it in core where we actually call it. If a integrator is only llc they can decide if they want to call it or not.

renefloor (approved):

PR #2774, opened 2026-06-19
fix(ui): preserve last-message preview during channel-state reloads

VelikovPetar:
# Submit a pull request

Linear: FLU-547

## CLA

### Summary

When the user opens a channel that has unread messages, the corresponding channel-list cell briefly renders the empty-state text ("No messages yet") before snapping back to the real last-message preview. This PR fixes the flicker and adds regression tests for the underlying state-emission scenarios.

### The bug

`StreamChannel._maybeInitChannel` calls `loadChannelAtMessage(lastReadMessageId)` → `_queryAtMessage` → `Channel.query(messagesPagination: PaginationParams(idAround: …))` when entering a channel with unreads (`packages/stream_chat_flutter_core/lib/src/stream_channel.dart`).

Inside `Channel.query` (`packages/stream_chat/lib/src/client/channel.dart`), the `idAround` path runs `state.truncate()` (which clears `messages` to `[]` and emits a new `ChannelState`) immediately followed by `updateChannelState(...)` (which emits the populated state). Both emissions go through `_channelStateController` (a `BehaviorSubject`) and reach listeners in separate microtasks — so every subscriber to `messagesStream` sees an empty list, then the populated list.

The channel-list cell's preview widget (`_ChannelLastMessageWithStatus` in `stream_channel_list_item.dart`, plus its public twin `ChannelLastMessageText`) was relying on a `Message? _currentLastMessage` cache field to ride out exactly this kind of transient empty emission via a `.latest` selector. The cache field was declared but never assigned, so the absorber was a no-op and the empty emission produced the visible flash.

### The fix

Assign `_currentLastMessage = message` (including `null`) on every build, gated on `channelState.isUpToDate`. This is done **before** computing `.latest`, so the freshly-stored value participates in the selector.

The gate matters: `_queryAtMessage` flips `isUpToDate = false` before truncating, so the transient empty list is skipped by the cache. A real `channel.truncated` event (`_listenChannelTruncated`) clears messages while leaving `isUpToDate = true`, so the cache correctly clears to `null` and the empty-state surfaces.

### Tests

New file: `packages/stream_chat_flutter/test/src/scroll_view/channel_scroll_view/stream_channel_list_item_test.dart`. Three scenarios:

- Preserves the last-known message when state is not up-to-date and emits empty.
- Rebinding the widget to a different channel shows that channel's state, not the previous one's.
- Shows the empty-state when the channel is truncated while up-to-date.

### Manual verification

- Open a channel with unread messages from the channel list → preview no longer flashes the empty-state.
- Truncate a channel (admin action) → preview switches to "No messages yet".
- Hard-delete the last message → preview switches to "No messages yet".

## UI Changes

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/74704fc3-ce88-4f1b-8053-0e5e96b2fa49" controls></video> | <video src="https://github.com/user-attachments/assets/f88a5a63-ba38-4321-91a0-fbcd5cdce762" controls></video> |

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved flickering of the last-message preview during channel-state reloads
  * Improved last-message preview accuracy when channel data isn’t up to date, including correct subtitle updates when rebinding and when showing empty-state text after truncation
* **Tests**
  * Added widget coverage for subtitle behavior across up-to-date, rebinding, and truncation scenarios
* **Documentation**
  * Updated the package changelog with the fix under Upcoming/Fixed


packages/stream_chat_flutter/lib/src/scroll_view/channel_scroll_view/stream_channel_list_item.dart
```diff
@@ -624,6 +624,13 @@ class _ChannelLastMessageWithStatusState extends State<_ChannelLastMessageWithSt
 
         // Find the last valid message.
         final message = messages.lastWhereOrNull(_defaultLastMessagePredicate);
+        // Cache only while the channel has the latest messages loaded
+        // (isUpToDate). When isUpToDate is false (e.g. Channel.query(idAround:)
+        // truncates state mid-load), the previous value keeps rendering instead
+        // of false rendering the empty state.
+        if (channelState.isUpToDate) {
```
xsahil03x, line 633: Should we move this after we calculate `latestLastMessage`? and set the `_currentLastMessage` to `latestLastMessage`?

VelikovPetar: There is an interesting edge-case which kinda prevents us doing this: While the `_currentLastMessage ` will be properly populated in both cases, there is a subtle difference in the calculation of `latestLastMessage`, visible in the case where a channel is actually truncated (all messages removed). 
If we move the statement after calculating the new `latestLastMessage`, the `latestLastMessage` will take into consideration the _old_ `_currentLastMessage`. And for the truncation case, it means that it will wrongfully be calculated as `final latestLastMessage = [message (null), _currentLastMessage (oldCachedMessage)].latest;` - which would be wrong. 
By keeping the `if (channelState.isUpToDate) _currentLastMessage = message;` before calculating `latestLastMessage`, it means that the new latest message will also be taken into consideration for the calculation: `final latestLastMessage = [message (null), _currentLastMessage (null)].latest;`

The concrete example where this logic fails:

- If we assume a real channel truncation (`channel.truncated` event) - (`messages = []`, `isUpToDate=true`, `cache=msg1` from the previous build):                                                            
                                                                                                                                                                                     
  Order A — assign first, then snapshot:       
  ```dart                                                                                                                                      
  if (channelState.isUpToDate) {                                                                                                                                                   
    _currentLastMessage = message;   // cache = null                                                                                                                                
  }                                                  
  final latestLastMessage = [message, _currentLastMessage].latest;                                                                                                                   
  // = [null, null].latest = null  → empty text shown ✓         
  ```            
                                                                                                                                                                                     
  Order B — snapshot first, then assign:
  ```dart
  final latestLastMessage = [message, _currentLastMessage].latest;                                                                                                                   
  // = [null, msg1].latest = msg1  ← captured BEFORE the assignment         
  if (channelState.isUpToDate) {                                                                                                                                                     
    _currentLastMessage = message;   // cache = null, but latestLastMessage is already msg1                                                                                         
  }                                                                                                                                                                                
  // display msg1 → stale "hello" still shown ❌          
  ```                                                                                                                            
  
  But this whole discussion made me realise that we can make this whole thing a bit more readable:
  ```dart
  final Message? latestLastMessage;
  if (channelState.isUpToDate) {
      latestLastMessage = message;
      _currentLastMessage = latestLastMessage;
  } else {
      latestLastMessage = [message, _currentLastMessage].latest;
  }
  ```
  I will make this update!

xsahil03x (approved):

PR #2787, opened 2026-06-25
fix(ui): Fix `shadowed` messages not hidden in the channel list

VelikovPetar:
# Submit a pull request

Linear: FLU-562

## CLA

We were correctly hiding `shadowed` messages from other users in the Message List, but not on the Channel List items. This PR enhances the `ChannelLastMessageText` predicate with checks for `shadowed` messages from other users.

## Screenshots / Videos

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/26a013b0-e961-4b7d-bd34-f26c04b816f1"></video> | <video src="https://github.com/user-attachments/assets/9c176959-67c3-4137-ac77-da195a3a962a"></video> |

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed channel list last-message previews where shadowed messages could be hidden or displayed incorrectly.
  * Shadowed messages now render only when authored by the current user; error and ephemeral messages are still excluded.
* **New Features**
  * Added support for customizing which message is used as the channel list preview via an optional last-message predicate.
* **Tests**
  * Added widget tests covering current-user shadowing behavior, error/ephemeral filtering, and predicate override behavior.


packages/stream_chat_flutter/test/src/scroll_view/channel_scroll_view/stream_channel_list_item_test.dart
```diff
@@ -0,0 +1,185 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:mocktail/mocktail.dart';
+import 'package:stream_chat_flutter/stream_chat_flutter.dart';
+
+import '../../mocks.dart';
```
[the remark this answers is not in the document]
VelikovPetar: We already do this in tests - see `mark_read_test.dart`, `gradient_avatar_test.dart` etc.. as examples.

packages/stream_chat_flutter/lib/src/scroll_view/channel_scroll_view/stream_channel_list_item.dart
```diff
@@ -694,18 +689,20 @@ class ChannelLastMessageText extends StatefulWidget {
...
-  // considered for the last message.
-  static bool _defaultLastMessagePredicate(Message message) {
-    if (message.isShadowed) return false;
+  @override
+  State<ChannelLastMessageText> createState() => _ChannelLastMessageTextState();
+}
+
+// The default predicate to determine if the message should be
+// considered for the last message.
+bool Function(Message) _defaultLastMessagePredicate(String? currentUserId) {
```
renefloor: This feels a bit confusing, I think it's already clearer with a typedef. Something like this:
```dart
typedef LastMessagePredicate = bool Function(Message);
LastMessagePredicate _defaultLastMessagePredicateForUser(String? currentUserId) {
```

VelikovPetar: Addressed [here](https://github.com/GetStream/stream-chat-flutter/pull/2787/commits/b7617ff129a60b65c7bcfb1de72add9afb851268), let me know what do you think! (note, the `LastMessagePredicate` is now part of the Public API)

renefloor: As you say, it's part of the public API, so maybe better to put it in `typedefs.dart` so it's part of the list and not accidentally removed.

VelikovPetar: Ah right, my mistake! Will update it!

VelikovPetar: Moved it [here](https://github.com/GetStream/stream-chat-flutter/pull/2787/commits/58c075077459fb5c681bc319fb37259401c98c94). I was thinking about the naming, perhaps something like `ChannelListItemLastMessagePredicate` is better fitting, but in the end I decided to keep it simple, but let me know if you think we should update this!

renefloor: I think this is fine, because it just handles a message, this could be used elsewhere in theory.

renefloor (approved):

PR #2789, opened 2026-06-26
fix(core): Fix backwards pagination not working if channel was never opened

VelikovPetar:
# Submit a pull request
<!--Internal tickets have to be added by Stream devs-->
Fixes FLU-563
<!--Optional to add github issue which is solved by this PR-->
Github Issue: #

## CLA

When a user enters a channel with `unread_count > 0` but the server-side read state has `last_read_message_id = null `and `last_read = 0001-01-01T00:00:00Z` (Go's `time.Time{}` zero value — the "never explicitly read" sentinel), `StreamChannel._maybeInitChannel` was unconditionally calling `loadChannelAtTimestamp(currentUserRead.lastRead)`.
                                                                                                                                                                                   
The backend silently ignores zero-time `created_at_around` (via `IsZero()`) and returns the channel tail. The client then mis-infers boundaries via `_inferBoundariesFromAnchorTimestamp`: because year-1 is before every loaded message, it concludes `endOfPrependReached: true`, sets `_topPaginationEnded = true`, and backwards pagination is permanently dead on the channel.                                                                                                                                   
                                                                                                                                                                                   
The fix guards the timestamp branch on `lastRead.toUtc().isAfter(DateTime.utc(1970, 1, 1))`. When it's the zero sentinel, the code falls through to the existing catch-all if `(channel.state?.isUpToDate == false) loadChannelAtMessage(null)` — i.e. stay on cached latest if up-to-date, otherwise reload latest.                                                       
                                                                                                                                                                                   
Adds a `_maybeInitChannel` test group with three regression guards: `idAround` for `lastReadMessageId`, `createdAtAround` for real timestamps, and no query for the Go zero-time case.

## Screenshots / Videos

<!-- Consider to add screenshots and/or videos to show the effect of the change -->

| Before | After |
| --- | --- |
|<video src="https://github.com/user-attachments/assets/c4dae301-6136-427f-8ed7-f638b214f958"></video> |<video src="https://github.com/user-attachments/assets/31c0179a-5900-478f-9ab9-af775a7524e4"></video> |





<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed backwards pagination for channels that have never been opened.
  * Improved StreamChannel initialization so “never read” states skip timestamp-based loading and load the latest messages instead.
  * Addressed a StreamChannelListController issue where deleted-channel notifications weren’t handled correctly.
* **Tests**
  * Updated StreamChannel widget tests to use a consistent pumping helper and expanded pagination initialization coverage.
* **Documentation**
  * Updated the core changelog under **Upcoming → Fixed** with the latest fixes.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/packages/stream_chat_flutter_core/lib/src/stream_channel.dart
+++ b/packages/stream_chat_flutter_core/lib/src/stream_channel.dart
@@ -201,6 +201,10 @@ class StreamChannel extends StatefulWidget {
 
 // ignore: public_member_api_docs
 class StreamChannelState extends State<StreamChannel> {
+  // Anything before this is treated as the server's "never read" sentinel
+  // (Go's zero `time.Time{}` → `DateTime.utc(1, 1, 1)`).
+  static final _minValidLastRead = DateTime.utc(1970, 1, 1);
+
   /// Current channel
   Channel get channel => widget.channel;
 
@@ -878,13 +882,10 @@ class StreamChannelState extends State<StreamChannel> {
         }
       }
 
-      // `lastRead` is `DateTime.utc(1, 1, 1)` (Go's zero `time.Time{}`) when
-      // the server has never recorded an explicit read for this user. The
-      // server ignores that as a `created_at_around` anchor and returns the
-      // channel tail, after which `_inferBoundariesFromAnchorTimestamp`
-      // would mis-conclude `_topPaginationEnded = true`. Skip in that case
-      // and fall through to the catch-all "load latest" below.
-      if (currentUserRead.lastRead.toUtc().isAfter(DateTime.utc(1970, 1, 1))) {
+      // Skip the "never read" sentinel: the server ignores it as
+      // `created_at_around` and returns the tail, which would mis-infer
+      // `_topPaginationEnded = true`. Fall through to load-latest below.
+      if (currentUserRead.lastRead.isAfter(_minValidLastRead)) {
         return loadChannelAtTimestamp(currentUserRead.lastRead);
       }
     }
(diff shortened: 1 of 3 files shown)


packages/stream_chat_flutter_core/lib/src/stream_channel.dart
```diff
@@ -878,8 +878,15 @@ class StreamChannelState extends State<StreamChannel> {
...
 
-      // Otherwise, load the channel at the last read date.
-      return loadChannelAtTimestamp(currentUserRead.lastRead);
+      // `lastRead` is `DateTime.utc(1, 1, 1)` (Go's zero `time.Time{}`) when
+      // the server has never recorded an explicit read for this user. The
+      // server ignores that as a `created_at_around` anchor and returns the
+      // channel tail, after which `_inferBoundariesFromAnchorTimestamp`
+      // would mis-conclude `_topPaginationEnded = true`. Skip in that case
+      // and fall through to the catch-all "load latest" below.
+      if (currentUserRead.lastRead.toUtc().isAfter(DateTime.utc(1970, 1, 1))) {
```
xsahil03x: I think the comparison works without converting them to utc. Also maybe good to extract the default somewhere in static or const value in the class.

xsahil03x (approved):

packages/stream_chat_flutter_core/lib/src/stream_channel.dart
VelikovPetar: https://github.com/GetStream/stream-chat-flutter/pull/2789/commits/a57bb1fd93e109541e01405223637ebe36ed9428

PR #2794, opened 2026-07-01
fix(llc): Don't upsert out-of-window messages on `message.updated`/`message.deleted` events

VelikovPetar:
# Submit a pull request

Linear: FLU-560

Github Issue: #

## CLA

`message.updated` and soft `message.deleted` events were being unconditionally upserted into `ChannelState.messages` (and thread reply lists) via `updateMessage` / `deleteMessage`. When the target message wasn't in the currently loaded window — e.g. it lived on an older page the client hadn't paged in yet — this created a phantom entry in the sorted list, leaving a visible gap between the loaded slice and history.

### Fix

- `_updateMessages` (and its `_updateThreadMessages` / `_updateChannelMessages` / `_mergeMessages` helpers) now take an `upsert` flag.
- `message.updated` event handler passes `upsert: false`.
- `message.deleted` event handler passes `upsert: false` for **soft** deletes (hard deletes still route through `deleteMessage` unchanged).
- When `upsert: false` and an id isn't already in the loaded list, the message is skipped for the channel/thread lists. Side effects such as `pinnedMessages` maintenance and `activeLiveLocations` expiration still fire (they don't depend on the message being in the window).

The guard is intentionally id-based and independent of `isUpToDate` — even at the latest page we may have paginated past older history and receive an event for a message no longer in memory.

## Test plan

- New `channel_test.dart` groups under `messageUpdated` and `messageDeleted`:
  - `should NOT insert unknown message into messages list` (out-of-window edit)
  - `should update message in place when it IS in the loaded window`
  - `should still add to pinnedMessages when pinned:true even if not in loaded window`
  - `should NOT insert unknown reply into threads[parentId]`
  - `should still expire activeLiveLocations for out-of-window message`
  - Soft delete equivalents: phantom-record guard, in-window mark-as-deleted, unpin-via-\_pinIsValid, live-location clear, hard-delete no-op.

## Screenshots / Videos

No UI changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented `message.updated` and soft `message.deleted` from creating/insertions when the affected message is outside the currently loaded window, including correct handling for thread replies, pinned messages, and live-location expiry.
  * Ensured hard deletes remain a no-op for out-of-window messages.

* **New Features**
  * Added an `upsert` control to message reconciliation so unknown out-of-window messages can be ignored (`upsert` defaults to `true`).

* **Tests**
  * Expanded automated coverage for loaded-window behavior across update/delete scenarios.


packages/stream_chat/lib/src/client/channel.dart
```diff
@@ -3194,7 +3194,7 @@ class ChannelClientState {
         final message = event.message;
         if (message == null) return;
 
-        return updateMessage(message);
+        return _updateMessages([message], upsert: false);
```
xsahil03x: should we add the upsert flag to `updateMessage` too and use it?

packages/stream_chat/lib/src/client/channel.dart
```diff
@@ -3209,7 +3209,9 @@ class ChannelClientState {
           deletedForMe: event.deletedForMe,
         );
 
-        return deleteMessage(message, hardDelete: hardDelete);
+        if (hardDelete) return deleteMessage(message, hardDelete: true);
+        // Soft delete, update the message only if loaded (upsert: false)
+        return _updateMessages([message], upsert: false);
```
xsahil03x: Let's use `updateMessage` here if we decide to do https://github.com/GetStream/stream-chat-flutter/pull/2794/changes#r3526391971

packages/stream_chat/lib/src/client/channel.dart
VelikovPetar: Addressed here: https://github.com/GetStream/stream-chat-flutter/pull/2794/commits/5bff89d61b0a719644ad14a71bb704cf27c5adb0

packages/stream_chat/lib/src/client/channel.dart
VelikovPetar: Addressed here: https://github.com/GetStream/stream-chat-flutter/pull/2794/commits/5bff89d61b0a719644ad14a71bb704cf27c5adb0

packages/stream_chat/lib/src/client/channel.dart
```diff
@@ -3209,7 +3209,9 @@ class ChannelClientState {
           deletedForMe: event.deletedForMe,
         );
 
-        return deleteMessage(message, hardDelete: hardDelete);
+        if (hardDelete) return deleteMessage(message, hardDelete: true);
+        // Soft delete, update the message only if loaded (upsert: false)
+        return updateMessage(message, upsert: false);
```
xsahil03x: Should we do the check here or instead modify the `_deleteMessages` and pass `upsert: false` always?

```dart
  void _deleteMessages(
    Iterable<Message> messages, {
    bool hardDelete = false,
  }) {
    if (messages.isEmpty) return;

    if (hardDelete) return _removeMessages(messages);
    return _updateMessages(messages, upsert: false);
  }
```

packages/stream_chat/lib/src/client/channel.dart
```diff
@@ -3930,18 +3937,20 @@ class ChannelClientState {
...
+    bool upsert = true,
   }) {
     if (messages.isEmpty) return;
 
-    _updateThreadMessages(messages, update: update);
-    _updateChannelMessages(messages, update: update);
+    _updateThreadMessages(messages, update: update, upsert: upsert);
+    _updateChannelMessages(messages, update: update, upsert: upsert);
     _updatePinnedMessages(messages, update: update);
     _updateActiveLiveLocations(messages);
```
xsahil03x, line 3945: Lets also add the flag to these two

packages/stream_chat/lib/src/client/channel.dart
```diff
@@ -3958,13 +3967,23 @@ class ChannelClientState {
...
+        upsert: upsert,
       );
 
+      // Don't create a phantom entry for a thread that wasn't loaded: with
+      // `upsert: false` an out-of-window reply is dropped, leaving the merged
+      // list empty. Writing it back would make `threads.containsKey(parentId)`
+      // report a thread that was never paged in.
+      if (existingThreadMessages == null && updatedThreadMessages.isEmpty) {
+        continue;
+      }
```
xsahil03x: We should probably early return if we are not upserting and there's no thread to update.

```dart
for (final MapEntry(key: thread, :value) in messagesByThread.entries) {
  final threadMessages = updatedThreads[thread] ?? <Message>[];

  // If the thread is not loaded and upsert is false, skip updating it.
  if (threadMessages.isEmpty && !upsert) continue;

    existing: threadMessages,
    toMerge: value,
    update: update,
    upsert: upsert,
  );

  // Update the thread with the modified message list.
  updatedThreads[thread] = updatedThreadMessages.toList();
}
```

packages/stream_chat/lib/src/client/channel.dart
VelikovPetar: I believe that the `_updatePinnedMessages` method is fine if it always does an upsertion. For example, if a message gets pinned, we would get a `message.updated` event for that message with the `pinned` indicator flipped. In such case, we should anyway add the message to the pinned messages state. Otherwise, we need to explicitly check if the `message.updated` event was received for the purpose of pinning a message. 

Similarly for the `_updateActiveLiveLocation` -> I think that we are also fine if we always upsert them? In this case it would mean that a message outside of the loaded window was updated with a live location, but that is (I believe) already a valid scenario? We should get all live locations when loading the channel, unrelated to which page we load?

xsahil03x: Makes sense 👍🏼

packages/stream_chat/lib/src/client/channel.dart
VelikovPetar: I think this looks cleaner, I will update it like this! Thanks for the suggestion!

packages/stream_chat/lib/src/client/channel.dart
VelikovPetar: Done: https://github.com/GetStream/stream-chat-flutter/pull/2794/commits/36f12dd616cf0b6aa20bd5654d0489e4860844e6

packages/stream_chat/lib/src/client/channel.dart
VelikovPetar: Done: https://github.com/GetStream/stream-chat-flutter/pull/2794/commits/36f12dd616cf0b6aa20bd5654d0489e4860844e6

xsahil03x (approved):

PR #2804, opened 2026-07-10
feat(llc): Implement local unread counts

renefloor:
# Submit a pull request

Linear: FLU-587

## CLA

This adds an option for local only unread counts, especially interesting for things like livestreams with many updates.

## Summary by CodeRabbit

- **New Features**
  - Added optional client-side local unread-count tracking for channels without read events (off by default via `isLocalUnreadCountEnabled`).
  - When enabled, `unreadCount` updates on new messages, decrements on hard deletes only, and `markRead`/`markUnread`/`markUnreadByTimestamp` adjust local unread boundaries without network calls.
- **Documentation**
  - Updated the SDK changelog and unread-count rules to reflect local behavior and how it interacts with read receipts.
- **Tests**
  - Added coverage for local unread counting, boundary semantics, state merge preservation, and no-network assertions (including new client/mocks support).


packages/stream_chat/lib/src/client/channel.dart
```diff
@@ -3209,6 +3247,15 @@ class ChannelClientState {
           deletedForMe: event.deletedForMe,
         );
 
+        // Decrement the locally-tracked unread count for hard-deleted
+        // messages that would have counted as unread. Soft-deleted messages
+        // keep their slot. Only applies to channels that track unread counts
+        // locally (see [Channel.usesLocalUnreadCount]) — server-driven
+        // channels get corrected counts from server read events instead.
+        if (hardDelete && _channel.usesLocalUnreadCount && MessageRules.canCountAsUnread(message, _channel)) {
```
VelikovPetar, line 3294: Just curious: Don't soft-deletes also decrement the unread count?

packages/stream_chat/lib/src/client/channel.dart
```diff
@@ -1884,9 +1914,17 @@ class Channel {
...
   ///
   /// All messages after the provided timestamp will be marked as unread.
+  ///
+  /// If [usesLocalUnreadCount] is `true` for this channel, this updates the
+  /// unread count locally, on-device, without making a network request.
   Future<EmptyResponse> markUnreadByTimestamp(DateTime timestamp) async {
     _checkInitialized();
 
+    if (usesLocalUnreadCount) {
+      state!.markUnreadLocally(lastRead: timestamp);
```
VelikovPetar: In the `markUnread(String messageId)` method, we call `markUnreadLocally` with a `-1ms` of the message timestamp -> but here we call the `markUnreadLocally` with the exact timestamp. I am curious whether we might be missing some edge-case here, because this method will usually be called with the exact `createdAt` timestamp of a specific message. So if we call:
```dart
final message = targetMessage;

channel.markUnread(message.id);
channel.markUnreadByTimestamp(message.createdAt);
```
Would result in calling `markUnreadLocally` with a different timestamp. Probably not a big deal, but I wanted to flag it nonetheless!

packages/stream_chat/lib/src/client/channel.dart
```diff
@@ -3563,6 +3610,68 @@ class ChannelClientState {
...
+    if (currentUser == null) return;
+
+    final now = DateTime.now();
+    final lastReadMessageId = messageId ?? messages.lastOrNull?.id;
+
+    final existingUserRead = currentUserRead;
+    updateRead([
+      Read(
+        user: currentUser,
+        lastRead: now,
```
VelikovPetar, line 3690: Question: Does the `lastRead` mark the moment the channel was read, or the timestamp of the last read message? (I think we should align this with how the server populates this field)

packages/stream_chat/lib/src/client/channel.dart
```diff
@@ -3645,6 +3754,47 @@ class ChannelClientState {
...
+  /// `unreadMessages` while still adopting the remote delivery fields.
+  ///
+  /// No-op unless [Channel.usesLocalUnreadCount] is enabled and a local read
+  /// already exists for the current user.
+  ChannelState _preserveLocalUnreadState(ChannelState remoteState) {
+    if (!_channel.usesLocalUnreadCount) return remoteState;
+
+    final localRead = currentUserRead;
+    final remoteReads = remoteState.read;
+    if (localRead == null || remoteReads == null) return remoteState;
```
VelikovPetar, line 3852: Question: Will we always get an initial `Read` object in the response for a channel which has read/delivery receipts disabled? I am worried that if we do not get an initial object, we won't have a baseline to update on `message.new/message.deleted` events.

packages/stream_chat/lib/src/client/channel.dart
renefloor: No — intentionally, and this matches iOS exactly.

Swift short-circuits soft deletes with a dedicated skip reason before any of the other eligibility checks run:

// ChannelReadUpdaterMiddleware.swift:216-247
if let skipReason = !event.hardDelete
    ? .messageIsSoftDeleted
    : unreadCountUpdateSkippingReason(...) { return log.debug(...) }
channelRead.unreadMessageCount = max(0, channelRead.unreadMessageCount - 1)

The reasoning: a soft-deleted message keeps its slot in the message list (it renders as "This message was deleted"), so there is still an item the user hasn't seen — the count should keep pointing at it. A hard delete removes the row entirely, so the count has to shrink or it points at nothing. There's a test for both directions (does not decrement unreadCount when a message is soft-deleted / decrements unreadCount when a counted message is hard-deleted).

One divergence worth naming, though it's out of scope for this PR: Swift's decrement isn't gated on local unread counting at all — it runs for every channel, mirroring the backend. Flutter gates it on usesLocalUnreadCount, so a server-driven channel keeps a stale unread count after a hard delete until the next query/read event. That's a pre-existing gap in Flutter (nothing decremented before this PR either).

Shall I remove the `_channel.usesLocalUnreadCount` check?

packages/stream_chat/lib/src/client/channel.dart
renefloor: The moment of reading. DateTime.now() is correct and aligned with the server.

The server treats the two as separate fields, and we already follow that for remote events (channel.dart:3393-3421):

- message.read → lastRead: event.createdAt (when the read happened)
- notification.mark_read → lastRead: event.createdAt, plus last_read_message_id carried independently

Swift is the same: resetChannelRead(lastReadAt: event.createdAt), and its local variant uses markChannelAsRead(cid:userId:at: .init()) — i.e. Date(), now (ChannelRepository.swift:74). So lastRead = timestamp, lastReadMessageId = which message. ✅

However, this question does surface a real gap in markReadLocally. When called as markRead(messageId: X) with an older X, we set lastRead = now and reset the count to 0 (Read's unreadMessages defaults to 0). The server instead recomputes unread as the number of messages after X — so messages between X and now stay unread. Locally we clear the whole count and silently ignore the messageId for counting purposes.

Swift sidesteps this by not accepting a message id in markReadLocally at all. Two options:

1. When messageId is provided, use the anchor's createdAt as lastRead and recompute via the same messages.where(MessageRules.canCountAsUnread).length pass that markUnreadLocally already uses — symmetric with the unread path and matches the server.
2. Keep it simple and document that messageId only affects lastReadMessageId locally, never the count.

packages/stream_chat/lib/src/client/channel.dart
renefloor: No, we might not — but we don't need one. Three layers cover it, and iOS handles it the same defensive way.

1. MessageRules.canCountAsUnread returns true when there's no read object (message_rules.dart:120-121), so the first message still counts rather than being dropped.
2. The unreadCount setter lazily synthesizes the baseline when none exists (channel.dart:3597-3610), using lastRead: lastMessageAt ?? DateTime.now(). Because addNewMessage calls updateMessage (which advances lastMessageAt) before incrementing, the synthesized boundary lands on the message that was just counted — count is 1, no double-count, no loss.
3. _preserveLocalUnreadState no-ops when localRead == null, so the remote payload wins on first sync. And if the remote payload has no read array at all, updateChannelState merges reads by user id (channel.dart:3735-3738), so an existing local read survives regardless.

iOS is defensive in the same places: the middleware uses loadOrCreateChannelRead, markChannelAsRead creates a read for members when one is missing, saveChannelRead only preserves local values when !dto.isInserted (ChannelReadDTO.swift:105-114 — the direct analogue of _preserveLocalUnreadState, including the "first insert takes the payload" rule), and ChatChannelVC returns true // no read state, always mark as read.

One genuine (minor) edge: if a message arrives while isUpToDate == false, updateMessage is skipped so lastMessageAt isn't advanced, and the synthesized lastRead sits before the counted message. The count is still right (1); only the boundary timestamp is slightly conservative, which could let a later recompute re-count that message. If we want to be strict, messages.lastOrNull?.createdAt ?? lastMessageAt ?? DateTime.now() closes it.

packages/stream_chat/lib/src/client/channel.dart
renefloor: Good catch on there being something off here — the differing timestamps turn out to be correct, but chasing it down surfaced a real bug next to it. Fixed in 07a56734ddd3410d6fdc8f9308012e3c5b29d64c.
Why the timestamps should differ. markUnreadLocally(lastRead: x) has one contract: everything strictly after x is unread — that's forced by canCountAsUnread, which treats !message.createdAt.isAfter(lastRead) as read. The two public methods feed it inputs of different inclusivity:

| Entry point              | Input semantics                         | Conversion                                                            |
|--------------------------|-----------------------------------------|-----------------------------------------------------------------------| 
| markUnread(messageId)    | inclusive — "from this message onwards" | anchor is in the unread set → boundary must sit just before it → -1µs |
| markUnreadByTimestamp(t) | exclusive — "all messages after t"      | already a boundary → pass through                                     |

This mirrors the two REST endpoints ({message_id} vs {message_timestamp}), and iOS resolves the anchor with the same split: .messageId → MessageDTO.loadMessage(before:) (strictly before), .messageTimestamp → loadMessage(beforeOrEqual:) (at-or-equal) in MessageRepository.swift:351-368. If the -1µs moved into markUnreadLocally, markUnreadByTimestamp(t) would start marking the message at t as unread, contradicting its own doc and the endpoint.

So your example producing two different boundaries is correct — the calls are asking for different things. markUnread(m.id) ≡ markUnreadByTimestamp(m.createdAt - 1µs), not markUnreadByTimestamp(m.createdAt). I've documented that on both methods and added a test that pins the difference.

I also checked the epsilon is web-safe, since Duration(microseconds: 1) against a millisecond-backed DateTime would silently be a no-op: modern dart2js/dart2wasm keeps a separate _microsecond field and subtract, isAfter, and DateTime.parse all honour it (js_shared/lib/date_time_patch.dart:132-145, 205-213). With sdk: ^3.11.0 we're clear.

The actual bug you flushed out. Neither local path was passing lastReadMessageId, so it stayed null — while markReadLocally sets it and the server path sets it from notification.mark_unread. StreamChannel.getFirstUnreadMessage falls back to getOldestRegularMessage() when it's null (stream_channel.dart:802-806), which returns null unless _topPaginationEnded. Net effect on a livestream channel: after a local mark-unread the unread divider and jump-to-unread button anchored to the oldest loaded message, or disappeared entirely if history wasn't fully paginated — which is the normal case. Count was right, anchor was wrong. Both call sites now resolve and pass the anchor, with tests.

packages/stream_chat/lib/src/client/channel.dart
```diff
@@ -3563,6 +3612,68 @@ class ChannelClientState {
...
+    updateRead([
+      Read(
+        user: currentUser,
+        lastRead: now,
+        lastReadMessageId: lastReadMessageId,
+        lastDeliveredAt: existingUserRead?.lastDeliveredAt,
+        lastDeliveredMessageId: existingUserRead?.lastDeliveredMessageId,
+      ),
+    ]);
+  }
```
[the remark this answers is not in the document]
renefloor: Valid — taken. delivery_events is configured independently of read_events (channel_config.dart:35), so a channel tracking unread counts locally can still have delivery receipts on. In that case markReadLocally pushes lastRead past a pending candidate so canMarkAsDelivered returns false, but the candidate stayed in _deliveryCandidates — and _markCandidatesAsDelivered doesn't re-validate before POSTing (channel_delivery_reporter.dart:155), so we'd have reported delivery for a message we just marked read. Added reconcileDelivery([_channel]) after updateRead, matching the message.read listener at channel.dart:3427.

No equivalent needed in markUnreadLocally: moving the boundary backwards can only make more messages eligible, and reconcileDelivery only prunes.

packages/stream_chat/lib/src/client/channel.dart
VelikovPetar: Aham interesting, I wasn't aware of this. 

> Shall I remove the _channel.usesLocalUnreadCount check?

I would leave this up to you, maybe perhaps we can do it separately in a different PR, to have a cleaner changelog.

packages/stream_chat/lib/src/client/channel.dart
VelikovPetar: Hmm so if I understood correctly - basically we always set the unread count to 0 - no matter how many messages are actually after the marked message? 

Concerning the option one: Can we reliable calculate this? Because the user might be in a state where the latest messages are not loaded in memory (jump-to-message flow), so we might not have all the messages after the anchor loaded in memory - resulting in a wrong count. 
Not sure if this would be solvable in any way, so might have to fallback to option 2.

packages/stream_chat/lib/src/client/channel.dart
VelikovPetar: Got it - thanks for explaining! 

Concerning the mentioned edge case: This is also a valid concern for 'regular' `lastRead` updates right? We can fix it separately I would say if that's the case.

packages/stream_chat/lib/src/client/channel.dart
renefloor:
```
1. In the low-level client, Swift makes option 2 structural, not a documented caveat. markRead/markReadLocally take no message id at all — only markUnread takes a criteria (ReadStateHandler.swift:30-59, ChannelRepository.swift:42-78). The DB write is unconditional:

// ChannelReadDTO.swift:120-133
func markChannelAsRead(cid: ChannelId, userId: UserId, at: Date) {
    read.lastReadAt = at.bridgeDate
    read.unreadMessageCount = 0

So at is always .init() (now), count always goes to 0, and it doesn't even touch lastReadMessageId. There is no "read up to message X" local concept to get wrong — no recomputation, so Petar's "can we reliably calculate this?" never comes up.

2. The jump-to-message case is handled by the UI never calling markRead in that state. ChatChannelVC.shouldMarkChannelRead (ChatChannelVC.swift:83-107):

guard isLastMessageVisibleOrSeen else { return false }
guard isFirstPageLoaded else { return false }          // == hasLoadedAllNextMessages
guard !hasMarkedMessageAsUnread else { return false }
guard hasSeenFirstUnreadMessage || isLocalReadEnabled else { return false }
return read.lastReadAt < lastMessage.createdAt

isFirstPageLoaded is exactly the jump-to-message guard: while the user is anchored on an old message with the newest page unloaded, mark-read is skipped entirely. Every call site is gated on it (viewDidAppear:233, scrollViewDidScroll:470, didUpdateMessages:532).

That's also your "only update unread when all unread messages are read", expressed as three sticky flags rather than a count comparison: hasSeenFirstUnreadMessage (the user scrolled past the unread separator, willDisplayMessageAt:415-421) and hasSeenLastMessage (:467-469, reset on each markRead()), with isLastMessageVisibleOrSeen allowing "was seen, then scrolled away".

3. What this means for the PR. Flutter already has the important half: _maybeMarkMessagesAsRead gates on isUpToDate + hasUnread (message_list_view.dart:1214-1228), and the SDK UI calls debouncedMarkRead.call() with no id — the messageId parameter on line 910 is never exercised by our own UI. So option 2 costs nothing in practice: the only caller that could hit the stale-count path is app code calling channel.markRead(messageId:) directly.

My recommendation: take option 2 and document it, since it matches iOS exactly. Optionally tighten it further — have markReadLocally ignore messageId for lastRead too (always now, count always 0), which is literally what markChannelAsRead does. Two divergences from Swift remain but are separate from this thread: Flutter has no hasSeenFirstUnreadMessage equivalent (we mark read as soon as the last item is visible, Swift also wants the unread separator seen), and no isMarkedAsUnread guard to stop mark-read from immediately undoing a manual mark-unread.
```

I think it would be good if we add a similar `hasSeenFirstUnreadMessage`, but maybe in a different PR. Currently the whole thread is always marked as read when the bottom message is read, but there might be message higher up in the list that aren't read yet. I think that's a bigger issue. Created ticket FLU-640

> Because the user might be in a state where the latest messages are not loaded in memory (jump-to-message flow)

So in that case it's currently not marked as read, because `isUpToDate` is false.

Btw, the comment about the server is wrong, the server ignores the message timestamp entirely for the read-state write. Reflected in FLU-642

packages/stream_chat/lib/src/client/channel.dart
renefloor: created FLU-641

packages/stream_chat/lib/src/client/channel.dart
renefloor: Created FLU-642.

I mainly compared to iOS, but they also mark lastRead to Date.now. I think android has a better solution as it marks lastRead to the createdAt of the last message.

packages/stream_chat/lib/src/client/channel.dart
VelikovPetar: Ok, the proposal sounds good to me! (resolving this remark)

VelikovPetar (approved):

PR #2809, opened 2026-07-13
fix(llc): normalize strings when comparing ComparableField

### LINKED ISSUE #2601: ComparableField String sorting overrides correct backend order with case-sensitive codepoint comparison

stream_chat_flutter

Web

stream_chat: ^9.23.0
stream_chat_flutter: ^9.23.0
stream_chat_flutter_core: ^9.23.0

### What happened?

The backend API returns users in the correct sort order. However, StreamUserListController re-sorts results client-side using ComparableField.compareTo, which uses String.compareTo — a codepoint-based comparison. This overrides the correct server-side ordering with an incorrect one:                                                                                                                                                       
                                                                                                                                                                                                                           
  1. Case-sensitive — lowercase names sort after all uppercase names (e.g., jhon appears after Zara)                                                                                                                     
  2. Not locale-aware — characters like Ł, Ø, Æ sort after all ASCII letters (e.g., Łukasz appears after Zara instead of after Lukas)
                                                                                                                                                                                                                           
  The server presumably handles locale-aware collation correctly, but the client discards that ordering by re-sorting with a naive codepoint comparison.                                                                   
                                                                                                                                                                                                                           
  The issue is in stream_chat/lib/src/core/models/comparable_field.dart, line 27:                                                                                                                                          
  (final String a, final String b) => a.compareTo(b),                                                                                                                                                                    

  If client-side re-sorting is necessary (e.g., for merging paginated results), it should at minimum be case-insensitive, and ideally normalize diacritics to match the backend's collation. 

### Steps to reproduce

```
1. Create a StreamUserListController with sort: [SortOption.asc('name')]                                                                                                                                               
2. Have users with names starting with lowercase letters or non-ASCII characters (Polish Ł, Norwegian Ø, etc.)                                                                                                           
3. Observe that these users are pushed to the end of the list, despite the backend returning them in the correct alphabetical position
```

### Flutter doctor output

```shell
[✓] Flutter (Channel stable, 3.38.7, on Ubuntu 24.04.4 LTS 6.8.0-107-generic, locale en_US.UTF-8) [33ms]
    • Flutter version 3.38.7 on channel stable at /home/rafcio/fvm/versions/3.38.7
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 3b62efc2a3 (3 months ago), 2026-01-13 13:47:42 -0800
    • Engine revision 78fc3012e4
    • Dart version 3.10.7
    • DevTools version 2.51.1
    • Feature flags: enable-web, enable-linux-desktop, enable-macos-desktop, enable-windows-desktop, enable-android, enable-ios, cli-animations, enable-native-assets, omit-legacy-version-file, enable-lldb-debugging
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

xsahil03x:
## Summary

Fixes #2601 — client-side `SortOption` on `name` (and any other string field) was returning results in a different order than the server because `ComparableField` used a raw codepoint compare. Lowercase names and non-ASCII names (Polish `Ł`, Norwegian `Ø`, etc.) got pushed to the end instead of sorting alphabetically.

Root cause: `ComparableField.compareTo`'s `String` branch was `a.compareTo(b)`, i.e. byte-order on UTF-16 code units.

Fix: route string comparisons through a new `normalizeStringForSort` helper that mirrors the backend's [`NormalizeName`](https://github.com/GetStream/chat/blob/b92bf7991752a29e9f67b6ffe69b0bc529f5c48c/lib/combined/utils/text.go#L88-L95) pipeline:

1. Fold Latin diacritics and ligatures (`é → e`, `Ł → l`, `Æ → ae`) via `package:diacritic`.
2. Preserve Japanese kana, Thai, and Vietnamese-specific runes unchanged (matches backend's `Normalize` early-exit — critical for Japanese combining sound marks in the Hiragana range).
3. Lowercase.
4. Trim leading/trailing ASCII apostrophes (U+0027) then whitespace.

Backend has no tie-breaker for `sort: name`, so we don't either — equal-normalized values compare as `0` and Dart's stable `List.sort` preserves input order.

## Notes

- `SortOption`'s `comparator` parameter remains the escape hatch for callers who need raw codepoint or locale-aware behavior.
- Known deferred divergences (documented in the normalizer file): CJK/Thai word segmentation (backend uses Kagome + TWC dictionaries — no viable pure-Dart port), emoji-to-text token replacement, extended Unicode `Mn` categories.
- Adds `diacritic: ^0.1.6` to `stream_chat` (already used by `stream_chat_flutter`).

## Test plan

- [x] Added `packages/stream_chat/test/src/core/util/string_sort_normalizer_test.dart` — ports 10 applicable cases from backend's [`TestNormalizeName`](https://github.com/GetStream/chat/blob/b92bf7991752a29e9f67b6ffe69b0bc529f5c48c/lib/combined/utils/text_test.go#L300-L341) + coverage for Vietnamese preservation, ligature folding, combining marks, apostrophe/whitespace trimming, and Japanese combining sound marks (U+3099 / U+309A).
- [x] Added regression test in `comparable_field_test.dart` that sorts the exact scenario from the bug report (`Zara`, `jhon`, `Łukasz`, `Øystein`, `Adam`, `Marek`) and verifies the expected alphabetical order.
- [x] Full `stream_chat` suite (1551 tests) passes.
- [x] `dart analyze --fatal-infos .` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved client-side sorting of names and other text fields.
  * Sorting is now case-insensitive and handles accented characters and ligatures more consistently.
  * Lowercase and non-ASCII names are no longer incorrectly placed at the end of sorted lists.

* **Documentation**
  * Clarified default text-sorting behavior and how to provide custom comparison rules.

### DIFF
--- a/packages/stream_chat/lib/src/core/models/comparable_field.dart
+++ b/packages/stream_chat/lib/src/core/models/comparable_field.dart
@@ -1,3 +1,5 @@
+import 'package:stream_chat/src/core/util/string_sort_normalizer.dart';
+
 /// A wrapper class for values that implements [Comparable].
 ///
 /// This class is used to compare values of different types in a way that
@@ -8,6 +10,10 @@
 /// For example, when sorting a list of objects with different types of fields,
 /// using this class will ensure that all values are compared correctly
 /// regardless of their type.
+///
+/// String comparisons fold diacritics and ligatures to base characters
+/// (`Ł→l`, `Ø→o`, `Æ→ae`, `é→e`) and ignore case, so ordering aligns with
+/// what the server returns for name-like sorts.
 class ComparableField<T> implements Comparable<ComparableField<T>> {
   const ComparableField._(this.value);
 
@@ -24,12 +30,20 @@ class ComparableField<T> implements Comparable<ComparableField<T>> {
   int compareTo(ComparableField<T> other) {
     return switch ((value, other.value)) {
       (final num a, final num b) => a.compareTo(b),
-      (final String a, final String b) => a.compareTo(b),
+      (final String a, final String b) => _compareStrings(a, b),
       (final DateTime a, final DateTime b) => a.compareTo(b),
       (final bool a, final bool b) when a == b => 0,
       (final bool a, final bool b) => a && !b ? 1 : -1, // true > false
       _ => 0, // All comparisons were equal or incomparable types
     };
+  }
+
+  // Compares using the shared server-parity normalizer.
+  static int _compareStrings(String a, String b) {
+    final normalizedA = normalizeStringForSort(a);
+    final normalizedB = normalizeStringForSort(b);
+
+    return normalizedA.compareTo(normalizedB);
   }
 }
 
--- a/packages/stream_chat/lib/src/core/util/string_sort_normalizer.dart
+++ b/packages/stream_chat/lib/src/core/util/string_sort_normalizer.dart
@@ -1,0 +1,95 @@
+import 'package:diacritic/diacritic.dart';
+import 'package:meta/meta.dart';
+
+const _maxAscii = 0x7F;
+const _apostrophe = 0x0027;
+
+// Japanese: Hiragana + Katakana blocks.
+const _hiraganaStart = 0x3040;
+const _hiraganaEnd = 0x309F;
+const _katakanaStart = 0x30A0;
+const _katakanaEnd = 0x30FF;
+
+// Thai.
+const _thaiStart = 0x0E00;
+const _thaiEnd = 0x0E7F;
+
+// Vietnamese: tone-marked Latin range + unique letters (Đ/đ, Ơ/ơ, Ư/ư).
+const _vietnameseToneStart = 0x1EA0;
+const _vietnameseToneEnd = 0x1EF1;
+const _vietnameseDUpper = 0x0110;
+const _vietnameseDLower = 0x0111;
+const _vietnameseOHornUpper = 0x01A0;
+const _vietnameseOHornLower = 0x01A1;
+const _vietnameseUHornUpper = 0x01AF;
+const _vietnameseUHornLower = 0x01B0;
+
+/// Normalizes `value` into a client-side sort key that matches how the
+/// backend sorts by name.
+///
+/// Applied in order:
+///
+/// 1. Fold Latin diacritics and ligatures — `é → e`, `Ł → l`, `Æ → ae`.
+/// 2. Preserve Japanese, Thai, and Vietnamese-specific runes unchanged.
+/// 3. Lowercase.
+/// 4. Trim leading/trailing ASCII apostrophes (U+0027) then whitespace.
+///
+/// Mirrors backend [`NormalizeName`][ref].
+///
+/// [ref]: https://github.com/GetStream/chat/blob/b92bf7991752a29e9f67b6ffe69b0bc529f5c48c/lib/combined/utils/text.go#L88-L95
+@internal
+String normalizeStringForSort(String value) {
+  if (value.isEmpty) return value;
+  final folded = value.runes.map(_foldRune).join();
+  return _trimApostropheEdges(folded.toLowerCase()).trim();
+}
+
+// Returns the sort form of `rune`. ASCII and preserved-script runes pass
+// through unchanged; everything else routes through `removeDiacritics`.
+String _foldRune(int rune) {
+  if (rune <= _maxAscii) return String.fromCharCode(rune);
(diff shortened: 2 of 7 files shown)


xsahil03x:
The `build (android)` / `build (ios)` failures on this PR are **not caused by these changes** — they're a fresh upstream `firebase_messaging` publish issue.

**Root cause** — [firebase/flutterfire#18438](https://github.com/firebase/flutterfire/issues/18438) (opened 2026-07-13, currently open):

`firebase_messaging 16.4.2` and `firebase_crashlytics 5.2.5` were published referencing a `FirebasePlugin` base class that doesn't exist in any published `firebase_core_platform_interface`. Latest published is `firebase_core_platform_interface 7.1.0`, which only exports the old `FirebasePluginPlatform`. Any resolve that lands on `firebase_messaging 16.4.2` fails at build time with:

```
firebase_messaging-16.4.2/lib/src/messaging.dart:11:33: Error: Type 'FirebasePlugin' not found.
firebase_messaging-16.4.2/lib/src/messaging.dart:21:36: Error: The getter 'pluginConstants' isn't defined for the type 'FirebaseMessaging'.
```

**Same failure exists on plain `master`** — CI for master's HEAD commit ([`249a92c45`](https://github.com/GetStream/stream-chat-flutter/actions/runs/29254252737), `distribute_internal`) fails with identical errors. This PR's diff touches only `packages/stream_chat/**` (pure Dart LLC) — zero changes to `sample_app/`, `melos.yaml`, or any Firebase-related config.

**Plan** — wait for Firebase to either publish the missing `firebase_core_platform_interface` release that defines `FirebasePlugin` or ship a `firebase_messaging 16.4.3` that reverts the rename. The `stream_chat` unit tests, analyze, format, and codecov checks are all green.

packages/stream_chat/lib/src/core/models/comparable_field.dart
```diff
@@ -24,13 +30,21 @@ class ComparableField<T> implements Comparable<ComparableField<T>> {
   int compareTo(ComparableField<T> other) {
     return switch ((value, other.value)) {
       (final num a, final num b) => a.compareTo(b),
-      (final String a, final String b) => a.compareTo(b),
+      (final String a, final String b) => _compareStrings(a, b),
```
VelikovPetar, line 33: Just to double check if it is OK to apply the normalisation everywhere, not just for `User.name` sorting - does the BE apply normalisation everywhere as well?  (I see the backend has a `NormalizeText` utility in addition to the `NormalizeText` method)

packages/stream_chat/lib/src/core/util/string_sort_normalizer.dart
```diff
@@ -0,0 +1,95 @@
...
+const _vietnameseUHornLower = 0x01B0;
+
+/// Normalizes `value` into a client-side sort key that matches how the
+/// backend sorts by name.
+///
+/// Applied in order:
+///
+/// 1. Fold Latin diacritics and ligatures — `é → e`, `Ł → l`, `Æ → ae`.
+/// 2. Preserve Japanese, Thai, and Vietnamese-specific runes unchanged.
+/// 3. Lowercase.
```
VelikovPetar, line 34: Is the lowercasing intentional? (I assume the idea is that `Lukas` and `lukas` appear in the search next to each other, but still I want to double check)

xsahil03x: Yes, backend converts them to lowercase as well.

https://github.com/GetStream/chat/blob/b92bf7991752a29e9f67b6ffe69b0bc529f5c48c/lib/combined/utils/text.go#L92

packages/stream_chat/lib/src/core/models/comparable_field.dart
xsahil03x: Backend uses `NormalizeText` for search and not for any sort operations. Also the difference between `NormalizeText` and `NormalizeName` is very minute with just a additional [stemEnglishWords](https://github.com/GetStream/chat/blob/b92bf7991752a29e9f67b6ffe69b0bc529f5c48c/lib/combined/utils/text.go#L83). 

We can look into it later if we see any issues.

VelikovPetar: OK fine with me 👍

packages/stream_chat/lib/src/core/util/string_sort_normalizer.dart
VelikovPetar: Got it, thanks for confirming!

VelikovPetar (approved):

PR #2816, opened 2026-07-14
feat(ui): add configurable `StreamBackButton` unread count

VelikovPetar:
# Submit a pull request

Linear: FLU-557

Github Issue: #

## CLA

The `StreamChannelHeader` back-button unread badge counted the **currently open** channel in its total, so it never reflected "unread elsewhere".

- `StreamBackButton` now takes an `unreadCount` configuration via the new `StreamBackButtonUnreadCount`:
  - `.total({excludeCid})` — total unread across channels, optionally excluding one channel.
  - `.channel(cid)` — a specific channel's unread count.
- `StreamChannelHeader` now uses `.total(excludeCid: channel.cid)`, so its back-button badge shows the unread messages waiting in **other** channels.
- `StreamUnreadIndicator` gained an opt-in `excludeCid` (bare usage is unchanged).
- `showUnreadCount` and `channelId` on `StreamBackButton` are **deprecated** but still functional — fully backwards compatible.

**Test instructions:** open a channel that has unread messages while other channels also have unread. The header back-button badge shows `total − thisChannel`; going back to the channel list, the badge shows the full total. Covered by unit tests in `back_button_test.dart` and `unread_indicator_test.dart` (incl. the deprecated paths).

## Screenshots / Videos

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/5b3b12c7-6fb9-449c-907f-ef3dbb97a6a6"></video> | <video src="https://github.com/user-attachments/assets/ca1b9c37-f89f-4293-9668-a58a0f94e6f2"></video> |

## Summary by CodeRabbit

- **New Features**
  - Added `unreadIndicator` support for back buttons, enabling total unread badges (optionally excluding the current channel) or per-channel unread badges.
  - Updated channel and thread headers to use `unreadIndicator` for leading back button badges.
- **Bug Fixes**
  - Unread badges for channel navigation no longer include the currently open channel; excluded unread subtraction is clamped to zero.
- **Documentation**
  - Added migration guidance and changelog updates for `unreadIndicator`, including deprecation of legacy unread-count options.
- **Tests**
  - Expanded widget and golden coverage, including RTL and brightness variants, for `excludeCid` and precedence behavior.


packages/stream_chat_flutter/lib/src/misc/back_button.dart
```diff
@@ -47,13 +67,56 @@ class StreamBackButton extends StatelessWidget {
...
+
+/// Configures the unread badge on a [StreamBackButton].
+sealed class StreamBackButtonUnreadCount {
+  const StreamBackButtonUnreadCount();
+
+  /// Shows the total unread message count across all channels.
+  ///
+  /// Set [excludeCid] to omit a channel's unread messages from the total -
+  /// for example, the currently open channel.
+  const factory StreamBackButtonUnreadCount.total({String? excludeCid}) = _TotalUnreadCount;
```
VelikovPetar: Not completely sure about the naming here, let me know if you think we should use something different (`.unreadMessages()`perhaps, to serve as a forecast for the potentially upcoming `.unreadChannels()`)

packages/stream_chat_flutter/lib/src/misc/back_button.dart
```diff
@@ -9,19 +9,39 @@ class StreamBackButton extends StatelessWidget {
...
 
   /// Channel ID used to retrieve unread count
+  @Deprecated(
+    "Use 'unreadCount: StreamBackButtonUnreadCount.channel(cid)' instead. "
+    'This will be removed in a future version.',
+  )
   final String? channelId;
 
+  /// The unread count configuration for the back button.
+  final StreamBackButtonUnreadCount? unreadCount;
```
xsahil03x: Should we instead just ask for a `Widget/StreamUnreadIndicator` here directly? It will make it more composable and the api cleaner.

packages/stream_chat_flutter/lib/src/indicators/unread_indicator.dart
```diff
@@ -35,7 +40,8 @@ class StreamUnreadIndicator extends StatelessWidget {
     this.alignment,
     this.offset,
     this.semanticLabel,
-  }) : _unreadType = _UnreadChannels(cid: cid);
+  }) : _unreadType = _UnreadChannels(cid: cid),
+       excludeCid = null;
```
xsahil03x, line 44: I think we can support `excludeCid` here too. wdyt?

packages/stream_chat_flutter/lib/src/misc/back_button.dart
VelikovPetar: Good idea, perhaps passing a pure `Widget` would be the most future-proof. Not sure if we would be opening the API a bit too much, but since it is a pattern already used in the SDK, I think we should be OK. I will draft a solution using a `Widget` instead of `StreamBackButtonUnreadCount`.

packages/stream_chat_flutter/lib/src/indicators/unread_indicator.dart
VelikovPetar: In theory yes, should be feasible. I decided to not do it now because it wasn't part of the initial bug report. However I have one concern: In that case, we could in theory pass both `cid` and `excludeCid` to the `_UnreadChannels` config -> This can become a bit confusing, but if we prioritize the setup in the following way:
1. If `cid != null` -> show that specific channel unread count
2. If `excludeCid != null` -> all unread channels - 1 (if excluded channel has unread)
3. Else -> all unread channels

(Ideally we would have different factory methods for `channels` vs `currentChannel`, but currently the `channels` factory handles both cases, so I think introducing new API might be confusing).

What do you think about this?

xsahil03x: lets skip it until we have a usecase

packages/stream_chat_flutter/lib/src/misc/back_button.dart
VelikovPetar: There is a small caveat with this approach: If we rework it to pass a `Widget? unreadIndicator`, passing a `StreamUnreadIndicator` complicates the setup a bit because currently, its placement is done by nesting a child:
```dart
StreamUnreadIndicator(
      offset: .zero,
      excludeCid: excludeCid,
      child: button, // <--- StreamBackButton
)
```

This would mean that when building a `StreamBackButton` with a `StreamUnreadIndicator`, We will have to pass `child: null` in the indicator, and handle the positioning inside the `StreamBackButton`, something along the lines of:
```dart
if (_effectiveUnreadIndicator case final ind?) {                                                                         
    button = Stack(                                                                                                                          
      clipBehavior: Clip.none,                                                                                                                                      
      children: [                                                                                                                                                          
        button,                                                                                                                     
        Positioned.fill(                                                                                                      
          child: FittedBox(                                                                                                      
            fit: BoxFit.none,                                                                                                                          
            alignment: AlignmentDirectional.topEnd,                                                                                                                                 
            child: ind,              // <---- StreamUnreadIndicator WITHOUT child                                                                                                                                         
          ),                                                                                                                                                                      
        ),                                                                                                                                                                       
      ],                                                                                                                                                                         
    );                                                                                                                                                                       
  }
```
Should be doable, but in my opinion, it makes the API pretty confusing. I wanted to bring this up, before committing to this solution.

xsahil03x: Should be good, just make sure we use the same position as we had before. Flutter Badge also supports a nullable child. 

https://api.flutter.dev/flutter/material/Badge/child.html

VelikovPetar: Addressed: https://github.com/GetStream/stream-chat-flutter/pull/2816/commits/4f3aa78456061a992b29da5cea8610c2801995e2

docs/docs_screenshots/test/localization/localization_rtl_test.dart
```diff
@@ -191,7 +191,7 @@ void main() {
           child: Scaffold(
             appBar: const StreamChannelHeader(
               automaticallyImplyLeading: false,
-              leading: StreamBackButton(showUnreadCount: false),
+              leading: StreamBackButton(unreadIndicator: null),
```
xsahil03x: We can remove it as its null be default

packages/stream_chat_flutter/lib/src/misc/thread_header.dart
```diff
@@ -71,7 +71,10 @@ class StreamThreadHeader extends StatelessWidget implements PreferredSizeWidget
 
     var leading = this.leading;
     if (leading == null && automaticallyImplyLeading) {
-      leading = StreamBackButton(channelId: channel?.cid, showUnreadCount: true);
+      final cid = channel?.cid;
+      leading = StreamBackButton(
+        unreadIndicator: cid != null ? StreamUnreadIndicator.channels(cid: cid) : const StreamUnreadIndicator(),
```
xsahil03x: :nit: lets use a switch statement here, will make it a bit cleaner

xsahil03x (approved):

docs/docs_screenshots/test/localization/localization_rtl_test.dart
xsahil03x: Also, not sure if it does anything rn as we only check for non null case and otherwise fallback to the deprecated props.

```
    if (unreadIndicator case final effective?) return effective;
```

VelikovPetar: Fixing the first remark! 
I am not sure I fully understand what do you mean with the 2nd comment here though, can you give some more details?

xsahil03x: I meant, even if we pass null here in order to not show the unread count, it won't always behave correctly as we fallback to the deprecated `showUnreadCount` boolean in case its null currently. So currently passing null does not always mean we are hiding the unread count.

VelikovPetar: Ah I see what you mean. I think we have to do it like this, in order to be backwards compatible (`unreadIndicator: null` -> fallback to legacy `showUnreadCount`)

PR #2819, opened 2026-07-15
fix(ui): Mark channel read immediately when the unread indicator is dismissed

VelikovPetar:
# Submit a pull request

Linear: FLU-551

Github Issue: #

## CLA

**Problem:** When a channel is being spammed with messages while the user sits at the bottom, tapping the close (X) on `StreamMessageListView`'s jump-to-unread indicator appears to do nothing — the badge stays / immediately reappears and the dismiss is effectively ignored.

**Root cause:** The dismiss (`onDismissTap`) was routed through the same `debounce`d mark-read used by the auto-mark-read-at-the-bottom path (`leading: true`, 1s). During a burst, the auto-mark-read continuously holds the debounce window open, so its leading edge is already consumed. The dismiss tap then lands *inside* the window and is suppressed — it never fires a `markRead` call.

**Fix:** The dismiss now bypasses the debouncer and calls `channel.markRead()` / `channel.markThreadRead()` **immediately**, so it can never be swallowed regardless of what the auto-mark-read window is doing. The debounced path is unchanged and still used by the auto-mark-read-at-the-bottom flow. Error handling and null-safety match the debounced path exactly.

**Tests:** Added regression tests in `mark_read_test.dart` under `unread indicator dismiss`:
- Dismiss marks the channel read immediately (single `markRead` call).
- Repeated dismiss taps each fire `markRead` (proves the action is no longer debounced — verified this test fails if wired back to the debounced path).

## Screenshots / Videos

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/7ff462f2-41ae-4fe2-aec6-9f37a9016d56"></video> | <video src="https://github.com/user-attachments/assets/745c4e28-8c39-4cb4-8d93-52bfea51c6f8"></video> |

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved unread indicator dismissal in `StreamMessageListView` so it consistently marks messages as read, even when channels receive rapid bursts of new messages.
  * Ensured repeated dismiss taps trigger read marking immediately (not debounced/ignored), reflecting one mark action per tap when configured.
  * Kept debounced read marking for automatic bottom-of-list and visibility-based behavior.
* **Tests**
  * Added coverage for unread indicator dismissal behavior, including repeated-tap call counts.

### DIFF
--- a/packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
+++ b/packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
@@ -920,6 +920,16 @@ class _StreamMessageListViewState extends State<StreamMessageListView> {
 
   Future<void> _markMessagesAsRead() async {
     if (widget.parentMessage case final parent?) {
+      // If we are in a thread, mark the thread as read immediately.
+      await streamChannel?.channel.markThreadRead(parent.id);
+    } else {
+      // Otherwise, mark the channel as read immediately.
+      await streamChannel?.channel.markRead();
+    }
+  }
+
+  Future<void> _markMessagesAsReadDebounced() async {
+    if (widget.parentMessage case final parent?) {
       // If we are in a thread, mark the thread as read.
       debouncedMarkThreadRead.call([parent.id]);
     } else {
@@ -1230,7 +1240,7 @@ class _StreamMessageListViewState extends State<StreamMessageListView> {
     if (!isInThread && !hasUnread) return;
 
     // Mark messages as read if it's allowed.
-    return _markMessagesAsRead();
+    return _markMessagesAsReadDebounced();
   }
 
   void _getOnThreadTap() {
--- a/packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
+++ b/packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
@@ -922,13 +922,14 @@ class _StreamMessageListViewState extends State<StreamMessageListView> {
     if (widget.parentMessage case final parent?) {
       // If we are in a thread, mark the thread as read immediately.
       await streamChannel?.channel.markThreadRead(parent.id);
-    } else {
-      // Otherwise, mark the channel as read immediately.
-      await streamChannel?.channel.markRead();
+      return;
     }
+
+    // Otherwise, mark the channel as read immediately.
+    await streamChannel?.channel.markRead();
   }
 
-  Future<void> _markMessagesAsReadDebounced() async {
+  Future<void> _debouncedMarkMessagesAsRead() async {
     if (widget.parentMessage case final parent?) {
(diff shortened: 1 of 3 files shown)


packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
```diff
@@ -919,6 +919,16 @@ class _StreamMessageListViewState extends State<StreamMessageListView> {
...
+    if (widget.parentMessage case final parent?) {
+      // If we are in a thread, mark the thread as read immediately.
+      await streamChannel?.channel.markThreadRead(parent.id);
+    } else {
+      // Otherwise, mark the channel as read immediately.
+      await streamChannel?.channel.markRead();
+    }
+  }
+
+  Future<void> _markMessagesAsReadDebounced() async {
```
xsahil03x: let name it, `debouncedMarkMessagesAsRead`
```suggestion
  Future<void> _debouncedMarkMessagesAsRead() async {
```

packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
```diff
@@ -919,6 +919,16 @@ class _StreamMessageListViewState extends State<StreamMessageListView> {
   );
 
   Future<void> _markMessagesAsRead() async {
+    if (widget.parentMessage case final parent?) {
+      // If we are in a thread, mark the thread as read immediately.
+      await streamChannel?.channel.markThreadRead(parent.id);
+    } else {
+      // Otherwise, mark the channel as read immediately.
+      await streamChannel?.channel.markRead();
+    }
```
xsahil03x, line 926: lets early return here

```suggestion
    if (widget.parentMessage case final parent?) {
      // If we are in a thread, mark the thread as read immediately.
      return streamChannel?.channel.markThreadRead(parent.id);
    }
    
      // Otherwise, mark the channel as read immediately.
      return streamChannel?.channel.markRead();
```

packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
VelikovPetar: Addressed: https://github.com/GetStream/stream-chat-flutter/pull/2819/changes/2c6531d0a521e440ac4c5c382bdfcfd59da1b9e6

packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
VelikovPetar: Addressed: https://github.com/GetStream/stream-chat-flutter/pull/2819/changes/2c6531d0a521e440ac4c5c382bdfcfd59da1b9e6

xsahil03x (approved):

PR #2821, opened 2026-07-16
ci(repo): mock firebase for e2e tests

testableapple:
Resolve https://linear.app/stream/issue/FLU-602

## What

Keeps the native Firebase SDK (Core / Crashlytics / Messaging) out of e2e test builds entirely — nothing is downloaded, compiled, linked, or initialized during e2e runs — and cleans up the iOS build caching in the e2e workflows based on measured data.

## How the Firebase mock works

- `sample_app/e2e_stubs/` contains three **pure-Dart no-op packages** named `firebase_core`, `firebase_crashlytics`, and `firebase_messaging`, stubbing exactly the API surface `sample_app` uses (`Firebase.initializeApp`, `FirebaseOptions`, `recordError`/`recordFlutterFatalError`, the messaging streams/token getters, etc.). Permission requests report denied, tokens resolve to `null`, message streams never emit.
- Because the stubs have no `flutter.plugin` section, the Flutter tool registers **no plugins** for them → no Firebase pods on iOS, no Firebase Gradle deps on Android (although we do not use it on Android on CI right now, but it supports it if we decide to do it on Android as well).
- `sample_app/pubspec_overrides.e2e.yaml` wires them in via `dependency_overrides`. The e2e workflows copy it over `pubspec_overrides.yaml` **before** `melos bootstrap` — melos merges its own managed overrides into the existing file and preserves the stub entries.
- No app code or `project.pbxproj` changes are needed: the existing `isE2eTestRun` guards cover runtime behavior, and `pod install` regenerates the CocoaPods build phases on the runner to match the resolved plugin set.

## Also in this PR

- **CocoaPods cache step moved after Bootstrap** so its key hashes the resolved `pubspec.lock` (the lock is gitignored and doesn't exist before `melos bootstrap` — previously that part of the key was always empty and never rotated on dependency changes).
- **Removed the Xcode compilation cache (`cache-xcode-cas`)**. A verbose diagnostic run proved it works end-to-end on CI (1,734 `replay cache hit` remarks, 0 misses) but delivers only ~10% build-time benefit after normalizing for runner speed (the net is roughly zero on a good day and negative on rotation days) — cache replay is IO-bound and the non-cacheable tail (linking, script phases, kernel compile) dominates on the standard 4-vCPU runners — while costing ~12s restore per run, ~25s save on key rotation, and ~1 GB of cache quota. Recoverable from git history if we move to larger runners.

## Validation

- CI run with stubs active: pod install contains **zero** Firebase/Google/nanopb/Promises targets; tests green.
- Locally: `.flutter-plugins-dependencies` has no firebase entries with stubs active, `flutter analyze` clean, and melos bootstrap preserves the stub overrides.
- Local development is unaffected — the override file is only applied in CI.

## Test (local)

Screenshot 2026-07-17 at 10 32 38

## Summary

* **Bug Fixes**
  * Improved iOS end-to-end test reliability by using local no-op Firebase stubs instead of real Firebase services.
  * Updated iOS CI caching so CocoaPods cache keys are derived after dependencies are bootstrapped.

* **Tests**
  * Added pure-Dart stubs for Firebase Core, Crashlytics, and Messaging to support e2e execution without Firebase connectivity.

* **Chores**
  * Simplified end-to-end workflows by removing the Xcode compilation caching step and standardizing workflow naming.


.github/workflows/e2e_test.yml
renefloor, line 1: Should we already do this for android as well? Currently only applied for iOS.

renefloor (approved):

.github/workflows/e2e_test.yml
testableapple: Android is much faster and the benefit of mocking it there is very low. But this way we at least can receive crash reports from Android as far as i understood.

PR #2837, opened 2026-07-22
fix(llc): update watcher count and watchers from realtime events

VelikovPetar:
# Submit a pull request

Linear: FLU-619

Github Issue: #

## CLA

`ChannelClientState.watcherCount` and `ChannelClientState.watchers` never reflected realtime events during a session:

- The server's `watcher_count` on `user.watching.start` / `user.watching.stop` / `message.new` was dropped into `Event.extraData`, so the count stayed frozen at the last full `channel.watch()` / `queryChannels` snapshot.
- Watchers were never removed from `watchers` on `user.watching.stop` — the additive keyed merge in `updateChannelState` (needed for watcher pagination) can only upsert, so the departing watcher lingered until the next full query.

iOS and Android already trust the server-provided `watcher_count` per event, so Flutter was the outlier.

### Changes
- Add `Event.watcherCount`, mapping the server `watcher_count` field (promoted to a first-class field instead of `extraData`).
- Apply the count on `user.watching.start`, `user.watching.stop`, and `message.new`.
- Remove the departing watcher from `ChannelClientState.watchers` on `user.watching.stop` (assign state directly, bypassing the additive merge).

`notification.message_new` / `notification.thread_message_new` are intentionally excluded: the backend fetches the watcher count (via `GetChannelWatchersCount`) only *after* the notification fan-out in `PostMessageInsert`, so these events structurally carry `watcher_count: 0`.

### Tests
- New `Watching Events` test group in `channel_test.dart`: start/stop update the count, stop removes the watcher, a null count preserves the existing value, `message.new` updates the count, and `notification.message_new` does not overwrite it.
- `melos run format` + `melos run analyze` clean; watching-event tests pass.

## Screenshots / Videos

No UI changes.

## Summary by CodeRabbit

- **New Features**
  - Added watcher counts to channel events, including watching and new-message events.
  - Watcher counts are now available through event data and remain synchronized with channel activity.

- **Bug Fixes**
  - Fixed stale watcher counts during active sessions.
  - Fixed watcher lists so users are removed when they stop watching.
  - Prevented unreliable notifications from overwriting accurate watcher counts.


packages/stream_chat/lib/src/client/channel.dart
```diff
@@ -2671,11 +2672,10 @@ class ChannelClientState {
       _channel.on(EventType.userWatchingStop).listen((event) {
         final watcher = event.user;
         if (watcher != null) {
-          final existingWatchers = channelState.watchers;
-          updateChannelState(
-            channelState.copyWith(
-              watchers: [...?existingWatchers?.where((user) => user.id != watcher.id)],
-            ),
+          final existingWatchers = channelState.watchers ?? const <User>[];
+          _channelState = channelState.copyWith(
```
xsahil03x, line 2676: Any particular reason to not use `updateChannelState` here?

VelikovPetar: The problem is that `updateChannelState` does a "merge" operation over the `watchers`:
```dart
final watchers = _channelState.watchers ?? const <User>[];
final newWatchers = watchers.merge(
      updatedState.watchers,
      key: (watcher) => watcher.id,
);
```
So basically, if the initial watcher state is: `["a", "b", "c"]`
And we receive a `user.stop.watching` for `c`, we would merge `["a", "b"]` into `["a", "b", "c"]` which wouldn't actually update the state. 

We could introduce some new flags to the `updateChannelState` method to handle such scenarios, but I think that it would make things more complicated.

Let me know what do you think!

xsahil03x (approved):

PR #2842, opened 2026-07-23
fix(ui): filter StreamTypingIndicator initial data by parentId

VelikovPetar:
# Submit a pull request

Linear: FLU-614

Github Issue: #

## CLA

`StreamTypingIndicator` seeded its `BetterStreamBuilder.initialData` from the **unfiltered** `typingEvents` map, while its stream filters typing events by `parentId`. As a result the very first frame could show typing users from the wrong context — main-channel typers in a thread header (`ThreadHeader`), or thread typers in a channel list tile — until the first stream emission corrected it (visible as a brief flash, stretched by the 300ms `AnimatedSwitcher` fade).

This PR extracts a shared `_typingUsers` helper that applies the `parentId` predicate and uses it for **both** `initialData` and the stream `map`, so the two can no longer diverge.

**Breaking-change verdict:** behavioral fix only — no signature changes, not source-breaking, no opt-out needed (nothing can reasonably depend on the transient wrong-user flash).

**Testing:** extended the existing `'it only shows typing users matching its parentId'` widget test to mount the indicator with a pre-existing main-channel typing event and assert the first frame shows nothing. Verified the new assertion fails without the fix. Also manually verified in the sample app (thread header no longer flashes main-channel typers).

## Screenshots / Videos

No UI changes (the fix removes a transient first-frame flash of wrong typing users; steady-state UI is unchanged).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the typing indicator so it only displays users typing in the current channel or thread context.
  * Prevented typing users from briefly showing from the wrong context on the indicator’s first frame by applying consistent filtering to both initial and live updates.

### DIFF
--- a/packages/stream_chat_flutter/lib/src/indicators/typing_indicator.dart
+++ b/packages/stream_chat_flutter/lib/src/indicators/typing_indicator.dart
@@ -34,20 +34,20 @@ class StreamTypingIndicator extends StatelessWidget {
   final String? parentId;
 
   // Compares the typing users by id so the indicator only rebuilds when the
-  // set of typing users changes. The stream maps to a new lazy Iterable on
-  // every typing event (and on the stale-typing cleanup), so identity equality
-  // would rebuild even when the same users are still typing. Order-sensitive
-  // because only the first user is shown by name.
-  bool _typingUsersEquals(Iterable<User>? previous, Iterable<User>? current) {
-    final previousIds = previous?.map((it) => it.id).toList() ?? const <String>[];
-    final currentIds = current?.map((it) => it.id).toList() ?? const <String>[];
-    return const ListEquality<String>().equals(previousIds, currentIds);
+  // set of typing users changes. The stream maps to a new list on every typing
+  // event (and on the stale-typing cleanup), so identity equality would rebuild
+  // even when the same users are still typing. Order-sensitive because only the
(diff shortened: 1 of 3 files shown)


packages/stream_chat_flutter/lib/src/indicators/typing_indicator.dart
```diff
@@ -44,17 +44,21 @@ class StreamTypingIndicator extends StatelessWidget {
     return const ListEquality<String>().equals(previousIds, currentIds);
   }
 
+  // Keeps only the users typing in the same context as this indicator: the
+  // thread identified by [parentId], or the main channel when it's null.
+  Iterable<User> _typingUsers(Map<User, Event> typingEvents) {
```
xsahil03x: Lets just switch to returning a List instead of Iterable, it will make it a bit performant as we are also converting them to list everytime we compare the data.

xsahil03x (approved):

packages/stream_chat_flutter/lib/src/indicators/typing_indicator.dart
VelikovPetar: Addressed in: 1fe313a892d94fbce7d0c6876a816b17acf751e5

xsahil03x (approved):

PR #2843, opened 2026-07-23
fix(ui): enrich link previews for uppercase URL schemes

xsahil03x:
## Summary

Fixes [FLU-625](https://linear.app/stream/issue/FLU-625). URLs with an uppercase scheme (e.g. `HTTPS://example.com`) were not enriched into link previews, while lowercase `https://` worked.

The composer's URL regex already matched uppercase schemes, but the raw matched text was forwarded to `client.enrichUrl(...)` verbatim — so the `/og` endpoint received `HTTPS://…` and failed to enrich. The fix normalizes the scheme (via `Uri.parse`, which lowercases it) before enriching, so `HTTPS://` behaves the same as `https://`.

Flutter counterpart to the React ([#3226](https://github.com/GetStream/stream-chat-react/pull/3226)) and Angular ([#734](https://github.com/GetStream/stream-chat-angular/pull/734)) fixes.

## Scope

This fixes the composer link-preview **enrichment** path. Display-side linkification of uppercase schemes in rendered message text lives in `stream_core_flutter` and is out of scope here.

## Testing

Added `message_composer_url_enrichment_test.dart` — asserts `enrichUrl` is called with the normalized `https://example.com` for both uppercase and lowercase input. Fails before the fix, passes after.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed link-preview (OG) enrichment when URL schemes are uppercase (e.g., `HTTPS://`) by normalizing the scheme before enrichment.
  - Improves preview selection by choosing the first eligible URL found in the message.
  - Clears any existing link-preview attachment when no qualifying preview candidate is available or link sending isn’t allowed.

- **Tests**
  - Expanded widget tests with table-driven cases covering multiple scheme/host casing combinations and URLs embedded in surrounding text, ensuring the normalized URL is used for enrichment.

### DIFF
--- a/packages/stream_chat_flutter/lib/src/message_input/stream_message_composer.dart
+++ b/packages/stream_chat_flutter/lib/src/message_input/stream_message_composer.dart
@@ -1293,19 +1293,22 @@ class DefaultStreamMessageComposerState extends State<DefaultStreamMessageCompos
     if (_lastSearchedContainsUrlText == value) return;
     _lastSearchedContainsUrlText = value;
 
-    final matchedUrls = _urlRegex.allMatches(value).where((it) {
-      final _parsedMatch = Uri.tryParse(it.group(0) ?? '')?.withScheme;
-      if (_parsedMatch == null) return false;
+    // Find the first url to preview, normalizing the scheme so links like
+    // `HTTPS://` enrich the same as `https://`.
+    String? firstMatchedUrl;
+    for (final match in _urlRegex.allMatches(value)) {
+      final url = Uri.tryParse(match.group(0) ?? '')?.withScheme;
+      if (url == null) continue;
+      if (!widget.props.ogPreviewFilter.call(url, value)) continue;
 
-      return widget.props.ogPreviewFilter.call(_parsedMatch, value);
-    }).toList();
-
-    // Reset the og attachment if the text doesn't contain any url
-    if (matchedUrls.isEmpty || !channel.canSendLinks) {
-      return _effectiveController.clearOGAttachment();
+      firstMatchedUrl = url.toString();
+      break;
     }
 
-    final firstMatchedUrl = matchedUrls.first.group(0)!;
+    // Reset the og attachment if the text doesn't contain any url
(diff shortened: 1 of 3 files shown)


renefloor (approved):

packages/stream_chat_flutter/lib/src/message_input/stream_message_composer.dart
```diff
@@ -1293,20 +1293,23 @@ class DefaultStreamMessageComposerState extends State<DefaultStreamMessageCompos
...
 
-    final matchedUrls = _urlRegex.allMatches(value).where((it) {
-      final _parsedMatch = Uri.tryParse(it.group(0) ?? '')?.withScheme;
-      if (_parsedMatch == null) return false;
-
-      return widget.props.ogPreviewFilter.call(_parsedMatch, value);
-    }).toList();
+    // Find the first url to preview, normalizing the scheme so links like
+    // `HTTPS://` enrich the same as `https://`.
+    String? firstMatchedUrl;
```
VelikovPetar, line 1298: Question: Wouldn't it be possible to end up without an enriched URL, it the `/og` endpoint doesn't return in time before sending the message, or if we send the message before this method fires (`350ms` debounce). 
I am curious if maybe the bug report was for that reason 🤔 Because I can see some messages getting enriched even with `HTTPS://` prefix

xsahil03x: Yes, it's a possibility. But when i tried reproducing it in the sample app i wasn't able to get any og for a link with uppercase `HTTPS://`.  One way to solve it is to use `url_enrichment` flag on channel but it will still fail as the backend extracts the url from text and the text will still be with Uppercase letters. So this also needs to be fixed in the backend code in order to work correctly.

VelikovPetar: I see, I was afraid that we would still see reports of this issue after the fix. But nevertheless, I see no reason to block this PR. But we should probably bring this up with the BE as well.

PR #2845, opened 2026-07-23
feat(ui): make StreamMessageListView auto-scroll configurable

xsahil03x:
## Summary

Adds `StreamMessageListViewConfiguration.autoScrollPolicy` so integrators can control whether and how `StreamMessageListView` scrolls to the newest message when a new message arrives.

Resolves [FLU-626](https://linear.app/stream/issue/FLU-626/make-auto-scroll-configurable). A customer (Kalshi) is doing custom message pinning, which conflicts with the list's built-in auto-scroll on the sender's own messages. Previously this behavior was hardcoded with no way to opt out.

## API

`StreamAutoScrollPolicy` is a strategy object (the `ScrollPhysics` idiom) with const presets plus a custom resolver:

| Policy | Behavior |
| --- | --- |
| `StreamAutoScrollPolicy.whenOwnMessageOrAtBottom` | **Default.** Scrolls on the current user's own messages, and on others' only while the newest message is visible at the bottom. Byte-identical to the previous hardcoded behavior. |
| `StreamAutoScrollPolicy.whenAtBottom` | Scrolls only while the newest message is visible at the bottom, regardless of sender. |
| `StreamAutoScrollPolicy.disabled` | Never auto-scrolls; the caller drives scrolling. |
| `StreamAutoScrollPolicy.custom(resolver)` | Fully custom — the resolver receives `StreamAutoScrollDetails { message, isOwnMessage, isAtBottom }` and returns a `StreamAutoScrollBehavior { none, jump, animate }`. |

```dart
StreamMessageListView(
  config: StreamMessageListViewConfiguration(
    autoScrollPolicy: StreamAutoScrollPolicy.disabled,
  ),
  scrollController: myItemScrollController, // drive scrolling yourself
)
```

## Notes

- Non-breaking: the default reproduces today's behavior exactly, so existing integrations are unaffected.

## Testing

- Unit tests for every preset's `resolve()` across the `isOwnMessage × isAtBottom` matrix, plus `custom` delegation.
- Widget tests: `disabled` suppresses own-message auto-scroll while scrolled up; `custom` returning `jump` jumps to bottom on an incoming message.
- The 6 pre-existing auto-scroll regression tests still pass, confirming the default is unchanged.
- `dart analyze --fatal-infos` clean; formatted at 120 width.

## Summary by CodeRabbit

* **New Features**
  * Added `autoScrollPolicy` to `StreamMessageListViewConfiguration` to control scrolling when new messages arrive.
  * Built-in options now support **animate** (own messages/when at bottom), **animate only when at bottom**, **fully disabled**, and **custom** behavior to choose **animate**, **jump**, or **no scroll**.
* **Bug Fixes**
  * Fixed `StreamTypingIndicator` briefly showing typing from the wrong context on the first frame.
* **Tests**
  * Added coverage for built-in and custom auto-scroll policy behavior, including disabled and jump modes.
* **Documentation**
  * Updated the package changelog to describe the new `autoScrollPolicy` behavior.


packages/stream_chat_flutter/lib/src/message_list_view/auto_scroll_policy.dart
```diff
@@ -0,0 +1,131 @@
...
+  AutoScrollBehavior resolve(AutoScrollDetails details) => AutoScrollBehavior.none;
+}
+
+class _CustomAutoScrollPolicy extends AutoScrollPolicy {
+  const _CustomAutoScrollPolicy(this._resolve);
+
+  final AutoScrollResolver _resolve;
+
+  @override
+  AutoScrollBehavior resolve(AutoScrollDetails details) => _resolve(details);
```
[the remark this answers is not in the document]
xsahil03x: Not applicable. The presets are `static const` singletons (`StreamAutoScrollPolicy.disabled`, `.whenAtBottom`, `.whenOwnMessageOrAtBottom`), so each is a single canonical instance and identity equality already makes equal presets compare equal inside `StreamMessageListViewConfiguration`. Value equality is intentionally omitted on the strategy object to match Flutter's `ScrollPhysics` convention; `custom` compares by identity, which is the expected behavior for a closure-backed policy. (This comment predates the switch from const factory constructors to static-const fields.)

packages/stream_chat_flutter/lib/src/message_list_view/auto_scroll_policy.dart
```diff
@@ -0,0 +1,136 @@
...
+
+/// The information available when resolving an [StreamAutoScrollBehavior] for a
+/// newly arrived message.
+///
+/// Passed to an [StreamAutoScrollResolver]. Not intended to be constructed directly.
+class StreamAutoScrollDetails {
+  /// Creates a set of details describing a newly arrived message.
+  const StreamAutoScrollDetails({
+    required this.message,
+    required this.isOwnMessage,
```
VelikovPetar: Do we have any benefit of passing `isOwnMessage` here? Its value could be deducted from the `message.user`.

packages/stream_chat_flutter/lib/src/message_list_view/auto_scroll_policy.dart
```diff
@@ -0,0 +1,136 @@
...
+
+  /// The message that just arrived.
+  final Message message;
+
+  /// Whether [message] was sent by the current user.
+  final bool isOwnMessage;
+
+  /// Whether the newest message is currently visible at the bottom of the
+  /// list.
+  final bool isAtBottom;
```
VelikovPetar, line 38: Should we maybe re-phrase the docs here? Something along the lines of: `Whether the message list is currently scrolled to the bottom`?

VelikovPetar (approved):

PR #2846, opened 2026-07-27
fix(llc, core, ui, localization): hide raw StreamChannel errors behind themed, connection-aware states

xsahil03x:
## FLU-628

Raw `StreamChannel` initialization errors — and the default list/scroll-view error states — were leaking unthemed, unlocalized error details to end users. This routes them through themed, localized, **connection-aware** default states (no internet / slow connection / generic), overridable at every level.

### What changed

**`stream_chat`**
- Add `StreamChatNetworkError.type` (`StreamChatNetworkErrorType`) exposing the transport failure kind; deprecate `isRequestCancelledError` in favor of `type == StreamChatNetworkErrorType.cancel`.
- A *failed* channel initialization no longer leaves `Channel.initialized` errored or its `name`/`image`/`extraData` setters throwing — a subsequent successful (re)init recovers cleanly (supports the retry flow).
- Clearer `toString()` for `StreamWebSocketError` / `StreamChatNetworkError`.

- Add `StreamChannelState.retry()` to re-run a failed channel init (the retry action for `StreamChannel.errorBuilder`).
- Add `DefaultStreamChannelBuilders`, an inherited widget supplying default loading/error builders to descendant `StreamChannel`s.
- The default `StreamChannel` error state no longer exposes raw error details.

**`stream_chat_flutter`**
- `StreamChat` now installs themed, connection-aware default loading & error states for `StreamChannel` — overridable per-`StreamChannel` or via `DefaultStreamChannelBuilders`.
- Default list/scroll-view error states (channel, message, member, user, thread, poll-vote, reaction, search, photo) are now connection-aware, falling back to each view's specific error text.
- `StreamScrollViewErrorWidget` gains `errorSubtitle`; `StreamScrollViewLoadingWidget` gains `size` (`StreamLoadingSpinnerSize`) and deprecates the now-unused `height`/`width`.

**`stream_chat_localizations`**
- New connection-error strings (`connectionError*`, `slowConnectionError*`, `genericError*`) for all supported locales.

### Testing
- All four touched package suites pass; `dart analyze --fatal-infos` clean.
- New tests cover the default channel error/loading states, the retry-clears-stale-error flow, and full localization coverage.

## Summary by CodeRabbit

* **New Features**
  * Added “Try Again” retry support for failed channel initialization.
  * Introduced default themed channel loading/error builders (connection-aware).
  * Added network error text resolution with localized connection/slow-connection/generic copy.
  * Added loading spinner sizing and optional error subtitles.
  * Added transport error classification to improve error display.
* **Bug Fixes**
  * Improved channel initialization/retry behavior and watcher state consistency.
  * Error UI now stays safe (no raw error leakage) while showing clearer, connection-specific guidance.
* **Deprecations**
  * Deprecated older cancellation flag, some loading widget sizing parameters, and a reaction picker callback.

### DIFF
--- a/packages/stream_chat/lib/src/client/channel.dart
+++ b/packages/stream_chat/lib/src/client/channel.dart
@@ -111,7 +111,7 @@ class Channel {
   ///
   /// {@macro name}
   set name(String? name) {
-    if (_initializedCompleter.isCompleted) {
+    if (_isInitialized) {
       throw StateError(
         'Once the channel is initialized you should use `channel.updateName` '
         'to update the channel name',
@@ -124,7 +124,7 @@ class Channel {
   ///
   /// {@macro image}
   set image(String? image) {
-    if (_initializedCompleter.isCompleted) {
+    if (_isInitialized) {
       throw StateError(
         'Once the channel is initialized you should use `channel.updateImage` '
         'to update the channel image',
@@ -134,7 +134,7 @@ class Channel {
   }
 
   set extraData(Map<String, Object?> extraData) {
-    if (_initializedCompleter.isCompleted) {
+    if (_isInitialized) {
       throw StateError(
         'Once the channel is initialized you should use `channel.update` '
         'to update channel data',
@@ -569,13 +569,16 @@ class Channel {
   StreamChatClient get client => _client;
   final StreamChatClient _client;
 
-  final Completer<bool> _initializedCompleter = Completer();
+  Completer<bool> _initializedCompleter = Completer();
 
   /// True if this is initialized.
   ///
   /// Call [watch] to initialize the client or instantiate it using
   /// [Channel.fromState].
   Future<bool> get initialized => _initializedCompleter.future;
+
+  // Whether the channel is successfully initialized and not disposed.
+  bool get _isInitialized => _initializedCompleter.isCompleted && state != null;
 
   final _cancelableAttachmentUploadRequest = <String, CancelToken>{};
   final _messageAttachmentsUploadCompleter = <String, Completer<Message>>{};
@@ -713,7 +716,7 @@ class Channel {
               }
             })
             .catchError((e, stk) {
-              if (e is StreamChatNetworkError && e.isRequestCancelledError) {
+              if (e is StreamChatNetworkError && e.type == .cancel) {
                 client.logger.info('Attachment ${it.id} upload cancelled');
 
                 // remove attachment from message if cancelled.
@@ -2042,6 +2045,14 @@ class Channel {
     PaginationParams? watchersPagination,
     bool preferOffline = false,
   }) async {
+    // A prior failed init left the completer errored; reset it so this attempt
+    // owns the `initialized` result. Must stay before the first `await` so a
+    // caller reading `initialized` right after `query()`/`watch()` begins sees
+    // the fresh completer.
+    if (_initializedCompleter.isCompleted && this.state == null) {
+      _initializedCompleter = Completer<bool>();
+    }
+
     ChannelState? channelState;
 
     try {
@@ -2376,7 +2387,7 @@ class Channel {
   }
 
   void _checkInitialized() {
-    if (_initializedCompleter.isCompleted && state != null) return;
+    if (_isInitialized) return;
 
     throw StateError(
       "Channel $cid hasn't been initialized yet or has been disposed. "
--- a/packages/stream_chat/lib/src/core/error/stream_chat_error.dart
+++ b/packages/stream_chat/lib/src/core/error/stream_chat_error.dart
@@ -165,7 +165,7 @@ class StreamChatNetworkError extends StreamChatError {
   bool get isRetriable => data == null;
 
   @override
-  List<Object?> get props => [...super.props, code, statusCode];
+  List<Object?> get props => [...super.props, code, statusCode, type];
 
   @override
   String toString({bool printStackTrace = false}) {
--- a/packages/stream_chat/lib/src/core/error/stream_chat_error.dart
+++ b/packages/stream_chat/lib/src/core/error/stream_chat_error.dart
@@ -4,12 +4,17 @@ import 'package:equatable/equatable.dart';
 import 'package:stream_chat/stream_chat.dart';
 import 'package:web_socket_channel/web_socket_channel.dart';
 
+/// Base class for all errors surfaced by the Stream Chat SDK.
 ///
+/// See also:
+///
+///  * [StreamChatNetworkError], raised by failed HTTP requests.
+///  * [StreamWebSocketError], raised on the realtime connection.
 class StreamChatError extends Equatable implements Exception {
-  ///
+  /// Creates a new [StreamChatError] with the given [message].
   const StreamChatError(this.message);
 
-  /// Error message
+  /// A human-readable description of what went wrong.
   final String message;
 
   @override
@@ -19,22 +24,22 @@ class StreamChatError extends Equatable implements Exception {
   String toString() => 'StreamChatError(message: $message)';
 }
 
-///
+/// An error received over the realtime (WebSocket) connection.
 class StreamWebSocketError extends StreamChatError {
-  ///
+  /// Creates a new [StreamWebSocketError] with the given [message].
   const StreamWebSocketError(
     super.message, {
     this.data,
   });
 
-  ///
+  /// Creates a [StreamWebSocketError] from a Stream error payload.
   factory StreamWebSocketError.fromStreamError(Map<String, Object?> error) {
     final data = ErrorResponse.fromJson(error);
     final message = data.message ?? '';
     return StreamWebSocketError(message, data: data);
   }
 
-  ///
+  /// Creates a [StreamWebSocketError] from a [WebSocketChannelException].
   factory StreamWebSocketError.fromWebSocketChannelError(
     WebSocketChannelException error,
   ) {
@@ -42,20 +47,20 @@ class StreamWebSocketError extends StreamChatError {
     return StreamWebSocketError(message);
   }
 
-  ///
+  /// The structured error returned by the server, if any.
+  final ErrorResponse? data;
+
+  /// The Stream error code, if one was provided.
   int? get code => data?.code;
 
-  ///
+  /// The [ChatErrorCode] for this error, or null if unrecognised.
   ChatErrorCode? get errorCode {
     final code = this.code;
     if (code == null) return null;
     return chatErrorCodeFromCode(code);
   }
 
-  /// Response body. please refer to [ErrorResponse].
-  final ErrorResponse? data;
-
-  ///
+  /// Whether the operation can be retried.
   bool get isRetriable => data == null;
 
   @override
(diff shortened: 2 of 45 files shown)


packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
```diff
@@ -535,9 +535,12 @@ class _StreamMessageListViewState extends State<StreamMessageListView> {
 
     Widget defaultErrorBuilder(BuildContext context, Object error) {
       if (widget.builders.error case final builder?) return builder(context, error);
+      final text = resolveNetworkErrorText(context, error);
       return Center(
         child: StreamScrollViewErrorWidget(
-          errorTitle: Text(context.translations.loadingMessagesError),
+          errorTitle: Text(text.title),
+          errorSubtitle: Text(text.description),
+          retryButtonText: Text(context.translations.tryAgainLabel),
```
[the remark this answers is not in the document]
xsahil03x: Good catch — fixed in `ffdad0cc4`. Both `StreamMessageListView` and `StreamChannelListView` now pass `fallbackTitle` (`loadingMessagesError` / `loadingChannelsError`), so unclassified/server errors keep their view-specific title while connection and timeout errors still resolve to the connection-aware copy.

packages/stream_chat_flutter/lib/src/scroll_view/member_scroll_view/stream_member_grid_view.dart
```diff
@@ -365,14 +365,20 @@ class StreamMemberGridView extends StatelessWidget {
...
+          error,
+          fallbackTitle: context.translations.loadingUsersError,
+        );
+        return Center(
+          child: StreamScrollViewErrorWidget(
+            errorTitle: Text(text.title),
+            onRetryPressed: controller.refresh,
           ),
+        );
+      },
```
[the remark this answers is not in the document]
xsahil03x: Intentional — these compact/inline list error states stay title-only. `errorSubtitle` is new in this PR; before it `StreamScrollViewErrorWidget` had no subtitle at all and every list rendered a single-line error, so keeping the inline lists single-line is faithful to the existing design. The connection-aware **title** (e.g. "No Internet Connection") already conveys the state here; only the larger full-screen states (`StreamChannel`, `StreamMessageListView`, `StreamChannelListView`) deliberately show the extra description line. Happy to revisit if we decide inline lists should grow to two lines.

packages/stream_chat_flutter/lib/src/scroll_view/member_scroll_view/stream_member_list_view.dart
```diff
@@ -332,13 +332,19 @@ class StreamMemberListView extends StatelessWidget {
...
+        error,
+        fallbackTitle: context.translations.loadingUsersError,
+      );
+      return Center(
+        child: StreamScrollViewErrorWidget(
+          errorTitle: Text(text.title),
+          onRetryPressed: controller.refresh,
         ),
+      );
+    },
```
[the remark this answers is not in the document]
xsahil03x: Same as the `stream_member_grid_view` thread: inline list error states are intentionally single-line (title only). `errorSubtitle` is new in this PR, so title-only keeps these lists consistent with their pre-PR single-line design; the connection-aware title carries the key info. Only the full-screen error states show the description line.

packages/stream_chat_flutter/lib/src/scroll_view/reaction_scroll_view/stream_reaction_list_view.dart
```diff
@@ -198,13 +198,19 @@ class StreamReactionListView extends StatelessWidget {
...
+      if (errorBuilder?.call(context, error) case final builder?) return builder;
+      final text = resolveNetworkErrorText(
+        context,
+        error,
+        fallbackTitle: context.translations.loadingReactionsError,
+      );
+      return Center(
+        child: StreamScrollViewErrorWidget(
+          errorTitle: Text(text.title),
+          onRetryPressed: controller.refresh,
```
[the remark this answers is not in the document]
xsahil03x: Intentional, same reasoning as the `stream_member_grid_view` thread — inline list error states stay single-line (title only), which is faithful to their pre-PR design since `errorSubtitle` is new here. Reserving the two-line title+description layout for the full-screen states.

packages/stream_chat_flutter/lib/src/scroll_view/stream_scroll_view_loading_widget.dart
```diff
@@ -1,24 +1,27 @@
...
 class StreamScrollViewLoadingWidget extends StatelessWidget {
   /// Creates a new instance of [StreamScrollViewLoadingWidget].
   const StreamScrollViewLoadingWidget({
     super.key,
-    this.height = 42,
-    this.width = 42,
+    this.size = StreamLoadingSpinnerSize.lg,
+    @Deprecated('No longer used; the spinner is sized via size') this.height = 42,
+    @Deprecated('No longer used; the spinner is sized via size') this.width = 42,
   });
```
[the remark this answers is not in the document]
xsahil03x: Deliberate — the deprecation reads "No longer used". `StreamScrollViewLoadingWidget` now renders the design-system `StreamLoadingSpinner`, which is sized via discrete `StreamLoadingSpinnerSize` tokens rather than arbitrary pixels, so there is no meaningful mapping from a px `height`/`width` to a token to preserve until removal. `size` (default `lg`) supersedes them; callers needing a specific size should move to `size`.

packages/stream_chat_localizations/lib/src/stream_chat_localizations_ca.dart
```diff
@@ -185,6 +185,24 @@ class StreamChatLocalizationsCa extends GlobalStreamChatLocalizations {
...
+  String get slowConnectionErrorTitle => 'Connexió a Internet lenta';
+
+  @override
+  String get slowConnectionErrorDescription => 'Sembla que hi ha un problema amb la teva connexió a Internet';
+
+  @override
+  String get genericErrorTitle => 'Error';
+
+  @override
+  String get genericErrorDescription => 'Vaja, alguna cosa ha anat malament';
```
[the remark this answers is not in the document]
xsahil03x: These are `@override` implementations — the doc comments live on the `Translations` interface in `stream_chat_flutter` (which documents `connectionErrorTitle` and friends). Every `@override` getter in these locale files follows the same undocumented-override convention (~294 overrides, only 2 `///` comments in the file), `public_member_api_docs` does not flag overrides, and `dart analyze --fatal-infos` is clean. Documenting only these 6 would be inconsistent with the rest of the file.

packages/stream_chat/lib/src/client/channel.dart
```diff
@@ -111,7 +111,7 @@ class Channel {
   ///
   /// {@macro name}
   set name(String? name) {
-    if (_initializedCompleter.isCompleted) {
+    if (_isInitialized) {
```
VelikovPetar, line 114: As far as I can tell, this check now includes the `state != null` check as well:
```dart
bool get _isInitialized => _initializedCompleter.isCompleted && state != null;
```

But previously, it only checked if the following condition is satisfied:
```dart
if (_initializedCompleter.isCompleted)
```

Just flagging it, in case this wasn't intentional.

packages/stream_chat/lib/src/core/error/stream_chat_error.dart
```diff
@@ -162,3 +181,47 @@ class StreamChatNetworkError extends StreamChatError {
...
+  return switch (type) {
+    DioExceptionType.connectionError => StreamChatNetworkErrorType.connectionError,
+    DioExceptionType.connectionTimeout => StreamChatNetworkErrorType.connectionTimeout,
+    DioExceptionType.sendTimeout => StreamChatNetworkErrorType.sendTimeout,
+    DioExceptionType.receiveTimeout => StreamChatNetworkErrorType.receiveTimeout,
+    DioExceptionType.badResponse => StreamChatNetworkErrorType.badResponse,
+    DioExceptionType.cancel => StreamChatNetworkErrorType.cancel,
+    DioExceptionType.badCertificate => StreamChatNetworkErrorType.badCertificate,
+    // Unknown or future dio types map to unknown.
+    _ => StreamChatNetworkErrorType.unknown,
```
VelikovPetar, line 229: Just for my understanding: How do we decide which `Dio` types we forward, and which do we collapse to `.unknown`?

packages/stream_chat/test/src/core/error/stream_chat_error_test.dart
```diff
@@ -215,5 +225,56 @@ void main() {
...
+        });
+      }
+    });
+
+    test('.type defaults to unknown for non-network errors', () {
+      final error = StreamChatNetworkError(ChatErrorCode.internalSystemError);
+      expect(error.type, StreamChatNetworkErrorType.unknown);
+    });
+
+    test('.type is not part of equality', () {
```
VelikovPetar: Just curious: Why don't we include the `.type` as part of the equality check?

packages/stream_chat_flutter/lib/src/scroll_view/stream_scroll_view_loading_widget.dart
```diff
@@ -1,24 +1,37 @@
...
+import 'package:stream_core_flutter/chat.dart';
 
 /// A widget that is displayed while the [StreamScrollView] is loading.
 class StreamScrollViewLoadingWidget extends StatelessWidget {
   /// Creates a new instance of [StreamScrollViewLoadingWidget].
   const StreamScrollViewLoadingWidget({
     super.key,
-    this.height = 42,
-    this.width = 42,
+    this.size = StreamLoadingSpinnerSize.lg,
```
VelikovPetar, line 9: Hmm should we make this addition backwards compatible? In theory, if a customer was setting custom `height/width`, they would now lose that customisation.

packages/stream_chat_flutter/lib/src/utils/network_error_text.dart
```diff
@@ -0,0 +1,32 @@
...
+/// "slow connection" copy. Everything else falls back to [fallbackTitle] /
+/// [fallbackDescription] when provided, otherwise to the generic localized copy.
+({String title, String description}) resolveNetworkErrorText(
+  BuildContext context,
+  Object? error, {
+  String? fallbackTitle,
+  String? fallbackDescription,
+}) {
+  final translations = context.translations;
+  return switch (error) {
```
VelikovPetar, line 18: I like this improvement, but I am a bit worried that it can be considered a breaking change. For example, a customer overriding the `translations.genericErrorTitle` might expect to always see that error message - with the new changes, they would see their localised error only when the error is not `.connectionError/.connectionTimeout/.sendTimeout/.receiveTimeout`. 

The change is most likely OK, but still I wanted to raise this concern.

packages/stream_chat_flutter/lib/stream_chat_flutter.dart
```diff
@@ -157,5 +157,6 @@ export 'src/utils/device_segmentation.dart';
 export 'src/utils/extensions.dart';
 export 'src/utils/helpers.dart';
 export 'src/utils/message_preview_formatter.dart';
+export 'src/utils/network_error_text.dart';
```
VelikovPetar: Do we want this publicly available?

packages/stream_chat/test/src/core/error/stream_chat_error_test.dart
xsahil03x: Good catch, might be a mistake from claude

packages/stream_chat/lib/src/core/error/stream_chat_error.dart
xsahil03x: We currently map all the current dio types, if something new comes up and someone reports then we can add a map for it otherwise this will start breaking.

packages/stream_chat/lib/src/client/channel.dart
xsahil03x: Intentional — it's the fix. After a *failed* init the completer is completed (with an error) but `state` is still null, so the old `isCompleted`-only guard wrongly threw on the `name`/`image`/`extraData` setters. `_isInitialized` (`isCompleted && state != null`) keeps those setters usable after a failed init so a retry can proceed; covered by the `setters remain usable after a failed watch()` test.

packages/stream_chat_flutter/lib/src/scroll_view/stream_scroll_view_loading_widget.dart
xsahil03x: Good call — kept it backwards compatible. `height`/`width` are still honored in `build` (now deprecated in favor of `size`), so a customer that set them doesn't lose their sizing until the eventual removal.

packages/stream_chat_flutter/lib/src/utils/network_error_text.dart
xsahil03x: Fair concern. It only affects the *default* error UI — any custom `errorBuilder` is untouched — and the connection/slow copy is itself localizable via the new `connectionError*` / `slowConnectionError*` keys. So no customization is actually lost: `genericError*` still covers everything unclassified, and the connection/timeout cases just gained their own overridable keys.

packages/stream_chat_flutter/lib/stream_chat_flutter.dart
xsahil03x: Good catch — made it internal (removed the barrel export), since it's an implementation detail of the default builders.

packages/stream_chat/lib/src/core/error/stream_chat_error.dart
VelikovPetar: What threw me off here, was the test where we also instantiate the `DieExceptionType.transformTimeout` - which is currently not mapped:

Screenshot 2026-07-28 at 13 41 25

xsahil03x: Yeah, this is due to this. let me bump dio
https://github.com/GetStream/stream-chat-flutter/pull/2846#discussion_r3665173299

VelikovPetar (approved):

PR #2852, opened 2026-07-29
feat(ui): add onReactionTap reporting the tapped reaction

xsahil03x:
## Summary

A message's displayed reactions only fired `onReactionsTap`, which hands back **just the `Message`** — integrators had no way to tell *which* reaction chip was tapped (e.g. to toggle that specific reaction). This adds `onReactionTap`, which reports the tapped reaction.

Requested on Slack (tap a specific reaction to increment/decrement it). Cross-SDK, React Native / UIKit already expose a per-reaction tap; this brings Flutter in line.

## What changed

- **New `onReactionTap` on `StreamMessageItem` and `StreamMessageListView`:**
  ```dart
  typedef OnReactionTap = void Function(BuildContext context, ReactionTapDetails details);

  class ReactionTapDetails {
    final Message message;
    final Reaction? reaction; // null for a clustered/overflow chip
  }
  ```
  - `reaction` is the tapped reaction — the user's **full own reaction** when they've reacted with that type (populated `user`/`score`/timestamps), otherwise a `type` + `emojiCode` template (mirrors the reaction picker's resolution)
  - `reaction` is `null` for a clustered or overflow chip, which maps to no single reaction
- **Payload is a `ReactionTapDetails` object** (not positional params) so it can grow non-breakingly, following Flutter's `<Event>Details` convention (`TapDownDetails`, `DragUpdateDetails`, …). The callback also receives the tapped message's `BuildContext`, so integrators can navigate or show overlays relative to it without threading a context in themselves.
- **Deprecated** `onReactionsTap` (and the `OnReactionsTap` typedef) in favor of `onReactionTap`, following the `onReactionPicked` → `onReactionSelected` deprecation pattern. Exclusivity is asserted on both the `StreamMessageItem` constructor and `StreamMessageItemProps`. Default "who reacted" behavior is unchanged.
- Internal plumbing (`StreamMessageContent`, `StreamMessageReactions` — not exported) migrated cleanly; deprecation is only on the public, exported entry points.

## Usage

```dart
StreamMessageListView(
  onReactionTap: (context, details) {
    final reaction = details.reaction;
    if (reaction == null) return; // clustered/overflow chip
    final channel = StreamChannel.of(context).channel;
    final own = details.message.ownReactions?.any((r) => r.type == reaction.type) ?? false;
    own
        ? channel.deleteReaction(details.message, reaction)
        : channel.sendReaction(details.message, reaction);
  },
);
```

## Dependency

Depends on **GetStream/stream-core-flutter#140** (`StreamReactions.onReactionPressed`). Pinned via a git ref in `melos.yaml` / `pubspec.yaml`; **the ref must be updated to the merged commit once #140 lands** (and swapped back to a pub version before release).

## Tests

`test/src/message_widget/stream_message_reactions_test.dart`:
- segmented: tapping a chip reports its reaction
- segmented: reports the full own reaction when the user has reacted with that type
- segmented: tapping the "+N" overflow chip reports `null`
- clustered: tapping the grouped chip reports `null`


packages/stream_chat_flutter/lib/src/utils/typedefs.dart
```diff
@@ -187,8 +187,17 @@ typedef OnMessageLongPress = void Function(Message);
...
+@Deprecated('Use OnReactionTap instead. OnReactionTap also reports the tapped reaction.')
 typedef OnReactionsTap = void Function(Message);
 
+/// {@template onReactionTap}
+/// The action to perform when a message's reaction is tapped.
+///
+/// Reports the tapped [Reaction], or `null` when the tap does not map to a
+/// single reaction (for example a clustered or overflow chip).
+/// {@endtemplate}
+typedef OnReactionTap = void Function(Reaction? reaction, Message message);
```
renefloor: I think we should by default now make a new object so we can always add more data whenever we want. Also context might be useful


```dart
typedef OnReactionTap = void Function(BuildContext context, ReactionTapData data);

class ReactionTapData {
    Message message;
    Reaction? reaction;
}
```

renefloor (approved):

PR #2855, opened 2026-07-30
fix(llc): Ignore events dispatched after the client is disposed

VelikovPetar:
# Submit a pull request

Linear: FLU-658

Github Issue: #

## CLA

`StreamChatClient.handleEvent` could throw `StateError: Cannot add new events after calling close` when the client is disposed while a **reconnect recovery** was still in flight.

### Root cause

`_onConnectionStatusChanged` is `async`. On an offline→online transition it awaits a recovery pipeline (`sync` / `queryChannelsOnline`) and only then emits the final `connectionRecovered` event. The driving subscription is fire-and-forget (`wsConnectionStatusStream.pairwise().listen(...)`). If `dispose()` closes `_eventController` while that continuation is suspended on an `await`, the trailing `handleEvent(connectionRecovered)` adds to a **closed** controller — surfacing as an unhandled async error (zone / `PlatformDispatcher.onError` / Crashlytics).

This is pre-existing on `master`; it was surfaced by the offline-reactions E2E test (`user adds a reaction while offline`) on #2847, where the Android emulator's slower timing loses the race deterministically.

### Fix

Guard `handleEvent` against a closed controller:

```dart
void handleEvent(Event event) {
  // Ignore events that arrive after the client has been disposed.
  if (_eventController.isClosed) return;
  ...
}
```

`_eventController.isClosed` is terminal and set only in `dispose()`; `disconnectUser()` (logout→login reuse) leaves it open, so live/reused clients are unaffected. Post-`close()` the broadcast stream is already complete, so no consumer ever received these events — the guard removes only the internal throw, not any delivery.

### Test instructions

A new regression test in `client_test.dart` (`dispose during reconnect recovery`) disposes the client while a reconnect recovery is suspended on a held-open `queryChannels`, then resumes it — reproducing the exact CI stack. Red without the guard, green with it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented a `StateError` when disposing the client while reconnect recovery is still in progress.
  * Ensured reconnect recovery can complete without emitting late events or errors after disposal.
* **Tests**
  * Added a new test covering disposal during in-flight reconnect recovery, verifying the client remains disconnected with no late recovery propagation.
* **Documentation**
  * Updated the changelog to include the reconnect recovery disposal fix.

### DIFF
--- a/packages/stream_chat/lib/src/client/client.dart
+++ b/packages/stream_chat/lib/src/client/client.dart
@@ -576,11 +576,14 @@ class StreamChatClient {
 
   /// Method called to add a new event to the [_eventController].
   void handleEvent(Event event) {
+    // Ignore events that arrive after the client has been disposed.
+    if (_eventController.isClosed) return;
+
     if (event.type == EventType.healthCheck) {
       return _handleHealthCheckEvent(event);
     }
     state.updateUser(event.user);
-    return _eventController.add(event);
+    return _eventController.safeAdd(event);
   }
 
   void _onConnectionStatusChanged(
--- a/packages/stream_chat/CHANGELOG.md
+++ b/packages/stream_chat/CHANGELOG.md
@@ -21,6 +21,7 @@
 - Fixed watchers not being removed from `ChannelClientState.watchers` on `user.watching.stop`.
 - Fixed `Channel.name`/`image`/`extraData` setters throwing after a *failed* initialization; they now only throw once the channel is successfully initialized.
 - Fixed `Channel.initialized` staying errored after a failed init; it now reflects a subsequent successful (re)initialization.
+- Fixed a `StateError` (`Cannot add new events after calling close`) thrown when the client is disposed while a reconnect recovery is still in flight.
 
 ## 10.2.0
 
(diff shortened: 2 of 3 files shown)


packages/stream_chat/lib/src/client/client.dart
xsahil03x: We should also use `.safeAdd` here and in all the cases where we are using `.add` in client.dart.
```suggestion
    return _eventController.safeAdd(event);
```

xsahil03x (approved):

packages/stream_chat/lib/src/client/client.dart
VelikovPetar: Fixing it! Do you think it makes sense to still keep the `isClosed` guard at the beginning of the `handleEvent` method? If we keep it - we still guard agains writes to `state.updateUser` and handling `health.check`.

xsahil03x: Yes, makes sense to block the other two functions

xsahil03x: we should also add the `.safeAdd` to remaining subjects in the `client.dart` file

VelikovPetar: This seems to be the only remaining unsafe usage - every similar controller already uses `.safeAdd`.

PR #2858, opened 2026-07-31
fix(llc): keep reaction group while its count stays positive

xsahil03x:
## Summary

`Message` dropped an entire `ReactionGroup` in two places when a group's summed scores were **non-positive**, even though its `count` was still positive — so count-based reaction UIs lost other users' reactions:

1. **`deleteMyReaction`** (optimistic delete) kept a group only when `count > 0 && sumScores > 0`. During an optimistic delete, a group whose scores net to zero/negative vanished until the next server event re-hydrated it.
2. **`_reactionGroupsReadValue`** (legacy-payload parse fallback that synthesizes groups from the old `reaction_counts`/`reaction_scores` maps) discarded a group when `sumScores == 0`, dropping a still-populated group at parse time. (Flagged in review by @renefloor.)

## Root cause / backend alignment

The backend derives reaction groups from the `reactions` table via `count(type)` / `sum(score)` grouped by `(message_id, type)`. A group exists **iff at least one reaction row remains** (`count >= 1`); `sum(score)` is an independent aggregate that may legitimately be `0` or **negative** and never gates whether the group exists.

Both call sites now gate purely on the count.

## Changes

- `deleteMyReaction`: retain the group while `updatedCount > 0`.
- `_reactionGroupsReadValue`: retain the group while `count > 0` (guard `count <= 0` so a malformed negative count can't build an invalid group).
- Regression tests: realistic delete-path fixture (own `score: 1`, other `score: -1` → net zero, then negative after delete) and a legacy-payload test covering zero and negative score totals.
- `stream_chat` CHANGELOG entries.

## Testing

`flutter test test/src/core/models/message_test.dart test/src/core/models/message_reaction_helper_test.dart` — all pass.

Fixes FLU-663

### DIFF
--- a/packages/stream_chat/lib/src/core/models/message.dart
+++ b/packages/stream_chat/lib/src/core/models/message.dart
@@ -172,8 +172,8 @@ class Message extends Equatable implements ComparableFieldProvider {
       final count = reactionCounts?[type] ?? 0;
       final sumScores = reactionScores?[type] ?? 0;
 
-      // Keep the group while count is positive; score may be zero.
-      if (count == 0) continue;
+      // Keep the group while count is positive; score may be zero or negative.
+      if (count <= 0) continue;
       final now = DateTime.timestamp();
       groups[type] = {
         'count': count,
@@ -1026,7 +1026,7 @@ extension MessageReactionHelper on Message {
       final group = reactionGroups.remove(type);
       if (group == null) continue;
 
-      // Keep the group while count is positive; score may be zero.
+      // Keep the group while count is positive; score may be zero or negative.
       final updatedCount = group.count - 1;
       final updatedSumScores = group.sumScores - reaction.score;
 
--- a/packages/stream_chat/lib/src/core/models/message.dart
+++ b/packages/stream_chat/lib/src/core/models/message.dart
@@ -1026,10 +1026,16 @@ extension MessageReactionHelper on Message {
       if (group == null) continue;
 
       // Update the reaction group.
+      //
+      // The group exists as long as at least one reaction remains, mirroring
+      // the backend which derives groups from the reaction count alone. The
+      // score is an independent aggregate and must not gate the group, else a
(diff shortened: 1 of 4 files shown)


packages/stream_chat/lib/src/core/models/message.dart
renefloor: One related spot worth a look

message.dart:175 has the same conflation in the legacy-payload fallback that synthesises groups from the old reaction_counts/reaction_scores maps:

if (count == 0 || sumScores == 0) continue;

A message with reaction_counts: {like: 2}, reaction_scores: {like: 0} gets its group discarded at parse time. Narrower blast radius than the delete path (only pre-reaction_groups payloads hit it), but it's the same reasoning error.

packages/stream_chat/test/src/core/models/message_reaction_helper_test.dart
```diff
@@ -361,6 +361,51 @@ void main() {
...
+        );
+
+        final otherUser = User(id: 'other-user-id');
+        final otherReaction = Reaction(
+          type: 'like',
+          score: 0,
+          user: otherUser,
+          userId: otherUser.id,
+          messageId: emptyMessage.id,
+        );
```
renefloor, line 384: The test fixture is unreachable in practice. Both reactions use score: 0, which can't occur with the default score = 1. It pins the regression correctly, but a more representative fixture would be group count: 2, sumScores: 0 with an own reaction of score: 1.

```suggestion
        final ownReaction = Reaction(
          type: 'like',
          score: 1,
          user: testUser,
          userId: testUser.id,
          messageId: emptyMessage.id,
        );

        final otherUser = User(id: 'other-user-id');
        final otherReaction = Reaction(
          type: 'like',
          score: -1,
          user: otherUser,
          userId: otherUser.id,
          messageId: emptyMessage.id,
        );
```

renefloor (approved):

PR #2869, opened 2026-08-05
test(repo): Add channel list and quoted reply e2e suites

testableapple:
# Submit a pull request

Linear: FLU-605
Linear: FLU-606

## CLA

Ports two more e2e suites to the Flutter sample app — **ChannelList** (13 tests) and **QuotedReply** (32 tests) — taking the suite to **82 tests, 2 skipped**. Both are ports of the native Android/iOS suites and reuse identifiers the SDK already exposes, so no test-only keys were added to the SDK.

### e2e suites

- `channel_list_test.dart` — ported from Android `ChannelListTests.kt` + iOS `ChannelList_Tests.swift` (SwiftUI *and* UIKit, which is where the two truncation cases come from).
- `quoted_reply_test.dart` — ported from Android `QuotedReplyTests.kt` plus the 4 cases that only exist in iOS SwiftUI's `QuotedReply_Tests.swift`.
- New page objects (`channel_list_page.dart`, `message_list_page.dart`) and assert mixins for the channel list and message list; `widget_test_extensions.dart` gained the shared `renderedText` / visibility helpers both suites need.
- Every test carries a verified Allure id.

### CI

- The PR workflow now splits the suite across a **2-entry `batch` matrix per platform**, mirroring android's `e2e-test.yml` and swift's `smoke-checks.yml`. Slicing is per test *file*, since a file is one `flutter test` invocation = one native build.
- Because `allure_launch` exports `LAUNCH_ID` per job, batching required hoisting it into its own job — **one per platform** (`allure_launch_android` / `allure_launch_ios`), via the new `.github/actions/allure-launch` composite action. They must stay separate: the reporter's `historyId` carries no platform, so a shared launch would read Android and iOS results for the same test as retries of one test and mask a failure on one platform.

#### Filed & Fixed — moved to #2874

The SDK fixes below, with their unit tests and changelog entries, live in **#2874**, which is based on `master` and reviewable on its own. The e2e tests covering them stay here and **fail until #2874 lands and `master` is merged into this branch** — expected, not a regression.

🔼 [FLU-668](https://linear.app/stream/issue/FLU-668) — `Channel.getReplies` added the parent message to the thread, so the root rendered twice.
🔼 [FLU-669](https://linear.app/stream/issue/FLU-669) — the thread-replies footer hardcoded `'$replyCount replies'`, reading **"1 replies"** and never localized. Asserted by `quoted_reply_test.dart`.
🔼 [FLU-685](https://linear.app/stream/issue/FLU-685) — a truncated channel kept its timestamp next to "No messages yet". iOS hides it (UIKit scenario 284); the ported e2e test had been written to Flutter's behaviour instead and now asserts `isDisplayed: false`.
🔼 [FLU-686](https://linear.app/stream/issue/FLU-686) — a channel-list row could preview another channel's last message after a reorder. Surfaced by review, not by a test.

Still fixed here, since it is sample-app code the suites drive rather than an SDK change:

🟢 `sample_app`'s `thread_page.dart` dropped `showScrollToBottom: false`, so threads get the same scroll-to-bottom button as channels (matching `channel_page.dart` and the native suites).
⚪ `StreamEphemeralMessage` is now exported (with a changelog entry) — the only `packages/` change left in this PR, because `message_list_page.dart` references it and the suites would not compile otherwise.

#### Filed

🔴 [FLU-670](https://linear.app/stream/issue/FLU-670) — the message-list loading skeleton overflows in a short viewport (Android only, so skipped only there).
🔴 [FLU-671](https://linear.app/stream/issue/FLU-671) — jumping to a quote inside a thread doesn't page in a target that isn't loaded.

### Test harness

- `streamTest` / `streamTestWithEnv` gained **`skipPlatforms`**, so a `skip:` can be narrowed to the platform where the issue reproduces. Used by one test whose failure is Android-only.
- `streamTest` now logs the full `FlutterErrorDetails` — `takeException()` hands back only the exception, so a layout error used to report "203 pixels" without ever naming the widget.
- Indexed finders (`find.byType(X).at(n)`) **throw** when the list hasn't rendered yet instead of resolving empty, which killed two Android tests mid-poll; `evaluateSafely()` now absorbs that so the wait loops actually wait.

### How this was tested

Whole `integration_test/` dir on the iPhone 17 (iOS 26.5) simulator, plus per-suite runs on device, taken while the SDK fixes were still on this branch. With them moved to #2874 the suites covering FLU-668/669/685 fail here by design. `dart analyze --fatal-infos` is clean on `integration_test/` and on `stream_chat_flutter`.

## Screenshots / Videos

N/A — test and CI changes only; the SDK fixes and their user-visible effects moved to #2874.

## Summary by CodeRabbit

* **New Features**
  * Added `StreamEphemeralMessage` to the public Flutter package exports.
  * Enabled the scroll-to-bottom control in thread conversations.

* **Bug Fixes**
  * Prevented duplicate parent messages in thread replies.
  * Corrected localized, singularized thread reply labels.
  * Improved channel preview timestamps after message updates or truncation.

* **Tests**
  * Expanded coverage for channel lists, quoted replies, threads, pagination, message states, and offline behavior.
  * Improved Android and iOS end-to-end test batching, artifact reporting, and reliability.


packages/stream_chat_flutter/lib/stream_chat_flutter.dart
```diff
@@ -97,6 +97,7 @@ export 'src/message_modal/moderated_message_actions_modal.dart';
 export 'src/message_widget/components/stream_message_footer.dart';
 export 'src/message_widget/components/stream_message_header.dart';
 export 'src/message_widget/components/stream_message_leading.dart';
+export 'src/message_widget/stream_ephemeral_message.dart';
```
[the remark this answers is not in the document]
testableapple: This is the existing convention here rather than an oversight: `message_list_view.dart` takes public types from the barrel and reaches into `src/` only for files the barrel doesn't export — all 10 of its direct `src/` imports are non-exported (`mlv_utils.dart`, `thread_separator.dart`, `empty_widget.dart`, …), while `stream_ephemeral_message.dart` is exported on the very line you flagged. Its immediate neighbours `StreamSystemMessage` (`:1106`) and `StreamModeratedMessage` (`:1116`) resolve the same way, so a direct import would split one build-method cluster across two import styles; package-wide it's 123 of 268 files under `lib/src/` importing the barrel. Leaving as is.

packages/stream_chat/lib/src/client/channel.dart
```diff
@@ -3927,7 +3927,7 @@ class ChannelClientState {
     final threadMessages = updatedThreads[parentId] ?? <Message>[];
     final updatedThreadMessages = _mergeMessagesIntoExisting(
       existing: threadMessages,
-      toMerge: messages,
+      toMerge: messages.where((it) => it.id != parentId),
```
testableapple, line 3930: A thread holds replies only, the parent is rendered from its own copy, so letting it in here would show the root twice. The offline path already filters on `parent_id`, this keeps the online one in line.

packages/stream_chat_flutter/lib/src/message_widget/stream_message_item.dart
```diff
@@ -506,7 +506,7 @@ class DefaultStreamMessageItem extends StatelessWidget {
           maxAvatars: 3,
           onTap: () => _onViewThread(context, message),
           showConnector: placement.contentKind != .jumbomoji,
-          label: Text('$replyCount replies'),
+          label: Text(context.translations.threadReplyCountText(replyCount)),
```
VelikovPetar, line 509: Can we cover this with a unit test (single vs multiple replies)?

sample_app/integration_test/channel_list_test.dart
```diff
@@ -0,0 +1,361 @@
...
+      await env.userRobot.assertMessage(sampleText);
+
+      step('AND the participant deletes the message');
+      await env.participantRobot.deleteMessage();
+
+      step('WHEN the user goes back to the channel list');
+      await env.userRobot.tapBackButton();
+
+      step('THEN the channel preview shows the deleted message');
+      // Android expects "No messages" here, because its preview drops deleted
```
VelikovPetar: Not really a big deal: Do we want to reference other platforms here (Android/iOS)? I am afraid that changing something in the referenced platform would make this comment outdated -> and introduce confusions about what is really correct.

Also I think the comment is incorrect, Android expects "Message Deleted" as well: https://github.com/GetStream/stream-chat-android/blob/develop/stream-chat-android-compose-sample/src/androidTestE2eDebug/kotlin/io/getstream/chat/android/compose/tests/ChannelListTests.kt#L167

sample_app/integration_test/channel_list_test.dart
```diff
@@ -0,0 +1,361 @@
...
+      await env.userRobot.assertMessage(newMessage);
+
+      step('AND the participant deletes the last message');
+      await env.participantRobot.deleteMessage();
+
+      step('WHEN the user goes back to the channel list');
+      await env.userRobot.tapBackButton();
+
+      step('THEN the channel preview shows the deleted message');
+      // Flutter keeps a soft-deleted message as the channel's last message and
```
VelikovPetar: Same as above

sample_app/integration_test/channel_list_test.dart
```diff
@@ -0,0 +1,361 @@
...
+
+      step('WHEN the user goes back to the channel list');
+      await env.userRobot.tapBackButton();
+
+      step('THEN the channel preview is empty');
+      await env.userRobot.assertMessageInChannelPreview('No messages yet');
+
+      step('AND the message timestamp is still shown');
+      // iOS expects it hidden here. On Flutter `lastMessageAt` lives on the
+      // channel model and truncation only empties `state.messages`, so nothing
```
VelikovPetar: Hmm could this be a bug? - I think we should investigate this, or at least check how Android/iOS/RN handle this case.

sample_app/integration_test/quoted_reply_test.dart
```diff
@@ -0,0 +1,786 @@
...
+        messagesCount: 1,
+        messagesText: parentText,
+        repliesCount: pageSize,
+      );
+      await env.userRobot.login().openChannel();
+
+      step('WHEN the user opens the thread and scrolls up to the top of it');
+      // Native scrolls twice; a swipe there is a full screen, `scrollMessageListUp`
+      // drags 300px, so reaching the top of the thread takes more of them.
+      await env.userRobot.openThreadFromReplies().scrollMessageListUp(times: 8);
```
VelikovPetar: I am a bit concerned that `scrollMessageListUp(times: 8)` could become flaky: For example, if we change the emulator/simulator on which the tests run to one with a different screen size, 8 scrolls up might not be enough to reach the desired message. Probably not something that will happen often (or ever), but I still wanted to bring this up (not sure if there is a more reliable way to scroll to the end of the list).

I don't see this as a blocker, maybe just something to improve if it is quick.

packages/stream_chat_flutter/lib/src/message_widget/stream_message_item.dart
testableapple: Added a couple of widget tests for this👍

sample_app/integration_test/channel_list_test.dart
testableapple: My bad, i did not clean this up. I mostly porting things from iOS/Android and adjusting them for Flutter, these comments are definitely useless. Removed them 👍

sample_app/integration_test/channel_list_test.dart
testableapple: Removed 👍

sample_app/integration_test/channel_list_test.dart
testableapple: nice catch, you are right! i filed a ticket (https://linear.app/stream/issue/FLU-685) and fixed it as part of this PR and added unit tests as well, please check it out

sample_app/integration_test/quoted_reply_test.dart
testableapple: i think you are right and it makes sense to fix it now before we introduce more tests and emulators/simulators. Updated 👍

VelikovPetar (approved):

PR #2870, opened 2026-08-06
feat(ui, localization, samples): message translations

renefloor:
# Submit a pull request

Linear: Fixes FLU-673

## CLA

The current SDK already supports translating messages, but the UI package misses crucial features. This PR adds the annotations from the new design system that shows the fact that it's translated and you can switch between original and translated message. The SDK also adds a flag `StreamChatConfigurationData.translationDisplayEnabled` so you can also disable auto translated messages.

The sample app adds a button in the message actions so you can request a translation, as the sdk never automatically translate messages, although the backend can do that for you. The button to translate is disabled if the message is already translated to your language.

## Screenshots / Videos

image
image
image

## Summary by CodeRabbit

* **New Features**
  * Messages now display available translations automatically, with controls to switch between translated and original text.
  * Translation display and annotations can be configured independently, with localized labels and source-language information.
  * Added translation requests, reaction long-press actions, optional safe-area handling, and customizable surface styles in the sample app.
  * Message translation components are publicly available for customization.

* **Bug Fixes**
  * Prevented unintended English translation when no language is configured.
  * Preserved original message text and previews when translation is unavailable.
  * Translation state now persists as lists update or paginate.


packages/stream_chat_localizations/example/lib/add_new_lang.dart
```diff
@@ -815,6 +815,21 @@ class NnStreamChatLocalizations extends GlobalStreamChatLocalizations {
...
+  String get originalLabel => 'Original';
+
+  @override
+  String get showOriginalLabel => 'Show original';
+
+  @override
+  String get showTranslationLabel => 'Show translation';
+
+  @override
+  String translatedFromLanguageText(String languageCode) => 'Translated from ${languageCode.toUpperCase()}';
```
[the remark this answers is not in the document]
renefloor: The doc comments for translatedLabel, originalLabel, showOriginalLabel, showTranslationLabel, and translatedFromLanguageText already exist on the abstract interface (stream_chat_flutter/lib/src/localization/translations.dart:730-756). Every concrete implementation in the repo — the base DefaultTranslations, all locale files in stream_chat_localizations/lib/src/, and this example file — consistently omits doc comments on `@override` members and relies on doc inheritance. Adding comments only in the example file would be inconsistent with the established convention, so I left it as-is.

packages/stream_chat_flutter/lib/src/stream_chat_configuration.dart
```diff
@@ -165,6 +165,7 @@ class StreamChatConfigurationData {
     List<StreamAttachmentWidgetBuilder>? attachmentBuilders,
     StreamReactionsType? reactionType,
     StreamReactionsPosition? reactionPosition,
+    bool translationDisplayEnabled = true,
```
VelikovPetar, line 168: Should we maybe default to `false`? Otherwise a customer might start seeing this when updating the SDK, which is kind of a breaking change. I understand that it is a new feature, but what I am worried is that there is customer who already handles the translation stuff on their own - and after updating they could potentially get duplicated labels/handlers for the translation.

packages/stream_chat_flutter/lib/src/stream_chat_configuration.dart
```diff
@@ -258,4 +263,14 @@ class StreamChatConfigurationData {
...
+
+  /// Whether a message automatically displays as its translation — and
+  /// shows a "Translated" annotation with a link to see the original —
+  /// when [Message.i18n] has one for the current user's language.
+  ///
+  /// Set to `false` to always show a message's original text regardless of
+  /// available translations.
+  ///
+  /// Defaults to `true`.
+  final bool translationDisplayEnabled;
```
VelikovPetar, line 275: Something maybe to consider: On Android, this config is two-part:
```kotlin
public data class TranslationConfig(
    val enabled: Boolean = true,
    val showOriginalEnabled: Boolean = true,
)
```
- `TranslationConfig.enabled` -> controls whether the message automatically displays its translation
- `TranslationConfig.showOriginalEnabled` -> controls whether the `Show original` is toggled. 

It was implemented like this because the features came at a different time and to avoid breaking changes - but maybe it makes sense if a customer would like to use auto translation, but disable the "Show original" toggle. 

_Again, just a suggestion, since the whole feature is new, we can keep it simple._

packages/stream_chat_flutter/lib/src/utils/extensions.dart
```diff
@@ -459,8 +459,15 @@ extension MessageX on Message {
...
   }
 
-  /// It returns the message with the translated text if available locally
-  Message translate(String language) => copyWith(text: i18n?['${language}_text'] ?? text);
+  /// It returns the message with the translated text if available locally.
+  ///
+  /// Returns the message unchanged when [language] is `null` or empty —
+  /// Stream's API defaults [User.language] to `''` rather than omitting it,
+  /// so both are treated as "no language to translate to".
+  Message translate(String? language) => switch (language) {
```
VelikovPetar, line 467: Perhaps we can add/extend the unit test for this?

packages/stream_chat_flutter/lib/src/stream_chat_configuration.dart
renefloor: I wonder why you would ever want to hide the `show original` button? I can only imagine that we want 2 toggles for showing translations at all and showing the translated message automatically. Or is that what is meant with `showOriginalEnabled`?

packages/stream_chat_flutter/lib/src/stream_chat_configuration.dart
renefloor: Translating already happens automatically on the master branch, so defaulting to false would be more breaking. This PR only adds the option to toggle between translation and original.
We currently translate already in [stream_message_text](https://github.com/GetStream/stream-chat-flutter/blob/master/packages/stream_chat_flutter/lib/src/message_widget/components/stream_message_text.dart#L70) and [stream_message_preview](https://github.com/GetStream/stream-chat-flutter/blob/master/packages/stream_chat_flutter/lib/src/channel/stream_message_preview_text.dart#L38).

Looks like we even try to default to english translations at the moment, which is odd.

packages/stream_chat_flutter/lib/src/stream_chat_configuration.dart
VelikovPetar: Yes the `showOriginalEnabled` controls whether the "Show original"/"Show translated" toggle is shown in the message view. 

On Android the "Show original" feature request came much later compared to the "Auto translation". So in order to be backwards compatible, we hid the "Show original" button by default - and customers can opt-in via the config to actually show the toggle (this has now changed with the major v7, and both flags are true by default, but they still exist separately). 

Now that we are introducing "both" features as a package - maybe it doesn't make sense to have them configured separately. But still, I think we should consider the possibility of a customer not wanting to show the "Show original" toggle, even when they use auto-translation.

renefloor: In that case maybe it's better to name it `showTranslationAnnotation` as it's also hiding 'show translation'. But we do want to have less configuration options and people now have a slot to manage their annotations directly, so I think it's better to write a doc on how to manage your own annotations.

packages/stream_chat_flutter/lib/src/stream_chat_configuration.dart
VelikovPetar: Aha OK, that makes sense (I saw that we already do the translation automatically, but for some reason this flag made me think that we weren't doing that previously...). In that case `true` is fine.

packages/stream_chat_flutter/lib/src/stream_chat_configuration.dart
VelikovPetar: Yeah I like that naming better. One thing that I can suggest is still wrapping the `translationDisplayEnabled` flag into a config object, something along the lines of:
```dart
class StreamTranslationsConfiguration {
  final bool translationDisplayEnabled;
}
```
So we can easily extend it (potentially with `showTranslationAnnotation`) without breaking changes, and without polluting the `StreamChatConfiguration` with multiple translations-related fields. 

> But we do want to have less configuration options and people now have a slot to manage their annotations directly

I am still a bit worried about having the "Show original/translation" displayed by default - mostly because of the reason I mentioned earlier about showing "double" labels (one from the SDK, and one from the integrator app).

xsahil03x: I personally don't like the name `translationDisplayEnabled` i think `showTranslationAnnotation` is better suited here

renefloor: But I would still prefer not to have this setting at all, but just explain how to disable using the header builder of the message in the factory.

sample_app/lib/widgets/custom_message_actions.dart
```diff
@@ -201,3 +204,69 @@ abstract final class _MessageInfoAction {
...
+    // Already covers the message's own language: the server includes a
+    // self-referential entry keyed by its source language, so this is also
+    // `true` when the original text is already in the user's language —
+    // either way, there's nothing left to translate.
+    final alreadyTranslated = message.i18n?['${userLanguage}_text'] != null;
+
+    final icons = context.streamIcons;
+    return [
+      StreamContextMenuAction(
+        label: const Text('Translate'),
```
xsahil03x, line 248:
```suggestion
        label: const Text('Translate Message'),
```

sample_app/lib/widgets/custom_message_actions.dart
```diff
@@ -201,3 +204,69 @@ abstract final class _MessageInfoAction {
...
+///
+/// Once merged, the message's `i18n` map carries the translation, and the
+/// SDK's own message rendering (`StreamMessageText`, `DefaultStreamMessageHeader`)
+/// picks it up automatically — this sample app only needs to fetch it.
+///
+/// Only shown for messages from other users — translating your own message
+/// isn't a useful action — and always translates directly to the current
+/// user's [User.language], with no language picker. Hidden entirely when
+/// that isn't set server-side, rather than guessing a target language.
+abstract final class _TranslateAction {
```
xsahil03x, line 223:
```suggestion
abstract final class _TranslateMessageAction {
```

sample_app/lib/widgets/custom_message_actions.dart
```diff
@@ -201,3 +204,69 @@ abstract final class _MessageInfoAction {
...
+      StreamContextMenuAction(
+        label: const Text('Translate'),
+        leading: Icon(icons.translate),
+        enabled: !alreadyTranslated,
+        onTap: () => _translate(context, message, userLanguage),
+      ),
+    ];
+  }
+
+  static Future<void> _translate(
```
xsahil03x, line 256:
```suggestion
  static Future<void> _translateMessage(
```

sample_app/lib/widgets/custom_message_actions.dart
```diff
@@ -201,3 +204,69 @@ abstract final class _MessageInfoAction {
...
+    Message message,
+    String language,
+  ) async {
+    final channel = StreamChannel.of(context).channel;
+    try {
+      final response = await channel.translateMessage(message.id, language);
+      channel.state?.updateMessage(response.message);
+    } catch (e) {
+      if (!context.mounted) return;
+      ScaffoldMessenger.of(context).showSnackBar(
```
xsahil03x, line 267: Let's use our StreamSnackbar instead

packages/stream_chat_flutter/lib/src/message_widget/stream_message_translation_store.dart
```diff
@@ -0,0 +1,48 @@
...
+/// Tracks, for messages within a single message list, which ones are
+/// currently showing their original text instead of a translation.
+///
+/// One store is owned per [StreamMessageListView] instance — a channel view
+/// and an open thread each get independent toggle state, since opening a
+/// thread creates its own [StreamMessageListView]. The store, rather than
+/// the individual message widget, is what makes the toggle survive message
+/// items being disposed and recreated as they scroll out of and back into
+/// the list's render window.
+class StreamMessageTranslationStore extends ChangeNotifier {
```
xsahil03x, line 12: we can use a value notifier here

packages/stream_chat_flutter/lib/src/stream_chat_configuration.dart
VelikovPetar: Still my main concern is customers reporting the new label appearing by default, without any opt-in after updating the SDK. But if you think it is safe to do that, I wouldn't push on adding this flag.

packages/stream_chat_flutter/lib/src/message_widget/components/stream_message_text.dart
```diff
@@ -30,11 +34,16 @@ class StreamMessageText extends StatelessWidget {
...
     this.onAnyMentionTap,
+    this.showOriginalText = false,
   });
 
   /// The message whose text to display.
   final Message message;
 
+  /// Whether to display [message]'s original text instead of a translation,
+  /// when [Message.i18n] has one for the current user's language.
+  final bool showOriginalText;
```
xsahil03x: I would rename this to `showTranslatedText` instead

packages/stream_chat_flutter/lib/stream_chat_flutter.dart
```diff
@@ -14,6 +14,7 @@ export 'package:stream_core_flutter/chat.dart'
         // additional features and customization options.
         StreamMessageContent,
         StreamMessageContentProps,
+        StreamMessageText,
```
renefloor: @xsahil03x we've hidden `StreamMessageContent` already, but didn't export `StreamMessageContent` from chat. `StreamMessageText` from core was exposed, but it makes more sense to have `StreamMessageText` from chat exposed. That does make it kind of a breaking change though, but a useful one and I don't think many already use `StreamMessageText`.

packages/stream_chat_flutter/lib/src/message_widget/components/stream_message_text.dart
renefloor: done in https://github.com/GetStream/stream-chat-flutter/pull/2870/commits/10651b5596b71d1bebda53ba4608f5c65a3e9ffc

sample_app/lib/widgets/custom_message_actions.dart
renefloor: done in https://github.com/GetStream/stream-chat-flutter/commit/10651b5596b71d1bebda53ba4608f5c65a3e9ffc

packages/stream_chat_flutter/lib/src/message_widget/stream_message_translation_store.dart
renefloor: done in https://github.com/GetStream/stream-chat-flutter/commit/10651b5596b71d1bebda53ba4608f5c65a3e9ffc
improved in https://github.com/GetStream/stream-chat-flutter/pull/2870/commits/20af936721183ac2da0c32f3e381dc15f41f18fe

packages/stream_chat_flutter/lib/src/stream_chat_configuration.dart
renefloor: Created a config class; https://github.com/GetStream/stream-chat-flutter/pull/2870/changes/10651b5596b71d1bebda53ba4608f5c65a3e9ffc#diff-a1c0525e7537e25cb3080dffa23a1f8fdd1d31f33c42212fa18fd2ab38774ac1

defaults are back as they were, so annotation doesn't show automatically
```dart
  const StreamMessageTranslationConfiguration({
    this.enabled = true,
    this.annotationEnabled = false,
  });
  ```

File: packages/stream_chat_flutter/lib/src/utils/extensions.dart (hunk 4 above)
Comment (renefloor) [reply packages/stream_chat_flutter/lib/src/utils/extensions.dart]: more tests are added here: https://github.com/GetStream/stream-chat-flutter/pull/2870/changes/10651b5596b71d1bebda53ba4608f5c65a3e9ffc#diff-24a0a21c912a04f109c2af71ce32b7a2140546b27e0ccad91f4a79166ea32375

File: packages/stream_chat_flutter/lib/src/stream_chat_configuration.dart (hunk 3 above)
Comment (VelikovPetar) [reply packages/stream_chat_flutter/lib/src/stream_chat_configuration.dart]: LGTM!

File: packages/stream_chat_flutter/lib/src/message_widget/components/stream_message_content.dart [hunk 12]
```diff
@@ -44,6 +44,7 @@ class StreamMessageContent extends StatefulWidget {
     this.onReactionLongPress,
     this.onQuotedMessageTap,
     this.reactionSorting,
+    this.showTranslatedText = true,
```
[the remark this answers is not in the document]
renefloor: StreamMessageContent is not exposed outside of the library in the master branch right?

packages/stream_chat_flutter/lib/src/message_widget/components/stream_message_text.dart
```diff
@@ -67,12 +77,19 @@ class StreamMessageText extends StatelessWidget {
...
-    return BetterStreamBuilder<String>(
-      initialData: streamChat.currentUser?.language ?? 'en',
-      stream: streamChat.currentUserStream.map((it) => it?.language ?? 'en'),
-      builder: (context, language) {
-        final messageText = message.translate(language).replaceMentions().text?.replaceAll('\n', '\n\n').trim();
+    // `BetterStreamBuilder`'s type parameter can't be nullable, so the
+    // current user's (possibly unset) language is wrapped in a record —
+    // itself always non-null — rather than defaulting it to an empty string.
+    return BetterStreamBuilder<({String? language})>(
+      initialData: (language: streamChat.currentUser?.language),
```
xsahil03x: do we need a record here?

packages/stream_chat_flutter/lib/src/message_widget/stream_message_item.dart
```diff
@@ -467,6 +466,9 @@ class DefaultStreamMessageItem extends StatelessWidget {
   Widget build(BuildContext context) {
```
xsahil03x, line 466: We should probably add a debug assertion if the StreamMessageTranslations is available... similar to how material does it .

```dart
    assert(debugCheckHasMessageTranslations(context));
```

packages/stream_chat_flutter/lib/src/message_widget/stream_message_translation_store.dart
```diff
@@ -0,0 +1,107 @@
...
+  /// Creates a message translation scope.
+  const StreamMessageTranslations({
+    super.key,
+    required StreamMessageTranslationStore store,
+    required super.child,
+  }) : super(notifier: store);
+
+  /// Returns the [StreamMessageTranslationStore] from the nearest ancestor
+  /// [StreamMessageTranslations], or `null` if there isn't one.
+  static StreamMessageTranslationStore? of(BuildContext context) {
```
xsahil03x: lets create a separate `maybeOf` for nullable returns, `.of` should not return a nullable value. We can match how we do it in StreamChat etc.

```dart
  /// Finds the [StreamChatState] from the closest [StreamChat] ancestor
  /// that encloses the given [context].
  ///
  /// This will throw a [FlutterError] if no [StreamChat] is found in the
  /// widget tree above the given context.
  ///
  /// Typical usage:
  ///
  /// ```dart
  /// final chatState = StreamChat.of(context);
  /// ```
  ///
  /// If you're calling this in the same `build()` method that creates the
  /// `StreamChat`, consider using a `Builder` or refactoring into a
  /// separate widget to obtain a context below the [StreamChat].
  ///
  /// If you want to return null instead of throwing, use [maybeOf].
  static StreamChatState of(BuildContext context) {
    final result = maybeOf(context);
    if (result != null) return result;

    throw FlutterError.fromParts(<DiagnosticsNode>[
      ErrorSummary(
        'StreamChat.of() called with a context that does not contain a '
        'StreamChat.',
      ),
      ErrorDescription(
        'No StreamChat ancestor could be found starting from the context '
        'that was passed to StreamChat.of(). This usually happens when the '
        'context used comes from the widget that creates the StreamChat '
        'itself.',
      ),
      ErrorHint(
        'To fix this, ensure that you are using a context that is a descendant '
        'of the StreamChat. You can use a Builder to get a new context that '
        'is under the StreamChat:\n\n'
        '  Builder(\n'
        '    builder: (context) {\n'
        '      final chatState = StreamChat.of(context);\n'
        '      ...\n'
        '    },\n'
        '  )',
      ),
      ErrorHint(
        'Alternatively, split your build method into smaller widgets so that '
        'you get a new BuildContext that is below the StreamChat in the '
        'widget tree.',
      ),
      context.describeElement('The context used was'),
    ]);
  }

  /// Finds the [StreamChatState] from the closest [StreamChat] ancestor
  /// that encloses the given context.
  ///
  /// Returns null if no such ancestor exists.
  ///
  /// See also:
  ///  * [of], which throws if no [StreamChat] is found.
  static StreamChatState? maybeOf(BuildContext context) {
    return StreamStateScope.maybeOf<StreamChatState>(context);
  }
```

packages/stream_chat_flutter/lib/src/message_widget/components/stream_message_text.dart
renefloor: Currently the BetterStreamBuilder needs a non-nullable type. Which would you prefer here? 

1. Keep the record — explicit that "unset" is a real state, no sentinel value.
2. `language ?? ''` — shortest, at the cost of `''` standing in for "unset".
3. Let `BetterStreamBuilder` take a nullable `T` — nicest at the call site, but it's declared `BetterStreamBuilder<T extends Object>` and `noDataBuilder` keys off `data == null`, so it needs a rework of the builder itself.

Happy with any of them; the record was just the option that needed no changes elsewhere.

packages/stream_chat_flutter/lib/src/message_widget/stream_message_item.dart
renefloor: In this PR we inject the StreamMessageTranslations in the MessageListView. Most customers will probably use that, but we do have separate message widgets in our tests, they would all break. Also customers that somehow show the messages differently have it breaking. We could only assert when customer has annotationsEnabled, as those are certain they want the translations. But I think I would prefer not to assert it here.

packages/stream_chat_flutter/lib/src/message_widget/stream_message_translation_store.dart
renefloor: Added: https://github.com/GetStream/stream-chat-flutter/pull/2870/commits/c244dce704668edca658f4f86a7528486a04dc01

packages/stream_chat_flutter/lib/src/message_widget/components/stream_message_text.dart
xsahil03x: lets make the BetterStreamBuilder nullable

packages/stream_chat_flutter/lib/src/message_widget/stream_message_item.dart
xsahil03x: Yeah, now its better that we use `.maybeOf` :+1:

packages/stream_chat_flutter/lib/src/stream_chat.dart
```diff
@@ -183,9 +183,24 @@ class StreamChatState extends State<StreamChat> {
...
+  /// Owned here so the toggle is shared by every message list in the app — a
+  /// channel and its open thread render the same parent message, and both
+  /// should agree on which text it shows. Matches the Swift and Android SDKs,
+  /// which key this state by message id above the individual list.
+  late final StreamMessageTranslationStore _translationStore;
+
   @override
   void initState() {
     super.initState();
+    _translationStore = StreamMessageTranslationStore();
```
xsahil03x: I think we can do it inline instead of overriding the initState.

packages/stream_chat_flutter/lib/src/message_widget/stream_message_translation_store.dart
```diff
@@ -101,20 +101,64 @@ class StreamMessageTranslationStore extends ValueNotifier<StreamMessageTranslati
...
+  ///
+  /// The given [context] is rebuilt when *this* message is toggled, and not
+  /// when another message in the same scope is. Prefer this over reading the
+  /// store through [of], which rebuilds on every change.
+  ///
+  /// This will throw a [FlutterError] if no [StreamMessageTranslations] is
+  /// found in the widget tree above the given context.
+  static bool isShowingOriginalTextOf(BuildContext context, String messageId) {
+    final scope = InheritedModel.inheritFrom<_StreamMessageTranslationsScope>(context, aspect: messageId);
+    if (scope != null) return scope.state.isShowingOriginalText(messageId);
```
xsahil03x: let match the style as we have in `toggleOriginalText`. throwing early.

packages/stream_chat_flutter/lib/src/message_widget/stream_message_translation_store.dart
```diff
@@ -101,20 +101,64 @@ class StreamMessageTranslationStore extends ValueNotifier<StreamMessageTranslati
...
+  /// its translation, in the nearest ancestor scope.
+  ///
+  /// The given [context] is rebuilt when *this* message is toggled, and not
+  /// when another message in the same scope is. Prefer this over reading the
+  /// store through [of], which rebuilds on every change.
+  ///
+  /// This will throw a [FlutterError] if no [StreamMessageTranslations] is
+  /// found in the widget tree above the given context.
+  static bool isShowingOriginalTextOf(BuildContext context, String messageId) {
+    final scope = InheritedModel.inheritFrom<_StreamMessageTranslationsScope>(context, aspect: messageId);
```
xsahil03x: WE should use `.of	` here directly right? we don't have to handle `throw _missingScopeError(context);` ourself

packages/stream_chat_flutter/lib/src/message_widget/stream_message_translation_store.dart
```diff
@@ -101,20 +101,64 @@ class StreamMessageTranslationStore extends ValueNotifier<StreamMessageTranslati
...
+  /// translation, in the nearest ancestor scope.
+  ///
+  /// A convenience for `StreamMessageTranslations.of(context)
+  /// .toggleOriginalText(messageId)` that does not make [context] depend on
+  /// the scope, so it is safe to call from a callback.
+  ///
+  /// This will throw a [FlutterError] if no [StreamMessageTranslations] is
+  /// found in the widget tree above the given context.
+  static void toggleOriginalText(BuildContext context, String messageId) {
+    final scope = context.getInheritedWidgetOfExactType<_StreamMessageTranslationsScope>();
```
xsahil03x, line 150: https://github.com/GetStream/stream-chat-flutter/pull/2870#discussion_r3852606043

packages/stream_chat_flutter/CHANGELOG.md
```diff
@@ -42,7 +42,7 @@
...
+- Added `StreamMessageTranslationStore`, tracking the annotation's toggle state by message id. `StreamChat` owns one and provides it through a `StreamMessageTranslations` scope, so a channel list and its open thread agree on which text a message shows; nest another `StreamMessageTranslations` to isolate a subtree, and call `clear()` to switch everything back to its translation. Read a single message's state with `StreamMessageTranslations.isShowingOriginalTextOf(context, messageId)`, which rebuilds the caller only when that message is toggled. Also drivable via `showTranslatedText` on `StreamMessageHeaderProps`, `StreamMessageContent`, and `StreamMessageText`, and `onToggleTranslatedText` on `StreamMessageHeaderProps`.
```
xsahil03x: This changelog entry is too big and verbose

renefloor: fixed: https://github.com/GetStream/stream-chat-flutter/pull/2870/commits/c2e3beb947c30ee5058f06d7938d2944b7ed7a63

packages/stream_chat_flutter/lib/src/message_widget/stream_message_translation_store.dart
renefloor: fixed: https://github.com/GetStream/stream-chat-flutter/pull/2870/commits/c2e3beb947c30ee5058f06d7938d2944b7ed7a63

packages/stream_chat_flutter/lib/src/message_widget/stream_message_translation_store.dart
renefloor: fixed: https://github.com/GetStream/stream-chat-flutter/pull/2870/commits/c2e3beb947c30ee5058f06d7938d2944b7ed7a63

packages/stream_chat_flutter/lib/src/message_widget/stream_message_translation_store.dart
renefloor: fixed: https://github.com/GetStream/stream-chat-flutter/pull/2870/commits/c2e3beb947c30ee5058f06d7938d2944b7ed7a63

xsahil03x (approved):

PR #2871, opened 2026-08-06
feat(ui, core, llc, localization): unread banners

renefloor:
# Submit a pull request

Linear: 
fixes FLU-672 (main issue)
fixes FLU-640
fixes FLU-648
fixes FLU-649
fixes FLU-650

Github Issue: #

## CLA

The requirements changed again while this was being build, so the linear tickets are not right.

Requirements: 
* When opening a channel with unread messages it shows the unread banner. That banner will stay there until you close the channel page and the message count increases when you get new messages.
* The scroll-to-bottom button only gets a count when message are received while you are scrolled up. The count goes away when you're at the bottom
* The pill at the top shows only when you open a channel with unreads and disappears after you get past the first message.
* Channel should open at the newest message, but that's kind of breaking so I've made it an 'opt-in' feature of the `StreamChannel` object.
* No separate 'new messages' banner when messages arrive while having the chat open.

## Screenshots / Videos

https://github.com/user-attachments/assets/d517932a-be28-4f9e-8594-57707c20a7f8

https://github.com/user-attachments/assets/f0d57385-1a06-47d0-93fb-c2afdcdc787f

https://github.com/user-attachments/assets/b329bbb9-070e-4ee8-9c52-4e26eff01e07

## Summary by CodeRabbit

- **New Features**
  - Added channel unread-state tracking and an `openAtFirstUnread` option.
  - Unread indicators can display provided counts or derive visibility from channel state.
  - Unread separators now support localized singular and plural message counts.
  - Exported the unread-message separator for public use.
- **Bug Fixes**
  - Improved unread counting, scrolling, dismissal, reset, and navigation.
  - Refined mark-read and manual mark-unread behavior.
  - Excluded non-qualifying messages from unread counts.
- **Deprecations**
  - Deprecated count-less separator translations with fallback compatibility.


packages/stream_chat_flutter/lib/src/message_list_view/mark_read_details.dart
```diff
@@ -0,0 +1,42 @@
+/// The information available when deciding whether to automatically mark a
+/// [StreamMessageListView]'s channel as read.
+///
+/// Passed to a caller-supplied predicate on
+/// [StreamMessageListViewConfiguration.shouldMarkRead]. Not intended to be
+/// constructed directly.
```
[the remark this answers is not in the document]
renefloor: Fixed differently than suggested: rather than switching this configuration file to the barrel import (that file already uses direct `src/` imports throughout — e.g. `auto_scroll_policy.dart` — which is the established pattern in this directory), I imported the barrel into `mark_read_details.dart` itself in [27e1a51](https://github.com/GetStream/stream-chat-flutter/commit/27e1a51a7). That file had zero imports, so its `[StreamMessageListView]`/`[...shouldMarkRead]` dartdoc references weren't resolving — that's the actual defect. Fixed in 0808f07.

packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
```diff
@@ -433,7 +583,10 @@ class _StreamMessageListViewState extends State<StreamMessageListView> {
     _itemPositionListener.itemPositions.removeListener(_handleItemPositionsChanged);
     debouncedMarkRead.cancel();
     debouncedMarkThreadRead.cancel();
-    _unreadState.dispose();
+    _unreadDivider.dispose();
+    _unreadDividerGrowth.dispose();
+    _hasSeenFirstUnread.dispose();
+    _scrollToBottomBadge.dispose();
```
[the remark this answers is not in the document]
renefloor: Good catch, fixed in [27e1a51](https://github.com/GetStream/stream-chat-flutter/commit/27e1a51a7) — added `_showScrollToBottom.dispose()` to `dispose()` alongside the other notifiers.

packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
```diff
@@ -913,20 +1097,25 @@ class _StreamMessageListViewState extends State<StreamMessageListView> {
...
+    // immediately from the server `Read`, rather than doing nothing.
+    final anchorId = _unreadDivider.value.anchorId ?? _unreadBaseline?.lastReadMessageId;
+    if (anchorId == null) return;
+
+    _hasSeenFirstUnread.value = true;
+    // Delegates to [_scrollToMessage], which falls back to
+    // [StreamChannelState.loadChannelAtMessage] when the anchor isn't in the
+    // currently loaded window — after which the real anchor resolves
+    // naturally via the retry in [_buildListView], rendering divider A too.
+    await _scrollToMessage(messageId: anchorId, highlight: false);
```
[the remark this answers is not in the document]
renefloor: Fixed in [0808f07](https://github.com/GetStream/stream-chat-flutter/commit/0808f07d6) — `_scrollToMessage` now returns `Future<bool>` reporting whether it actually scrolled, and `_onUnreadPillJumpTap` only sets `_hasSeenFirstUnread` after a successful jump.

packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
```diff
@@ -1212,61 +1417,163 @@ class _StreamMessageListViewState extends State<StreamMessageListView> {
...
 
-      // Mark messages as read if needed.
-      if (widget.config.markReadWhenAtTheBottom) {
-        _maybeMarkMessagesAsRead().ignore();
-      }
+    const positionsEquality = UnorderedIterableEquality<ItemPosition>();
+    if (!positionsEquality.equals(itemPositions, _markUnreadViewportSnapshot)) {
+      _markUnreadViewportDiverged = true;
     }
   }
```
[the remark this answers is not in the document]
renefloor: Fixed in [0808f07](https://github.com/GetStream/stream-chat-flutter/commit/0808f07d6) — `_markUnreadViewportSnapshot` now stores visible item *indices* (`List<int>`) instead of full `ItemPosition`s, so a fractional leading/trailing-edge change from an unrelated relayout no longer latches divergence.

packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
```diff
@@ -1212,61 +1417,163 @@ class _StreamMessageListViewState extends State<StreamMessageListView> {
...
+
+    final visibleIndices = itemPositions.map((position) => position.index).toList();
+    if (visibleIndices.isEmpty) return false;
+
+    final isAnchorVisible = visibleIndices.contains(anchorItemIndex);
+    // Smaller item indices are newer/closer to the bottom. If even the
+    // newest visible item is older than the anchor, the anchor has scrolled
+    // off the bottom of the viewport — the user scrolled past it.
+    final isScrolledPast = visibleIndices.reduce(min) > anchorItemIndex;
+    if (!isAnchorVisible && !isScrolledPast) return false;
```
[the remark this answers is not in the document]
renefloor: Verified and I don't think this needs a code change: the index→message mapping in the `itemBuilder`/`separatorBuilder` above (`messages[i - 2]`) is applied unconditionally regardless of `widget.config.reverse` — item index 2 is always the newest message, whichever way `reverse` flips the *visual* scroll direction. So "every visible index > anchorIndex" reliably means "scrolled past toward older messages" in both layouts; flipping the comparison for `reverse: false` would actually break it. Same reasoning applies to the neighboring `isLastItemFullyVisible` check (fixed `lastItemIndex = 2`, edge-sign comparison) a few lines up, which has the same shape and isn't reverse-branched either. Happy to add a short comment noting this if it'd help future readers, but leaving the logic as-is.

packages/stream_chat_flutter/lib/src/message_list_view/mark_read_details.dart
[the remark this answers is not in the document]
renefloor: It was pushed to a different branch, it should be available now

packages/stream_chat_flutter/lib/src/message_list_view/unread_indicator_button.dart
```diff
@@ -29,21 +30,27 @@ import 'package:stream_core_flutter/chat.dart' as core;
...
-  /// Receives the ID of the last message the current user has read,
-  /// which can be used to scroll to that position.
-  final Future<void> Function(String? lastReadMessageId) onJumpTap;
+  /// This is the pre-existing unread boundary's count, captured when the
+  /// channel was opened — it does not change for the lifetime of the
+  /// session.
+  final int unreadCount;
+
+  /// Called when the jump-to-unread area is tapped.
+  final Future<void> Function() onJumpTap;
```
VelikovPetar: Wouldn't this be a breaking change? (`Function(String? lastReadMessageId)` -> `Function()`?
Also the mandatory `unreadCount` means that a customer directly using the `UnreadIndicatorButton` would fail to compile when updating the SDK (edge-case, but still possible IMO).

packages/stream_chat_flutter/lib/src/message_list_view/unread_messages_separator.dart
```diff
@@ -106,7 +106,7 @@ class UnreadMessagesSeparator extends StatelessWidget {
         child: Padding(
           padding: effectiveContentPadding,
           child: Text(
-            context.translations.unreadMessagesSeparatorText(),
```
VelikovPetar, line 109: Small concern here -> a customer overriding the `context.translations.unreadMessagesSeparatorText` would no longer see their custom text - is that OK? 

Also, should we maybe now deprecate/mark for removal the `unreadMessagesSeparatorText`? I believe it is no longer used.

packages/stream_chat_flutter_core/lib/src/stream_channel.dart
```diff
@@ -76,6 +78,21 @@ class StreamChannel extends StatefulWidget {
...
+  /// `false` to always open at the latest message instead — the message
+  /// list then surfaces pre-existing unread via its unread divider and
+  /// jump-to-unread pill rather than by scrolling there automatically.
+  ///
+  /// Has no effect on [StreamChannel.value], which never repositions the
+  /// loaded window.
+  ///
+  /// Only read once, during channel initialization — changing it after this
+  /// widget has mounted does not reposition the current viewport.
+  final bool openAtFirstUnread;
```
VelikovPetar, line 94: Maybe `openAtLastRead` would be a more fitting name for the flag? (because we are actually loading the channel at the `lastReadMessageId`/`lastRead` timestamp)

_I don't really have a strong preference here, just posting as a suggestion._

packages/stream_chat_flutter/lib/src/message_list_view/mark_read_details.dart
```diff
@@ -0,0 +1,44 @@
...
+  /// The channel's current unread count.
+  final int unreadCount;
+}
+
+/// Signature for overriding [StreamMessageListView]'s automatic mark-read
+/// gating.
+///
+/// Return `true` to mark the channel as read, `false` to skip it for now —
+/// the list retries on the next relevant scroll or message event.
+typedef StreamShouldMarkReadPredicate = bool Function(StreamMarkReadDetails details);
```
VelikovPetar: Should we maybe move this to `typedefs.dart`?

packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
```diff
@@ -392,6 +533,21 @@ class _StreamMessageListViewState extends State<StreamMessageListView> {
         final currentUser = streamChannel?.channel.client.state.currentUser;
         final isAtBottom = !_showScrollToBottom.value;
 
+        // The scroll-to-bottom badge and divider A's growing count only
+        // apply to the channel's own message stream (not thread replies),
+        // and never count the current user's own messages.
+        final isOwnMessage = message.user?.id == currentUser?.id;
+        if (!_isThreadConversation && !isOwnMessage) {
```
VelikovPetar: Does it make sense to use the `canCountAsUnread` extension here, which also covers the checks for messages which should not be considered in the unread count?

packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
VelikovPetar: Wouldn't this prevent incrementing the `_unreadDividerGrowth.value` when messages arrive during scrolling? Should we move the unread logic before this, or handle it in a separate subscription?

packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
```diff
@@ -1212,61 +1432,170 @@ class _StreamMessageListViewState extends State<StreamMessageListView> {
...
     final isInThread = widget.parentMessage != null;
 
-    // A server-side thread object only exists once the parent has at least
-    // one reply; markThreadRead on a reply-less parent returns 404.
-    if (isInThread && (widget.parentMessage?.replyCount ?? 0) == 0) return;
+    if (isInThread) {
+      // A server-side thread object only exists once the parent has at
+      // least one reply; markThreadRead on a reply-less parent returns 404.
+      if ((widget.parentMessage?.replyCount ?? 0) == 0) return;
+      if (!(channel.state?.isUpToDate ?? false)) return;
```
VelikovPetar: Do we need the `channel.state?.isUpToDate` check for threads? I believe now we would prevent marking threads as read, if the main channel is not up to date, which I don't think is intentional.

packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
```diff
@@ -1212,61 +1432,170 @@ class _StreamMessageListViewState extends State<StreamMessageListView> {
...
-  // 3. In a thread, the parent has at least one reply — the server-side
-  //    thread object doesn't exist until the first reply lands.
+  // In a thread: the parent must have at least one reply (the server-side
+  // thread object doesn't exist until the first reply lands), and the
+  // channel must be up to date.
   //
-  // If any of the conditions are not met, the function returns early.
-  // Otherwise, it calls the _markMessagesAsRead function to mark the messages
-  // as read.
+  // In the channel, mirrors iOS's `shouldMarkChannelRead` gating:
```
VelikovPetar: In general I would avoid referencing other platforms in docs - I think they can become very confusing, especially if the referenced platform's code changes.

packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
```diff
@@ -295,17 +294,150 @@ class _StreamMessageListViewState extends State<StreamMessageListView> {
...
+  // Reacts to a `currentUserReadStream` emission. An explicit mark-unread
+  // moves the read boundary backward — treat it as a new session start for
+  // divider A/the pill, mirroring iOS's `forceUpdate` path.
+  void _handleCurrentUserReadChanged() {
+    if (_isThreadConversation) return;
+
+    final channel = streamChannel?.channel;
+    if (channel == null) return;
+
+    if (channel.state?.isMarkedAsUnread ?? false) {
```
VelikovPetar: While testing, I noticed 2 strange UI behaviours, which I am not sure are intentional:

### Marking message as unread
1. Mark a message as unread
2. The pill appears
3. The slightest scroll dismisses the pill

https://github.com/user-attachments/assets/3d91061e-d92d-4a20-92fb-93fcc109c614

### Dismissing the unread pill
1. Mark a message as unread
2. The pill appears
3. Manually dismiss the the pill by clicking on it
4. Every next message arriving cause the pill to re-appear and disappear immediately 

https://github.com/user-attachments/assets/11ce3db5-b55a-45f4-b325-7f494583d8d0

packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
```diff
@@ -1212,61 +1432,170 @@ class _StreamMessageListViewState extends State<StreamMessageListView> {
...
+    final unreadCount = channel.state?.unreadCount ?? 0;
+    if (unreadCount <= 0) return;
+
+    final noPreexistingUnread = _unreadBaselineCaptured && _unreadBaseline == null;
+    // Equivalent to `channel.usesLocalUnreadCount`, spelled out via
+    // `channel.client` rather than `Channel`'s private client field so it
+    // stays evaluable against a test double that only implements the public
+    // API surface.
+    final usesLocalUnreadCount = channel.client.isLocalUnreadCountEnabled && !channel.canUseReadReceipts;
+    final hasSeenFirstUnreadMessage = noPreexistingUnread || _hasSeenFirstUnread.value || usesLocalUnreadCount;
```
VelikovPetar: I believe we have an edge case here:

`hasSeenFirstUnreadMessage` is only satisfiable here via `_hasSeenFirstUnread`, which requires divider A's anchor to have resolved - `_maybeUpdateHasSeenFirstUnread` bails while `anchorId` is
null (`message_list_view.dart:1489`). However there is a state where `unreadMessages > 0` and `lastReadMessageId == null`:

- **Never-read channels**: `unreadMessages > 0` with `lastReadMessageId == null` (a member who
  never opened the channel — the same state the `_minValidLastRead` sentinel in `StreamChannelState` exists for). `getFirstUnreadMessage` then falls back to "oldest regular message", which deliberately returns null until `_topPaginationEnded` (`stream_chat_flutter_core/lib/src/stream_channel.dart:769`, `:775`).

Steps to reproduce:
1. Have a never-opened channel with messages more than the first page
2. Open the channel
3. Tapping on the unread pill does nothing

https://github.com/user-attachments/assets/b740e294-b08e-417c-a837-7c07e9a1ab9e

packages/stream_chat_flutter/lib/src/message_list_view/stream_message_list_view_configuration.dart
```diff
@@ -36,13 +37,32 @@ class StreamMessageListViewConfiguration {
     this.keyboardDismissBehavior = .onDrag,
     this.scrollPhysics = const ClampingScrollPhysics(),
     this.autoScrollPolicy = .whenOwnMessageOrAtBottom,
+    this.shouldMarkRead,
```
VelikovPetar: Just thinking out loud: Is this something a customer would ever configure? I am thinking whether it is too complicated to be configurable... (of course, no harm in having it, but it is another public API to maintain)

packages/stream_chat_flutter/lib/src/message_list_view/unread_indicator_button.dart
renefloor: Good catch — you were right, and it was breaking twice over: the `onJumpTap` signature change *and* the new `required unreadCount`. Both reverted in 241908e9.

`onJumpTap` keeps `Future<void> Function(String? lastReadMessageId)`, and `unreadCount` is now an optional `int?`:

- omitted → the original behaviour, the widget subscribes to `currentUserReadStream` itself and hides while there's nothing unread;
- supplied → purely presentational, renders unconditionally with that count and never touches read state. That's the mode `StreamMessageListView` uses, so the pill can keep a count frozen at channel open.

In presentational mode `onJumpTap` receives `null`, since the host owns the boundary there — documented on the parameter.

Added `unread_indicator_button_test.dart` covering both modes, including that the host-driven mode never reads channel state at all. Changelog entry moved from 🔄 Changed to ✅ Added to match.

packages/stream_chat_flutter/lib/src/message_list_view/unread_messages_separator.dart
renefloor: Both points valid. Fixed in 241908e9.

`unreadMessagesSeparatorLabel` now has a default implementation that delegates to `unreadMessagesSeparatorText()`, added in both `Translations` and `GlobalStreamChatLocalizations`. So a class that overrides only the old string keeps compiling **and** keeps showing its custom text — the silent behaviour change is gone. The bundled locales override the label to show the count.

And yes, `unreadMessagesSeparatorText` is now `@Deprecated`. It isn't dead code though — it's the fallback the default label delegates to, so it stays functional until the next major.

packages/stream_chat_flutter_core/lib/src/stream_channel.dart
renefloor: Checked both sibling SDKs before deciding, and kept `openAtFirstUnread` — they consistently use "first unread" for this concept:

- **stream-chat-swiftui**: `firstUnreadMessageId` (`ChatChannelViewModel`, `ChatChannelDataSource`, `MessageListView`)
- **stream-chat-android**: `scrollToFirstUnreadMessage()` (`MessageListController` and both view models), `onScrollToFirstUnread` / `ScrollToFirstUnreadButtonParams` in Compose

Neither names it after last-read; they use `lastReadMessageId` only for the read-state field itself.

You're right that the implementation loads at `lastReadMessageId`/`lastRead`, so `openAtLastRead` describes the mechanism more literally. I went with the flag naming the intent plus matching the siblings' vocabulary, since that's what shows up in specs and support threads across platforms. Easy to switch if you'd rather it describe the mechanism.

packages/stream_chat_flutter/lib/src/message_list_view/mark_read_details.dart
renefloor: Moot now — `shouldMarkRead` was removed in 241908e9 (see the thread on `StreamMessageListViewConfiguration`), and this typedef existed only as its signature, so the whole `mark_read_details.dart` file went with it, `StreamMarkReadDetails` included.

To be explicit that I didn't over-read your comment: the class went because it was purely the predicate's parameter object — its own doc said "not intended to be constructed directly", and every reference to it traced back to `shouldMarkRead` (the config doc, the one construction site, three test locals). It was never in a released version, so no deprecation was needed.

If we reinstate the hook, the typedef goes in `typedefs.dart` as you suggested.

packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
renefloor: Right in intent, and adopted — but I couldn't call `MessageRules.canCountAsUnread` directly.

It reads `channel.usesLocalUnreadCount`, which is an extension getter (`ChannelCapabilityCheck`) reaching `Channel`'s private `_client` field. Extension members resolve statically, so it can't be stubbed on a `MockChannel`: it throws `NoSuchMethodError: Class 'MockChannel' has no instance getter '_client'` and broke 15 existing widget tests.

So 241908e9 adds `_countsTowardsUnreadIndicators`, the message- and sender-level half of that rule — silent, shadowed, ephemeral, thread-only, restricted, muted-sender and own messages. The channel-level half (`isMuted`, `canUseReadReceipts`, `usesLocalUnreadCount`) is deliberately left out and documented as such: those decide whether the *server* tracks an unread count for the channel, whereas the badge and divider are local "what arrived while you weren't looking" counters that should keep working either way.

Added six tests, each followed by a control message so they can't pass vacuously. Ephemeral, shadowed and muted-sender all failed before the fix; silent and thread-only turned out to be filtered upstream already.

If you'd rather have the real shared rule, the fix belongs on the `stream_chat` side — making `usesLocalUnreadCount` a normal getter, or having `canCountAsUnread` take the pieces it needs rather than the whole `Channel`. Happy to do that instead.

packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
renefloor: Confirmed, and worse than just the divider — the `isScrolling` guard sat above the counting, so anything landing mid-drag or mid-fling was dropped from **both** `_unreadDividerGrowth` and `_scrollToBottomBadge`, permanently.

Fixed in 241908e9 by moving the counting above the guard rather than adding a second subscription: the counting doesn't touch scrolling, so there was no reason for it to sit behind a scroll guard in the first place. The guard now only protects the auto-scroll decision below it, which is what it was for.

Regression test holds a drag open with `startGesture` so the underlying `ScrollPosition` genuinely reports `isScrolling == true`, delivers a message, and asserts the badge before releasing. It fails on the old ordering.

packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
renefloor: Agreed, that wasn't intentional. A thread read has nothing to do with where the parent channel's own loaded window sits, so gating on it silently blocked thread reads whenever the channel was scrolled back into history.

Removed in 241908e9 — the thread branch now only requires the parent to have at least one reply (the 404 guard). Test added that pumps a thread with `isUpToDate: false` and verifies `markThreadRead` still fires; the doc comment above the method now says a thread read is independent of the channel's window, so this doesn't get re-added by mistake.

packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
renefloor: Fair point — removed in 241908e9. Gone from `_maybeMarkMessagesAsRead` (the `shouldMarkChannelRead` reference and the "escape hatch" note), from `_handleCurrentUserReadChanged` (`forceUpdate`), from `ChannelClientState.isMarkedAsUnread` over in `stream_chat`, and from the changelog entry. The conditions are now described on their own terms.

I did keep the two WhatsApp references on the divider's growing count. Those describe product behaviour we're deliberately matching rather than another SDK's internals, so they can't go stale the way a code reference can — but say the word and they go too.

packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
renefloor: Both real, both fixed in 241908e9. The recordings made these easy to pin down — thanks.

**Pill dismissed by the slightest scroll.** A manual mark-unread restarts the pill's session, and its anchor is the message you were just looking at, so it's on screen from the outset — and merely *seeing* the anchor counted as reaching the boundary. `_maybeUpdateHasSeenFirstUnread` now requires `isScrolledPast` rather than `isAnchorVisible` for sessions that came from an explicit mark-unread, tracked with `_unreadFromManualMarkUnread`. Scrolling down past the boundary still retires the pill, which is the intended way out.

**Pill re-appearing and vanishing on every arrival.** The reset in `_handleCurrentUserReadChanged` ran on *every* read-stream emission while `isMarkedAsUnread` was set, and each new message re-emits it — so `_hasSeenFirstUnread` was cleared and immediately re-latched. It now runs only on the transition into that state, via `_wasMarkedAsUnread`.

Tests for both. The flicker one pumps frame by frame and asserts the pill is absent on each frame — worth noting an end-state assertion after `pumpAndSettle` passes even against the unfixed code, because it reappears and re-latches inside the same settle. Both fail against the old code.

packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
renefloor: Confirmed, and the analysis was spot on — thanks for tracing it to `_topPaginationEnded`.

Fixed in 241908e9: when there's no resolved anchor **and** no `lastReadMessageId`, `_onUnreadPillJumpTap` now scrolls to the oldest currently loaded message, which is as far back as the boundary can be and pulls the next page in on arrival, instead of returning early. `_hasSeenFirstUnread` is deliberately not latched there, so the pill stays up until the real boundary is actually reached. Test covers the tap landing on the oldest loaded message with the pill still shown afterwards.

**One thing I'd like your view on, since it's a behaviour call I left out of the commit.** The fallback makes the tap do something, but the underlying asymmetry is still there. The mark-read gate already exempts channels using local unread counts, because they have no server read state to anchor against — and a never-read channel has the same problem: no boundary exists until top pagination ends. Positioning can also miss the boundary with `openAtFirstUnread: true` in two other spots in `stream_channel.dart`: the >1000-member case where `loadChannelAtMessage` throws and we fall back, and the `_minValidLastRead` sentinel path that deliberately falls through to load-latest.

In all three the user lands at the bottom with `unreadCount > 0`, and auto mark-read is now blocked where it previously wasn't — since before this PR you'd have been positioned at the boundary and satisfied that condition by construction. Should never-read channels get the same exemption as local-unread ones? That would restore the old behaviour for the one population where the gating genuinely changed, and make this jump fallback a nicety rather than the only route out.

packages/stream_chat_flutter/lib/src/message_list_view/stream_message_list_view_configuration.dart
renefloor: Removed in 241908e9 — agreed, and "we can always add a config later, never remove one" won it.

Checked the siblings first in case there was an equivalent worth keeping parity with, and neither exposes one: stream-chat-swiftui's `canMarkRead` / `hasSetInitialCanMarkRead` are `private` view-model state, and stream-chat-android has only `readEventsEnabled`, which is the server channel config rather than a UI gating hook.

What settled it is that the tightened gating turns out not to be a regression for anyone on the defaults: with `openAtFirstUnread: true` you land on the boundary, so the new "has the boundary been seen" condition is satisfied at open — exactly as before, where the bottom wasn't visible at open either and you had to scroll down to mark read. It only bites under the new `openAtFirstUnread: false`, where blocking the mark-read is the entire point of the flag. So there's no one to hand an escape hatch to.

**One residual worth a doc note, though.** The pill's dismiss button is the user-facing way to clear a boundary you'll never scroll back to, and the pill is gated on `showUnreadIndicator`. So `showUnreadIndicator: false` combined with `openAtFirstUnread: false` — or with one of the fallback paths in the `hasSeenFirstUnreadMessage` thread — leaves a channel that won't auto-mark-read and has no pill to dismiss, the only way out being `markReadWhenAtTheBottom: false` and calling `markRead()` yourself. Happy to add a note on whichever of those flags you think is the right home for it, or to reinstate the predicate if you'd rather keep the seam after all.

packages/stream_chat_flutter/lib/src/localization/translations.dart
```diff
@@ -100,8 +100,22 @@ abstract class Translations {
...
+  ///
+  /// Defaults to the count-less `unreadMessagesSeparatorText` so that
+  /// implementations written before this method existed — including ones
+  /// that customise only that older string — keep rendering their own text
+  /// rather than silently reverting to the built-in copy. Override this to
+  /// show the count.
+  String unreadMessagesSeparatorLabel({required int count}) {
+    // ignore: deprecated_member_use_from_same_package
+    return unreadMessagesSeparatorText();
+  }
```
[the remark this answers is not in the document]
renefloor: That's true for every thing we add here. Integrators can choose to `extend` if they only want to change some strings, or they can choose to 'implement' if they want to make sure all strings are implemented for a language.

packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
renefloor: Test after the fix:

https://github.com/user-attachments/assets/8645ef7a-1449-4e52-975e-ab144de7f775

https://github.com/user-attachments/assets/d8a30b51-3281-4234-a345-1ea95899f4b5

packages/stream_chat_flutter_core/lib/src/stream_channel.dart
VelikovPetar: Makes sense, thanks for the explanation!

packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
VelikovPetar: OK sounds like a good compromise. I am just thinking whether silent messages should really be excluded -> those should also be visible messages in the channel (as opposed to shadowed ones).

packages/stream_chat_flutter/lib/src/message_list_view/stream_message_list_view_configuration.dart
VelikovPetar: Probably we can add the doc on the `showUnreadIndicator` - I think it makes more sense.

packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
VelikovPetar: I think we should be fine if never-read channels get the same exception as local-unread ones. It is a relatively rare case, so I think it a good approach.

packages/stream_chat_flutter/lib/src/message_list_view/message_list_view.dart
renefloor: Checked both siblings, and they agree on excluding silent messages from unread counts.

**iOS** documents it on the model itself:

> Silent messages are special messages that don't increase the unread messages count nor mark a channel as unread.

— `ChatMessage.isSilent`. It's then enforced at the increment site: `ChannelReadUpdaterMiddleware.unreadCountUpdateSkippingReason` returns `.messageIsSilent`, in the same list as own / muted-author / thread-reply / shadowed.

**Android** does the same in `ChannelStateLogic.updateCurrentUserRead` — an explicit `// Skip update for silent messages` branch, again alongside the muted / own / shadowed / thread-reply skips.

Since your question is really about the *visible* indicator rather than the stored count, I checked whether either platform applies a looser rule there — neither does. iOS sets `jumpToUnreadMessagesButton.content` from `channel.unreadCount`, and Android's `UnreadLabelCalculator` reads `channelUserRead.unreadMessages`, so both indicators inherit the exclusion from the count itself.

So I'd keep excluding it. You're right that silent messages are visible in the channel where shadowed ones aren't, but the two properties are independent: "silent" means *doesn't notify, doesn't bump unread*, not *hidden*. And our own `MessageRules.canCountAsUnread` excludes silent as well, so counting it in the badge and divider would make the divider read "3 unread messages" while `channel.state.unreadCount` in the same view says 2.

One correction to what I wrote earlier in this thread: I said silent and thread-only "turned out to be filtered upstream already". That was wrong for silent — `newMessageStream` yields whatever lands at the tail, silent included, so the check in `_countsTowardsUnreadIndicators` is load-bearing rather than belt-and-braces.

VelikovPetar: Sounds good to me, let's not complicate things!

VelikovPetar (approved):

PR #2873, opened 2026-08-07
feat(core): dynamic search debounce based on query length

xsahil03x:
### 🎯 Goal

Short search queries (1–2 characters) are low-selectivity and expensive on the backend, causing high latency and timeouts on the `search` / `queryUsers` / `queryMembers` endpoints. This adapts the debounce interval to the query length: short queries wait longer before hitting the backend; longer, more selective queries use the standard delay.

Linear: [FLU-684](https://linear.app/stream/issue/FLU-684/dynamic-search-debounce-based-on-query-length)

### 📝 What changed

- **`stream_chat_flutter_core`** — added query-length-aware debounced search to `StreamUserListController`, `StreamMemberListController`, and `StreamMessageSearchListController`:
  - `search(String query)` — the primary entry point. Users match on name + id, members on name, messages on text (all as autocomplete filters). A blank query restores the base `filter` for the user/member lists, and clears the results for message search (the backend rejects empty message searches).
  - `searchWithFilter(Filter filter)` — search with an explicit `Filter`, for callers that need to match on other fields, or to keep the base `filter` applied while searching.
  - `clearResults()` — cancels any pending search and clears the current results (e.g. when the input is emptied).
  - Debounce adapts to query length — **500ms for ≤2 chars, 300ms for 3+**. The policy is internal and unexported; it is not part of the public API.
  - Results from **superseded** queries are dropped, so a slower, older query can't overwrite a newer one.
  - Reuses `rate_limiter`'s `debounce` (two instances picked by length); no hand-rolled `Timer`.
- **`stream_chat`** — exported `FilterOperator` alongside `Filter` (needed to inspect a filter's operator when choosing the debounce delay).
- **`rate_limiter`** — bumped to `^1.1.0` (via `melos.yaml`), which reads the wall clock through `package:clock` instead of `DateTime.now()`. That makes the debounce controllable by `FakeAsync`, so the tests below run in virtual time. It also applies to the debounces that already existed (the composer, mark-read, persistence), which is why `message_composer_url_enrichment_test.dart` changed — see Testing.
- **`sample_app`** — replaced the four hand-rolled `Timer`-debounced search sites with the controller's debounce. Channel-list message search uses `search()`; the three user-directory sites (add-members, new chat, new group) use `searchWithFilter()` because they must keep their exclusion filter applied while searching, and they match on name or id to mirror what `search()` builds.

#### `search()` replaces the base filter, it does not narrow it

`search(query)` sets the autocomplete filter as the active filter rather than ANDing it with the constructor's `filter`; a blank query restores the base filter. `searchWithFilter()` behaves the same way, so the two are consistent, and it is the way to compose a search with the base scope:

```dart
controller.searchWithFilter(Filter.and([
  Filter.notIn('id', excludedIds),
  if (query.isNotEmpty) Filter.autoComplete('name', query),
]));
```

Merging implicitly was tried first and dropped, because it is wrong in two ways. A base filter that already carries a text operator contradicts the search (`autoComplete('name', 'jo')` AND `autoComplete('name', 'abc')` matches nothing), and its text also skews the debounce itself — `searchQueryLength()` takes the longest text in a compound filter, so a base `autoComplete('name', 'jonathan')` would make a 1-char query use the 300ms delay instead of 500ms. This also matches iOS, where `search(term:)` builds the query and `search(query:)` takes the caller's.

### ✅ Non-breaking

Purely additive — new methods and one new export only. No existing signatures, defaults, or exports changed. `doInitialLoad` gains a "newest-wins" guard (fixes an overlapping-load race; single-load behavior is unchanged).

### 🚫 Out of scope (follow-ups)

- Channel-*name* search (`StreamChannelListController`'s filter is constructor-only; not wired as text search in Flutter).
- Mention autocomplete (separate UI surface, already 300ms-debounced).
- Cancelling in-flight HTTP requests.

### 🧪 Testing

New: `search_debouncer_test.dart` + controller tests for user/member/message search (debounce timing, length policy, superseded-result guard, blank-query handling, `searchWithFilter`, `clearResults`).

All debounce tests run under `FakeAsync`, so they assert the real 500ms/300ms policy at its exact boundaries with no wall-clock waits:

```dart
controller.search('ab');                              // 2 chars → short query
async.elapse(const Duration(milliseconds: 499));
expect(queryCount, 0);
async.elapse(const Duration(milliseconds: 1));
expect(queryCount, 1);                                // 500ms, not 300ms
```

`message_composer_url_enrichment_test.dart` moves off `tester.runAsync` for the same reason. It drove the composer's 350ms enrichment debounce with a real 500ms sleep, which advances real time but not the clock `flutter_test` installs — so under `^1.1.0` the debounce rescheduled itself and `enrichUrl` was never called. It now advances the fake clock with `tester.pump`, which also drops 8 real sleeps from the suite.

Full `stream_chat_flutter_core` suite: **361 passing**. Analyze clean, formatted.

## Summary by CodeRabbit

* **New Features**
  * Added debounced text search and filter-based search for member, user, and message lists.
  * Added result clearing and cancellation support.
  * Exported `FilterOperator` for public use.
* **Bug Fixes**
  * Prevented outdated search responses and errors from replacing newer results.
  * Improved blank-search, pagination, and pending-search handling.
* **Sample App**
  * Simplified search across chat creation, member selection, and channel search.
* **Documentation**
  * Updated changelogs and message-search documentation.


packages/stream_chat_flutter_core/lib/src/search_debouncer.dart
```diff
@@ -0,0 +1,50 @@
...
+/// backend; longer, more selective queries use the standard delay.
+///
+/// Backed by two [Debounce] functions — one per delay — because a [Debounce]'s
+/// wait is fixed at construction. Each call runs the debouncer for the matching
+/// query length and cancels the other so only a single search is scheduled.
+@internal
+class SearchDebouncer {
+  /// Creates a [SearchDebouncer] that runs [onSearch] after the
+  /// length-dependent delay elapses.
+  SearchDebouncer(Function onSearch)
```
VelikovPetar: Do you think it would make sense to make the `_shortQueryMaxLength`/`_shortQueryDelay`/`_defaultDelay` configurable - in case we need to fine-tune them per use-case (ex. message search vs user search)? 

_Just a suggestion, I am aware that this will most likely not change in foreseeable future._

packages/stream_chat_flutter_core/lib/src/search_debounce_mixin.dart
```diff
@@ -0,0 +1,81 @@
...
+  /// Cancels any pending or in-flight search and clears the current results.
+  ///
+  /// Consider calling this when the search input is emptied, so results for an
+  /// abandoned query are neither shown nor repopulated by a late response.
+  void clearResults() {
+    _searchDebouncer.cancel();
+    // Bump the generation so an already-running load is treated as superseded
+    // and cannot repopulate the cleared results when its response arrives.
+    _loadGeneration += 1;
+    value = PagedValue<Key, Value>.loading();
```
VelikovPetar: Is setting `loading()` here correct? `clearResults` doesn't mean that the controller will keep loading - perhaps we should reset to an empty list instead?

packages/stream_chat_flutter_core/lib/src/stream_member_list_controller.dart
```diff
@@ -92,6 +97,19 @@ class StreamMemberListController extends PagedValueNotifier<int, Member> {
...
 
+  /// Searches members matching [query], debounced by the query length.
+  ///
+  /// [query] selects the debounce delay by its length — short, low-selectivity
+  /// queries wait longer before hitting the backend to reduce load, longer
+  /// queries use the standard delay. When [filter] is provided it becomes the
+  /// active filter for the reload; otherwise the current filter is reused.
+  /// Rapidly superseded searches are dropped, so only the latest query's
+  /// results are applied.
+  void search(String query, {Filter? filter}) {
```
VelikovPetar: I find this API contract a bit strange: Basically the consumer **should** make sure the `query` and `filter` are in sync right? 
Otherwise, they might end up with unexpected results - especially if paginating data. 

Also, if a `filter` is not provided, the `search` will use the current filter value, potentially ignoring the `query`?

I am thinking if we should maybe split the search methods: 
- One which does a plain `query` search -> using a filter built internally from the `query` term (ex: .`autocomplete('name', 'query`)`). The newly built filter becomes the active filter.
- One which does a `Filter` search - setting the passed `filter` as the current filter. 

The way I am seeing this whole thing is like the following: We need to debounce calls to the `doInitialLoad` - so that rapid changes in the `filter` are not all forwarded to the server. Basically to make sure we do not fire too many different initial queries. Pagination in the end should not be affected - because once we have the initial page from `doInitialLoad` - we can freely paginate, without worrying about the debouncing.

packages/stream_chat_flutter_core/lib/src/stream_user_list_controller.dart
```diff
@@ -100,6 +105,19 @@ class StreamUserListController extends PagedValueNotifier<int, User> {
...
 
+  /// Searches users matching [query], debounced by the query length.
+  ///
+  /// [query] selects the debounce delay by its length — short, low-selectivity
+  /// queries wait longer before hitting the backend to reduce load, longer
+  /// queries use the standard delay. When [filter] is provided it becomes the
+  /// active filter for the reload; otherwise the current filter is reused.
+  /// Rapidly superseded searches are dropped, so only the latest query's
+  /// results are applied.
+  void search(String query, {Filter? filter}) {
```
VelikovPetar: Same concerns as the `StreamMemberListController`.

packages/stream_chat_flutter_core/lib/src/search_debouncer.dart
xsahil03x: Good call. I pulled the timing into an internal `SearchDebouncePolicy` object (instead of loose constants), injected per controller via a `@visibleForTesting` param — which also lets the tests run with a zero delay deterministically. Kept it non-public for now; if a concrete per-use-case need shows up we can promote it to a public option in a follow-up.

packages/stream_chat_flutter_core/lib/src/search_debounce_mixin.dart
xsahil03x: Good catch — `clearResults()` now clears to an empty list instead of `loading()`. Separately, `debouncedSearch` moves to a loading state while a search is actually running, so starting a search from the cleared/empty list shows a loading indicator rather than briefly rendering "no results".

packages/stream_chat_flutter_core/lib/src/stream_member_list_controller.dart
xsahil03x: Agreed, the `{Filter?}` contract was awkward. Split it into two methods:

- `search(String query)` builds the name-autocomplete filter internally and merges it with the controller's base `filter`, so the query and filter can't drift out of sync.
- `searchWithFilter(Filter filter)` sets an explicit filter as the active filter.

Both debounce the reload by the search text's length (for `searchWithFilter`, extracted from the filter; a filter with no search text reloads immediately). Pagination via `loadMore` is never debounced — it just extends the current search, as you noted.

packages/stream_chat_flutter_core/lib/src/stream_user_list_controller.dart
xsahil03x: Same split here — `search(String query)` (autocompletes on name and id, merged with the base `filter`) plus `searchWithFilter(Filter filter)`. Details in the `StreamMemberListController` thread.

packages/stream_chat_flutter_core/lib/src/search_debounce_mixin.dart
```diff
@@ -0,0 +1,81 @@
...
+/// }
+/// ```
+mixin SearchDebounceMixin<Key, Value> on PagedValueNotifier<Key, Value> {
+  late final _searchDebouncer = SearchDebouncer(doInitialLoad);
+
+  int _loadGeneration = 0;
+
+  /// Schedules a debounced reload whose delay adapts to [queryLength].
+  @internal
+  void debouncedSearch(int queryLength) => _searchDebouncer(queryLength);
```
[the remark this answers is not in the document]
xsahil03x: Done — `debouncedSearch` now bumps the load generation before scheduling, so any in-flight load is treated as stale via `isStale` and its response is discarded. `doInitialLoad` and `loadMore` both guard on the generation before applying results.

packages/stream_chat_flutter_core/lib/src/search_debounce_mixin.dart
```diff
@@ -0,0 +1,81 @@
...
+  ///
+  /// Consider calling this when the search input is emptied, so results for an
+  /// abandoned query are neither shown nor repopulated by a late response.
+  void clearResults() {
+    _searchDebouncer.cancel();
+    // Bump the generation so an already-running load is treated as superseded
+    // and cannot repopulate the cleared results when its response arrives.
+    _loadGeneration += 1;
+    value = PagedValue<Key, Value>.loading();
+  }
```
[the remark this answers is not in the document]
xsahil03x: Done — the `debouncedSearch` doc now describes the generation bump and why an in-flight load is superseded, rather than the previous wording.

packages/stream_chat_flutter_core/lib/src/search_debounce_mixin.dart
```diff
@@ -0,0 +1,81 @@
...
+  /// Whether the load identified by [generation] has been superseded by a
+  /// newer one, in which case its result should be discarded.
+  @internal
+  bool isStale(int generation) => generation != _loadGeneration;
+
+  @override
+  void dispose() {
+    _searchDebouncer.cancel();
+    super.dispose();
+  }
```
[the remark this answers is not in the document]
xsahil03x: Done — added a doc comment to `dispose`.

packages/stream_chat_flutter_core/pubspec.yaml
```diff
@@ -30,6 +30,7 @@ dependencies:
   freezed_annotation: ^3.0.0
   meta: ^1.9.1
   package_info_plus: ">=9.0.1 <11.0.0"
+  rate_limiter: ^1.0.0
```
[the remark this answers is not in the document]
xsahil03x: `rate_limiter` is already managed centrally in `melos.yaml` (`command.bootstrap.dependencies`, line 93). The constraint here is the synced output of that, so nothing to change.

sample_app/lib/widgets/channel_list.dart
```diff
@@ -34,19 +34,11 @@ class _ChannelList extends State<ChannelList> {
...
-        setState(() {
-          _isSearchActive = _controller.text.isNotEmpty;
-        });
-        if (_isSearchActive) _messageSearchListController.doInitialLoad();
-      }
-    });
+    final query = _controller.text;
+    setState(() => _isSearchActive = query.isNotEmpty);
+    if (query.isEmpty) return _messageSearchListController.clearResults();
+    return _messageSearchListController.search(query);
```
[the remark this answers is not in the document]
xsahil03x: Done — `_messageSearchListController.dispose()` is now called alongside the other controllers, and the empty-query path calls `clearResults()`.

sample_app/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -718,6 +623,20 @@
 			defaultConfigurationName = Release;
 		};
 /* End XCConfigurationList section */
+
+/* Begin XCLocalSwiftPackageReference section */
+		781AD8BC2B33823900A9FFBB /* XCLocalSwiftPackageReference "FlutterGeneratedPluginSwiftPackage" */ = {
+			isa = XCLocalSwiftPackageReference;
+			relativePath = Flutter/ephemeral/Packages/FlutterGeneratedPluginSwiftPackage;
```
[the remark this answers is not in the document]
xsahil03x: This `project.pbxproj` change was unrelated to this PR and has been reverted — the file now matches `master`.

packages/stream_chat_flutter_core/lib/src/stream_member_list_controller.dart
```diff
@@ -92,6 +102,34 @@ class StreamMemberListController extends PagedValueNotifier<int, Member> {
...
+  /// base [filter]. Rapidly superseded searches are dropped, so only the latest
+  /// query's results are applied.
+  ///
+  /// To search on other fields, consider [searchWithFilter].
+  void search(String query) {
+    final trimmed = query.trim();
+    if (trimmed.isEmpty) return searchWithFilter(filter);
+
+    final searchFilter = Filter.autoComplete('name', trimmed);
+    searchWithFilter(filter.merge(searchFilter));
```
VelikovPetar: Small question: here we merge `Filter.autoComplete('name', trimmed)` with the initial `filter`, but in the `void searchWithFilter(Filter? filter)` - we directly search using the passed `filter` -> is this discrepancy intentional?

Edit: And a bigger question: Is it really expected to merge the filters here? I am concerned if the end result is submitting a non-optimised and slow filter to the backend.

_-Same for the UserController-_

packages/stream_chat_flutter_core/lib/src/search_debouncer.dart
VelikovPetar: Looks good! 
I just same a small concern about exposing the `SearchDebouncePolicy` on the controllers: Even though marked as `internal`, I believe customers could still pass a value for it if they opt-in? 
I am thinking if we should keep it fully private (for our internal customisation only) - and fully remove the option from the public API.

VelikovPetar (approved):

PR #2874, opened 2026-08-07
fix(llc, ui): Issues found by the e2e suites

testableapple:
## 🎯 Goal

Split out of #2869 — these are the SDK fixes the e2e suites uncovered, with their unit tests and changelog entries. #2869 keeps only the e2e suites and CI.

**Based on `master`** and reviewable on its own. #2869 picks these up by merging `master` once this lands; until then the e2e tests covering them fail there by design.

## 🛠 Implementation details

🟢 [FLU-668](https://linear.app/stream/issue/FLU-668) — `ChannelClientState.updateThreadInfo` no longer merges the parent message into a thread's reply list. A backend that returns the root alongside the replies made `StreamMessageListView` render it twice, since it also renders `parentMessage` in its own slot. This aligns the online path with the offline one, which already filtered on `parent_id`.

🟢 [FLU-669](https://linear.app/stream/issue/FLU-669) — the thread-replies footer under a message in the channel hardcoded `'$replyCount replies'`, so a single reply read **"1 replies"** and the label was never localized. It now uses `threadReplyCountText`, which is translated in 11 locales and is what the two sibling call sites already used.

🟢 [FLU-686](https://linear.app/stream/issue/FLU-686) — surfaced while reviewing the truncated-channel timestamp (FLU-685), not by a test: a channel-list row could preview **another channel's** last message. Rows are unkeyed, so a reorder rebinds the same `State` to a different channel, and the preserved last-known message that #2774 added (to keep the preview alive across a channel-state reload) was still used as a fallback — so an empty, not-yet-up-to-date channel showed the previous one's message. The resolver now drops the cache when it is rebound to a different `ChannelClientState`. Shipped in 10.2.0, so it is fixed here rather than introduced here.

> **Review follow-up:** the truncated-channel timestamp ([FLU-685](https://linear.app/stream/issue/FLU-685)) is deliberately **not** fixed in this PR. The UI fix (deriving the date from the previewed message) was pulled after review as too fragile, and the LLC route — applying the `channel.truncated` event's channel object, tried in `f25298a0b` and reverted in `253b97306` — can't clear `last_message_at` when a truncation carries no system message (`ChannelModel.copyWith`/`merge` null-coalesce every field), so it needs its own design. Tracked in [FLU-687](https://linear.app/stream/issue/FLU-687).

## 🧪 Testing

Unit tests moved across with the fixes:

- `channel_test.dart` — `.getReplies` keeps the parent message out of the thread. **383 tests green** in `stream_chat`.
- `stream_message_item_test.dart` (new) — the thread-reply label is read from an **injected** `Translations` returning sentinels, so both the singular and the plural case prove the widget goes through the translations table. A hardcoded label passes a plural-only assertion, which is why the default-English variant of this test was not enough. A separate plain `test()` pins `DefaultTranslations`' own pluralization.
- `stream_channel_list_item_test.dart` — the rebind regression for FLU-686, which only catches the bug because the replacement channel is stubbed `isUpToDate == false`.

**19 tests green** in the two `stream_chat_flutter` files. Each fix was mutation-checked — reverting the production line fails the matching test.

The e2e tests that cover these live in #2869 and **fail there until this lands and `master` is merged in**; that is expected and called out in that PR. Nothing here depends on #2869.

## ☑️ Contributor Checklist

- [x] Changelog entries added (`stream_chat`, `stream_chat_flutter`)
- [x] Unit tests added
- [x] `dart analyze --fatal-infos` and `dart format` clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented thread parent messages from appearing twice when loading replies.
  * Localized and pluralized thread-reply counts.
  * Fixed stale channel previews after channel changes or rebinding.
  * Updated channel timestamps correctly after truncation.
  * Ensured dates reflect the latest previewable message and are cleared when no valid message exists.
  * Cleared outdated preview messages when filtering rules change.


packages/stream_chat/lib/src/client/channel.dart
```diff
@@ -3927,7 +3927,7 @@ class ChannelClientState {
     final threadMessages = updatedThreads[parentId] ?? <Message>[];
     final updatedThreadMessages = _mergeMessagesIntoExisting(
       existing: threadMessages,
-      toMerge: messages,
+      toMerge: messages.where((it) => it.id != parentId),
     );
```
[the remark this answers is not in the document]
testableapple: I don't think this one can happen, so leaving `existing` unfiltered.

Thread membership on disk is the message's own `parent_id` column, not the in-memory map key. `updateChannelThreads` flattens `threads.values` and saves the messages unchanged, discarding the key (`chat_persistence_client.dart:375`) — so writing a parent into `threads[parentId]` pre-fix only ever upserted that message row with its own `parent_id`, which for a root is `null`.

On the way back in, hydration goes through `getChannelThreads` (`channel.dart:2562`) → `messageDao.getThreadMessages(cid)`, whose query filters `messages.parentId.isNotNull()` (`message_dao.dart:202`), and the result is grouped by each row's own `message.parentId!` (`stream_chat_persistence_client.dart:279`). A message can therefore only land in `threads[X]` if its own `parent_id == X`, and a root can't be its own parent.

So a pre-fix cache can't reintroduce the parent into its own thread, and filtering `threadMessages` would be dead code. Happy to reconsider if you can point at a write path that persists the map key as `parent_id`.

packages/stream_chat_flutter/lib/src/scroll_view/channel_scroll_view/stream_channel_list_item.dart
```diff
@@ -490,8 +490,13 @@ class _ChannelListDeliveryStatus extends StatelessWidget {
...
 
-/// A widget that displays the channel last message date.
-class ChannelLastMessageDate extends StatelessWidget {
+/// A widget that displays the date of the message shown in the channel
+/// preview.
+///
+/// Reads the same message the preview subtitle does, so the two never disagree
+/// and nothing is displayed when there is no message to preview. Deliberately
+/// not [Channel.lastMessageAt], which survives a truncated channel.
+class ChannelLastMessageDate extends StatefulWidget {
```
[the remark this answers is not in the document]
testableapple: Check it out one more time.

packages/stream_chat_flutter/lib/src/scroll_view/channel_scroll_view/stream_channel_list_item.dart
```diff
@@ -490,8 +490,15 @@ class _ChannelListDeliveryStatus extends StatelessWidget {
...
-class ChannelLastMessageDate extends StatelessWidget {
+/// A widget that displays the date of a channel's latest previewable sent
+/// message.
+///
+/// Displays nothing when there is none. Drafts are ignored — they have no sent
+/// date — so this is not always the message the preview subtitle shows: with a
+/// draft the subtitle previews the draft while this still dates the last sent
+/// message. Deliberately not [Channel.lastMessageAt], which survives a
+/// truncated channel.
+class ChannelLastMessageDate extends StatefulWidget {
```
VelikovPetar: I am afraid that this might be a potential breaking change: Changing the base class from `StatelessWidget` to a `StatefulWidget`. A customer subclassing `ChannelLastMessageDate` would cause a silent failure (their `build` method will no longer be called automatically) - a pretty extreme edge case, but still possible in theory.

packages/stream_chat_flutter/lib/src/scroll_view/channel_scroll_view/stream_channel_list_item.dart
```diff
@@ -512,16 +519,33 @@ class ChannelLastMessageDate extends StatelessWidget {
...
   Widget build(BuildContext context) {
-    return BetterStreamBuilder<DateTime>(
-      stream: channel.lastMessageAtStream,
-      initialData: channel.lastMessageAt,
-      builder: (context, lastMessageAt) => StreamTimestamp(
-        date: lastMessageAt.toLocal(),
-        style: textStyle,
-        formatter: formatter,
-      ),
+    final channelState = widget.channel.state;
```
VelikovPetar, line 553: I am curious: Wouldn't it be possible to fix this on LLC level? From what I understand, we are not (fully) handling the `channel.truncated` event -> we should probably make sure all the affected `channel` fields are updated properly when handling the event.

packages/stream_chat_flutter/lib/src/scroll_view/channel_scroll_view/stream_channel_list_item.dart
testableapple: Good catch, reverted it and updated, it's a StatelessWidget again and just delegates to a private stateful child that holds the cache. Subclassing works as before.

packages/stream_chat_flutter/lib/src/scroll_view/channel_scroll_view/stream_channel_list_item.dart
testableapple: Not sure we can fix it fully on the LLC level, but i did some improvements, could you please check it out one more time?

```
The handler now applies event.channel like channel.updated does, but copyWith/merge can't null a field, so a truncation without a system message still leaves lastMessageAt stale — hence the UI still deriving the date from the message list.
```

VelikovPetar: Let me have a look, I will also check what the server does in this exact case.

packages/stream_chat_flutter/lib/src/scroll_view/channel_scroll_view/stream_channel_list_item.dart
renefloor: In theory you're right, but I would not expect anybody to subclass the `ChannelLastMessageDate`. Subclassing widgets already basically never happens, but I also don't see anything specifically in `ChannelLastMessageDate` that makes it worth subclassing. I would not consider that as a breaking change, although perhaps in theory it is one.

Edit: btw, your solution with having a private stateful child is also fine, it's quite clean code.

packages/stream_chat_flutter/lib/src/scroll_view/channel_scroll_view/stream_channel_list_item.dart
testableapple: I reverted the fix on LLC level as discussed internally.

VelikovPetar (approved):

xsahil03x (approved):

PR #2908, opened 2026-08-20
refactor unread into controller

renefloor:
# Submit a pull request
[FLU-672](https://linear.app/stream/issue/FLU-672) 

## CLA

This is a follow up on #2871

That PR already has reviews, but unread logic was getting spread on a lot of places. Here we try to make the MLV clean again by extracting all the unread logic.

### DIFF
--- a/packages/stream_chat_flutter/lib/src/message_list_view/message_list_unread_controller.dart
+++ b/packages/stream_chat_flutter/lib/src/message_list_view/message_list_unread_controller.dart
@@ -225,15 +225,15 @@ class MessageListUnreadController {
   /// meaning [resolveDividerAnchor] is worth retrying once layout settles.
   bool get needsAnchorResolution => _unreadBaseline != null && _unreadDivider.value.anchorId == null;
 
-  /// Debounced channel mark-read.
-  late final debouncedMarkRead = debounce(
+  // Debounced channel mark-read.
+  late final _debouncedMarkRead = debounce(
     ([String? id]) => _channel()?.markRead(messageId: id),
     const Duration(seconds: 1),
     leading: true,
   );
 
-  /// Debounced thread mark-read.
-  late final debouncedMarkThreadRead = debounce(
+  // Debounced thread mark-read.
+  late final _debouncedMarkThreadRead = debounce(
     (String parentId) => _channel()?.markThreadRead(parentId),
     const Duration(seconds: 1),
     leading: true,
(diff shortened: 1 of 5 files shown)


packages/stream_chat_flutter/lib/src/message_list_view/message_list_unread_controller.dart
```diff
@@ -0,0 +1,735 @@
...
+    if (!isAtBottom) _scrollToBottomBadge.value += 1;
+  }
+
+  /// Processes an item-positions tick, with [isAtBottom] reporting whether
+  /// the newest message is fully visible.
+  void handleItemPositionsChanged(
+    Iterable<ItemPosition> itemPositions, {
+    required bool isAtBottom,
+  }) {
+    _hasLaidOut.value = true;
```
VelikovPetar, line 411: Should we maybe have the `if (itemPositions.isEmpty) return;` guard here as well? We currently only have this guard before calling `handleItemPositionsChanged` in the `message_list_view`, but maybe we could introduce it here as an additional safeguard in case we introduce a new scenario where this method is called.

packages/stream_chat_flutter/lib/src/message_list_view/message_list_unread_controller.dart
```diff
@@ -0,0 +1,735 @@
...
+
+  /// Messages that arrived while the user was scrolled away from the bottom.
+  ValueListenable<int> get scrollToBottomBadge => _scrollToBottomBadge;
+
+  /// Whether a baseline was captured but its anchor still needs resolving,
+  /// meaning [resolveDividerAnchor] is worth retrying once layout settles.
+  bool get needsAnchorResolution => _unreadBaseline != null && _unreadDivider.value.anchorId == null;
+
+  /// Debounced channel mark-read.
+  late final debouncedMarkRead = debounce(
```
VelikovPetar: I think this can be private? (also `debouncedMarkThreadRead`)

VelikovPetar (approved):

packages/stream_chat_flutter/lib/src/message_list_view/message_list_unread_controller.dart
renefloor: fixed: https://github.com/GetStream/stream-chat-flutter/pull/2908/commits/1092b9de59aff4804b8844a32874457e5029be38

packages/stream_chat_flutter/lib/src/message_list_view/message_list_unread_controller.dart
renefloor: fixed: https://github.com/GetStream/stream-chat-flutter/pull/2908/commits/1092b9de59aff4804b8844a32874457e5029be38

PR #2905, opened 2026-08-20
test(llc): expand channel event handling coverage

VelikovPetar:
# Submit a pull request

Linear: FLU-723

Github Issue: #

## CLA

- [x] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required).
- [x] The code changes follow best practices
- [x] Code changes are tested (add some information if not applicable)

## Description of the pull request

Test-only change: expands WebSocket event handling coverage in `channel_test.dart` (+43 tests, `stream_chat` suite goes from 1624 to 1667 tests). No production code is touched.

New coverage inside the `WS events` group:

- **Typing events** (new group): `typing.start` / `typing.stop` from other users, current-user events ignored, missing-user guards.
- **Message new**: `notification.message_new` adds the message and counts unread; channel messages are not appended while the channel is not up to date (thread-only replies still are); missing-message guard.
- **Reaction events** (new group): `reaction.new` / `reaction.updated` / `reaction.deleted` updating `ownReactions` on channel and thread messages.
- **Poll events** (new group): create/update/close, vote and answer cast/change/remove, own-vote preservation, unknown-poll and missing-payload guards.
- **Message deleted**: in-window hard delete removes the message; `deletedForMe` propagation.
- **Channel updated events** (new group): event channel merged into the model; member list replaced from the event.
- **Channel truncated events** (new group): `channel.truncated` clears messages and wipes persistence; `notification.channel_truncated` keeps the event's system message.
- **Member events**: `member.added` / `member.removed` / `member.updated`, non-member user events ignored, and `user.banned` / `user.unbanned` member refresh (including the app-level ban without cid guard).
- **Watching events**: `user.watching.stop` without a watcher count preserves the count.
- **Missing-payload guards** for draft, reminder, and channel push preference events.

No CHANGELOG entry: test-only, no observable behavior change.

## Screenshots / Videos

No UI changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Tests**
  * Expanded coverage for channel events, including typing indicators, messages, reactions, polls, membership, drafts, reminders, and push preferences.
  * Added validation for unread counts, watcher and member updates, thread synchronization, persistence cleanup, and delivery behavior.
  * Added safeguards against phantom messages or threads and verified preservation of cached reactions and polls.

PR #2911, opened 2026-08-20
refactor(llc): extract channel event handling into handler and state mutations

VelikovPetar:
# Submit a pull request

Linear: FLU-723

Github Issue: #

## CLA

> Stacked on #2905 (`test/channel-event-handling-coverage`) — merge that one first, then retarget this PR to `master`.

Moves the channel event handling out of `ChannelClientState` into two new internal classes, with no public API or behavior changes:

- **`ChannelEventHandler`** — validates and routes each WS event from a single subscription (replacing the 36 per-event subscriptions), preserving the original per-subscription execution order via a three-block dispatch. Also owns the side effects an event triggers outside the channel state: member refresh on ban/unban, persisted-message cleanup on truncation, and delivery reconciliation.
- **`ChannelStateMutations`** — owns the state writes, one semantic method per event (`onMemberRemoved`, `onPollVoteCasted`, …). The few writes that previously went through private state (typing events, watcher removal, member refresh, user message deletion) stay private on `ChannelClientState` and are injected as tear-offs, so the state class gains no new members.

### Why these classes exist

Previously, every event listener interleaved three unrelated responsibilities: deciding whether an event applies (payload/identity/cid guards), computing the resulting state (list surgery, poll merges, unread math), and performing side effects (persistence, delivery reconciliation, member re-fetch). A backend payload change and a state-logic change would land in the same method, and none of it could be tested without a full channel lifecycle.

The split separates those along the same lines as the feeds SDK (our newest state architecture — thin event handlers that guard and route, with all mutation logic owned by semantic methods on the state side):

- Each class now has one reason to change: event-shape concerns live in the handler; domain state rules live in the mutations.
- The channel state's mutation surface is explicit for the first time — a reviewable list of named methods instead of logic scattered across listener bodies.
- Write access is enforced structurally: the handler holds no state reference and cannot mutate anything; only the mutations object holds the write capability, including the five injected private paths.
- Each layer is unit-testable in isolation: routing/guards against mocked mutations, state-write logic against a mocked state.

### How this flows into v11

This is the largest subset of the v11 channel refactor achievable without breaking changes, and each piece maps forward:

- **`ChannelStateMutations` is the embryonic write side** of v11's read-only/mutable state split — its method list *is* the mutation contract the mutable state owner needs, discovered and test-pinned now. The five tear-offs mark, by name, exactly which writes must become first-class members of it.
- **`ChannelEventHandler` is the embryonic event-bus subscriber.** The three-block string-typed dispatch exists only to preserve the legacy subscription order; with v11's sealed domain events, the payload guards migrate into the typed event mapping and the ordering constraint can be consciously re-evaluated.
- **The unified REST/WS write path becomes a local change.** The mutation methods take domain payloads (`Message`, `Poll` + `PollVote`, …), not raw events, so routing API responses through the same semantic methods — feeds' single-write-path design, the structural fix for the WS-vs-API dual-write races — only needs new plumbing, not another logic move.
- The side effects deliberately kept in the handler (persistence cleanup, delivery reconciliation, member refresh) are exactly what becomes independent bus subscribers in v11, so they stay clearly marked in the routing layer rather than buried inside state mutations.

Net effect: v11's breaking release is left with visibility moves (hiding mutators, exposing read-only state, file split) instead of logic untangling.

Commits are structured for review: the shown-in-channel predicate extraction, the handler extraction, the handler/mutations split, and doc/changelog updates.

**Testing:** the event coverage added in #2905 (written against the old implementation) passes unchanged against the new one; this PR adds dedicated unit tests for the handler (60 — routing, guards, delegation) and the mutations (53 — state-write logic). Known latent issues are deliberately preserved, not fixed (e.g. `member.added` not deduping, the unguarded `lastReadAt!` in `notification.mark_unread`).

## Screenshots / Videos

No UI changes.

### DIFF
--- a/packages/stream_chat/lib/src/client/channel.dart
+++ b/packages/stream_chat/lib/src/client/channel.dart
@@ -5,6 +5,7 @@ import 'dart:math' as math;
 
 import 'package:collection/collection.dart';
 import 'package:rxdart/rxdart.dart';
+import 'package:stream_chat/src/client/channel_event_handler.dart';
 import 'package:stream_chat/src/client/retry_queue.dart';
 import 'package:stream_chat/src/core/util/utils.dart';
 import 'package:stream_chat/stream_chat.dart';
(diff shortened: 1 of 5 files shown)


VelikovPetar:
Superseded by #2943.

Closing rather than rebasing: #2930 moves `ChannelClientState` out of `channel.dart`, and **99% of this PR's `channel.dart` diff** (924 of 925 changed lines) sits inside that moved block. The refactor's target file no longer contains the code, so neither a merge nor a rebase can resolve it.

#2943 re-hosts the same extraction onto `channel_client_state.dart`. Diffed against this PR's own post-refactor region, it differs by exactly 8 lines, all of them #2930/#2935 adaptations (3 × `_channel.client`, 4 × `scheduleRetry`, 1 × `watcherCountStream.distinct()`).

One behavioral integration was required: #2933 landed on `master` afterwards, turning `_listenChannelMessageCount` into `_listenChannelCounts` so channel events also refresh `Channel.memberCount`. That listener is one of the 36 this PR deletes, so the change was folded into the new structure — `_onChannelCounts` on the handler and `onChannelCounts({memberCount, messageCount})` on the mutations — preserving #2933's semantics including the partial case. All 5 of master's `Channel member count events` tests pass unchanged through the new handler path.

PR #2922, opened 2026-08-25
fix(ui): migrate to file_picker 12

### LINKED ISSUE #2917: Remove KGP from Android dependencies

stream_chat_flutter

Android

`environment:
  sdk: ^3.12.0

dependencies:
  flutter:
    sdk: flutter

  cupertino_icons: ^1.0.8
  stream_chat_flutter: ^10.3.0

### What happened?

`PS D:\AndroidProjects\stream_chat_demo> flutter build apk --release
WARNING: Your app uses the following plugins that apply Kotlin Gradle Plugin (KGP): desktop_drop, device_info_plus, file_picker, package_info_plus, photo_manager, record_android, share_plus, wakelock_plus
Future versions of Flutter will fail to build if your app uses plugins that apply KGP.

Please check the changelogs of these plugins and upgrade to a version that supports Built-in Kotlin.
If no such version exists, report the issue to the plugin. If necessary, here is a guide on filing 
an issue against a plugin: https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-app-developers#report-incompatible-kotlin-gradle-plugin-usage-to-plugin-authors

If you are a plugin author, please migrate your plugin to Built-in Kotlin using this guide: https://docs.flutter.dev/release/breaking-changes/migrate-to-built-in-kotlin/for-plugin-authors
Font asset "MaterialIcons-Regular.otf" was tree-shaken, reducing it from 1645184 to 1368 bytes (99.9% reduction). Tree-shaking can be disabled by providing the --no-tree-shake-icons flag when building your app.
D:\AndroidProjects\stream_chat_demo\android\app\src\main\java\io\flutter\plugins\GeneratedPluginRegistrant.java:39: error: cannot find symbol
      flutterEngine.getPlugins().add(new com.mr.flutter.plugin.filepicker.FilePickerPlugin());
                                                                         ^
  symbol:   class FilePickerPlugin
  location: package com.mr.flutter.plugin.filepicker
1 error

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:compileReleaseJavaWithJavac'.
> Compilation failed; see the compiler output below.
  D:\AndroidProjects\stream_chat_demo\android\app\src\main\java\io\flutter\plugins\GeneratedPluginRegistrant.java:39: error: cannot find symbol
        flutterEngine.getPlugins().add(new com.mr.flutter.plugin.filepicker.FilePickerPlugin());
                                                                           ^
    symbol:   class FilePickerPlugin
    location: package com.mr.flutter.plugin.filepicker
  1 error

* Try:
> Check your code and dependencies to fix the compilation error(s)
> Run with --scan to generate a Build Scan (Powered by Develocity).

BUILD FAILED in 10m 58s
Running Gradle task 'assembleRelease'...                          659.2s
Gradle task assembleRelease failed with exit code 1`

bitgandtter:
# Submit a pull request

Linear: FLU-

Github Issue: #2917 (partially — this removes `file_picker` from the list of plugins that apply KGP)

## CLA

Migrates `stream_chat_flutter` from `file_picker` ^11 to ^12.

### Why this matters beyond the version number

`stream_chat_flutter` is the only package in a typical dependency tree that constrains `file_picker`, and `file_picker` 11 is the last remaining declarer of `win32 ^5` that does not move — `device_info_plus`, `package_info_plus`, `share_plus`, `flutter_secure_storage_windows`, `wakelock_plus` and `win32_registry` all require `win32 ^6.0.1` in their current releases. `file_picker` 12 went federated and dropped its `win32` dependency entirely, so this one constraint holds the whole floor down. In the app this was developed against, this change alone let **23 packages** upgrade in a single `pub get`.

It also unblocks **AGP 9**. `file_picker` 11 guards its Kotlin plugin on the AGP major only:

```groovy
def isAgp9OrAbove = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION.tokenize('.')[0].toInteger() >= 9
if (!isAgp9OrAbove) { apply plugin: 'org.jetbrains.kotlin.android' }
```

Under AGP 9 with `android.builtInKotlin=false` — which is what `flutter create` writes on 3.47.1 — nothing applies a Kotlin plugin to it and its Kotlin is never compiled. The build then fails in `GeneratedPluginRegistrant.java` with `cannot find symbol: class FilePickerPlugin`, a Java error that points nowhere near the cause. `file_picker` 12's Android implementation (`android_file_picker`) reads the property and works in both modes.

### What changed

**`PlatformFileX` becomes asynchronous.** `file_picker` 12 removed the eagerly-loaded `PlatformFile.bytes` and `.size` getters in favour of `readAsBytes()` / `readAsByteStream()`. The new implementation mirrors the `XFileX` extension that already sits directly below it in the same file, so the shape is not new to this codebase.

**Both attachment handlers use `FilePicker.pickFile`.** `pickFiles` now returns `List<PlatformFile>` rather than a result wrapper. Rather than adapting `.files.first`, the handlers use the singular `pickFile`, which returns the `PlatformFile?` these methods already wanted. That also fixes a latent bug: `result?.files.first` threw on an empty selection instead of returning `null`.

**Not a breaking change for callers.** `StreamAttachmentHandler.pickFile`'s public signature is untouched. `withData`, `withReadStream` and `lockParentWindow` remain parameters; they are simply no longer forwarded, since `file_picker` 12 deprecated them in favour of on-demand reads — which is exactly what `toAttachmentFile` now does.

`FileType` and `FilePickerStatus` are unchanged in 12, so no other call site needed touching.

### How this was tested

- `dart analyze lib` on the patched package: clean.
- `flutter test test/src/attachment/attachment_handler_test.dart test/src/utils/extension_test.dart`: 42 passing. These are the suites covering the changed paths. I did not run the full package suite to completion locally.
- Consumed from a production Flutter app via a `dependency_overrides` git ref and exercised end to end: `dart analyze --fatal-infos` clean, 4720 unit/widget tests green, and a Maestro E2E suite (14 flows, three screen sizes) covering attachment picking on device. Built on Flutter 3.47.1 / Dart 3.13.1, AGP 9.3.2, Gradle 9.7.1, KGP 2.4.10 — a configuration that could not build at all before this change.

Happy to adjust the approach if you would rather keep `pickFiles` and adapt the list handling instead, or if the async change on a public extension needs a deprecation path.

## Screenshots / Videos

Not applicable — no user-visible or layout change; this is a dependency migration behind the existing attachment-picker API.

## Summary by CodeRabbit

* **Bug Fixes**
  * File selection now safely returns no attachment when the picker is closed or no file is selected.
  * Improved attachment creation by reading file content on demand and calculating size from loaded bytes.
  * Preserved parent-window locking behavior across supported desktop platforms.
  * Updated file selection for compatibility with `file_picker` 12.

* **Documentation**
  * Documented migration notes, deprecated options, and an Android build configuration workaround.


packages/stream_chat_flutter/pubspec.yaml
```diff
@@ -30,7 +30,7 @@ dependencies:
   diacritic: ^0.1.6
   dio: ^5.11.0
   ezanimation: ^0.6.0
-  file_picker: ^11.0.0
```
renefloor, line 33: Can you also update the version change in [melos.yaml](https://github.com/GetStream/stream-chat-flutter/blob/master/melos.yaml#L55), otherwise this change will be reverted with `melos bs`

packages/stream_chat_flutter/lib/src/utils/extensions.dart
```diff
@@ -115,19 +115,24 @@ extension IterableExtension<T> on Iterable<T> {
...
+      size: bytes.length,
       bytes: bytes,
-      size: size,
     );
   }
 
   /// Converts the [PlatformFile] to a [Attachment].
-  Attachment toAttachment({required String type}) {
-    final file = toAttachmentFile;
+  Future<Attachment> toAttachment({required String type}) async {
```
renefloor, line 132: This is technically a breaking change, because the method is exported.
This at least needs a separate mention in `⚠️ Changed`, now it's a bit hidden in the changelog.

packages/stream_chat_flutter/lib/src/attachment/handler/stream_attachment_handler_io.dart
```diff
@@ -134,19 +134,24 @@ class StreamAttachmentHandler extends StreamAttachmentHandlerBase {
...
   }) async {
-    final result = await FilePicker.pickFiles(
+    // pickFile (singular) since file_picker 12: it returns the `PlatformFile?`
+    // this method already wanted, instead of a result wrapper we immediately
+    // took `.files.first` from — which threw on an empty selection.
+    //
+    // `withData` / `withReadStream` are no longer forwarded: file_picker 12
+    // deprecated them in favour of reading through PlatformFile.readAsBytes()
+    // / readAsByteStream(), which is what toAttachmentFile now does. They stay
+    // in this method's signature so this is not a breaking change for callers.
```
renefloor: No need for such a comment in the code itself, this is more for the PR description.

packages/stream_chat_flutter/pubspec.yaml
bitgandtter: Done — `melos.yaml` now pins `file_picker: ^12.1.0`. This was also the cause of every red job: `melos bootstrap` rewrote the package manifest back to 11, so analyze failed on `pickFile` and `readAsBytes` being undefined, and everything downstream compiled against the old API. Verified locally with `melos bootstrap` followed by `melos run analyze` and `melos run format`.

packages/stream_chat_flutter/lib/src/utils/extensions.dart
bitgandtter: Agreed. Moved to its own `⚠️ Changed` entry marked BREAKING, saying that both extensions are exported and callers have to `await` them. The dependency bump and the `withData`/`withReadStream` note stay under `🔄 Changed`.

packages/stream_chat_flutter/lib/src/attachment/handler/stream_attachment_handler_io.dart
bitgandtter: Removed — both handlers now just call `FilePicker.pickFile`. Also trimmed the doc comment on `toAttachmentFile` to the one line a caller needs.

packages/stream_chat_flutter/lib/src/attachment/handler/stream_attachment_handler_io.dart
```diff
@@ -134,19 +134,24 @@ class StreamAttachmentHandler extends StreamAttachmentHandlerBase {
...
       dialogTitle: dialogTitle,
       initialDirectory: initialDirectory,
       type: type,
       allowedExtensions: allowedExtensions,
       onFileLoading: onFileLoading,
       compressionQuality: compressionQuality,
-      withData: withData,
-      withReadStream: withReadStream,
-      lockParentWindow: lockParentWindow,
     );
```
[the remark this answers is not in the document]
bitgandtter: Good catch — `lockParentWindow` was silently dropped. Restored via `windowsOptions: WindowsOptions(lockParentWindow: lockParentWindow)` and the matching `linuxOptions`, so the parameter keeps working on desktop instead of falling back to the `false` default.

melos.yaml
```diff
@@ -52,7 +52,7 @@ command:
       # upper bounds once FlutterFire ships a fixed release.
       firebase_crashlytics: '>=5.2.0 <5.2.5'
       firebase_messaging: '>=16.0.0 <16.4.2'
-      file_picker: ^11.0.0
+      file_picker: ^12.1.0
```
xsahil03x: A caret on 12 pins us to a bump we'd have to repeat almost immediately — 13.1.0 is already out and 13.0.0 only removed parameters this PR stopped passing. A range covers both majors, and `melos.yaml` already uses this form for `firebase_crashlytics` a few lines up.

```suggestion
      file_picker: '>=12.0.0 <14.0.0'
```

packages/stream_chat_flutter/pubspec.yaml
```diff
@@ -30,7 +30,7 @@ dependencies:
...
+  file_picker: ^12.1.0
```
xsahil03x: Same range here, so the published constraint matches `melos.yaml`.

```suggestion
  file_picker: '>=12.0.0 <14.0.0'
```

packages/stream_chat_flutter/lib/src/attachment/handler/stream_attachment_handler_io.dart
```diff
@@ -134,19 +134,18 @@ class StreamAttachmentHandler extends StreamAttachmentHandlerBase {
     bool withReadStream = false,
```
xsahil03x: `withData` and `withReadStream` are now accepted and ignored. The default is `withData: true`, so the common path is unchanged — but a caller who passed `false` on mobile to avoid materializing a large video now gets an unconditional `readAsBytes()` and no way out. Deprecate them rather than silently dropping them.

```suggestion
    @Deprecated('Content is read on demand; this no longer has any effect.') bool withReadStream = false,
```

`withData` on the line above needs the same annotation, as do both parameters in `stream_attachment_handler_html.dart` and in `stream_attachment_handler_base.dart:34-35` — none of those lines are reachable by a suggestion from this diff. Leave `lockParentWindow` alone; it's still honoured through `WindowsOptions`/`LinuxOptions`.

packages/stream_chat_flutter/lib/src/attachment/handler/stream_attachment_handler_html.dart
```diff
@@ -25,19 +25,18 @@ class StreamAttachmentHandler extends StreamAttachmentHandlerBase {
     bool withReadStream = false,
```
xsahil03x: Same deprecation here, plus `withData` on the line above.

```suggestion
    @Deprecated('Content is read on demand; this no longer has any effect.') bool withReadStream = false,
```

packages/stream_chat_flutter/CHANGELOG.md
```diff
@@ -8,13 +8,17 @@
 
 - Long-pressing a reaction chip no longer opens the message actions modal; the chips always claim the long press. Left unset, `onReactionLongPress` defaults to opening the `ReactionDetailSheet`.
 - Tapping or long-pressing a reaction chip now opens the `ReactionDetailSheet` pre-filtered to that reaction; it previously opened unfiltered. Clustered and overflow chips map to no single reaction, so they still open unfiltered.
+- **BREAKING**: `PlatformFileX.toAttachmentFile` and `PlatformFileX.toAttachment` are now asynchronous, returning `Future<AttachmentFile>` and `Future<Attachment>`. `file_picker` 12 removed `PlatformFile`'s eagerly-loaded `bytes` and `size` getters, so the content is read on demand. Both are exported, so callers must `await` them; this matches the existing `XFileX` extensions.
```
xsahil03x: This hunk isn't on current `master` — the surrounding context (the `🔄 Changed` section, the reaction-chip entries) comes from a base that has since moved, so it needs a rebase before the entry lands anywhere useful.

When you redo it: `## Upcoming` on master has no breaking section, but `## 10.0.1` in this file carries a `🛑️ Breaking` header, so add that section rather than prefixing `**BREAKING**` inside `⚠️ Changed`. No suggestion here because the right edit depends on what the rebase produces.

Worth adding one line for integrators while you're in here: an app on a pre-3.44 Android template has no `subprojects { project.evaluationDependsOn(":app") }` in its root `build.gradle` and will hit the same `cannot find symbol: class FilePickerPlugin` that the `sample_app` change fixes. That failure points nowhere near its cause, and nothing else in this PR helps them.

packages/stream_chat_flutter/CHANGELOG.md
```diff
@@ -8,13 +8,17 @@
...
 
 🔄 Changed
 
 - Raised minimum Flutter to `>=3.44.0` and Dart SDK to `^3.12.0`.
+- Bumped `file_picker` to `^12.1.0`.
+- `StreamAttachmentHandler.pickFile` no longer forwards `withData` and `withReadStream`; `file_picker` deprecated them in favour of reading the content on demand. The method's own signature is unchanged.
```
xsahil03x: These need to track the constraint range and the deprecation.

```suggestion
- Bumped `file_picker` to `>=12.0.0 <14.0.0`.
- Deprecated `withData` and `withReadStream` on `StreamAttachmentHandler.pickFile`. Content is now read on demand, so both are ignored.
```

xsahil03x (commented):
Thanks for this — the `win32` floor analysis and the AGP 9 diagnosis are both right, and switching to `FilePicker.pickFile` is the correct call rather than adapting `.files.first`.

Two asks inline. The first is the one that matters.

**Use a range, not a caret.** `file_picker` 13.1.0 is out, and 13.0.0 only removed the deprecated parameters this PR had already stopped passing. Every symbol the patched code touches — `FilePicker.pickFile`, `windowsOptions`/`linuxOptions`, `readAsBytes()`, `name`, `path`, `FileType`, `FilePickerStatus` — is unchanged across 12.0.0 and 13.1.0. (`PlatformFile` became an abstract base class in `file_picker_platform_interface` 4.0.0, but the surface used here is identical.) `'>=12.0.0 <14.0.0'` keeps your 12.x testing a valid resolution, lets apps already on 13 resolve, and avoids repeating the bump in a few weeks.

**The async change on `PlatformFileX` is the right shape.** To answer the deprecation path you offered in the description: keeping the sync pair working isn't possible. `AttachmentFile` asserts `!CurrentPlatform.isWeb || bytes != null`, and neither 12 nor 13 exposes bytes synchronously, so a sync bytes-less build throws on web and yields an attachment `toMultipartFile()` cannot upload.

The changelog entry is worth expanding with your own reasoning: anyone holding a `PlatformFile` obtained it from `FilePicker.pickFiles()`, whose return type changed in 12, so that call site is being rewritten regardless and the `await` costs nothing on top.

Optional, and not a blocker: the read could be skipped entirely on IO. All three internal consumers of `file.bytes` fall back to `path` (`image_attachment_thumbnail.dart:147`, `extensions.dart:592`, `extensions.dart:707`), so reading only on web would preserve what `withData: false` used to buy. It costs `AttachmentFile.toMultipartFile`'s bytes fallback for unreadable paths, and it would invalidate the E2E run you already did.

The `sample_app` Gradle change is correct and correctly scoped — every `packages/*/example/android/build.gradle` already carries `evaluationDependsOn(':app')`, and only `sample_app` did not.

The CLA checkbox is still unticked.

melos.yaml
bitgandtter: Done in 272ba9f — `'>=12.0.0 <14.0.0'`. I also ran `flutter analyze` on the package against both 12.1.0 and 13.1.0; both are clean.

packages/stream_chat_flutter/pubspec.yaml
bitgandtter: Done in 272ba9f, same range as `melos.yaml`.

packages/stream_chat_flutter/lib/src/attachment/handler/stream_attachment_handler_io.dart
bitgandtter: Done in 272ba9f: `withData` and `withReadStream` are deprecated in `stream_attachment_handler_io.dart`, `stream_attachment_handler_html.dart` and `stream_attachment_handler_base.dart`; `lockParentWindow` is untouched.

One more place needed it: `StreamFilePicker` (the SDK's own attachment picker option) declares both flags, defaults `withData` to `false`, and forwarded them to `pickFile`. I deprecated both fields there too and stopped forwarding them, so the SDK doesn't call its own deprecated parameters.

packages/stream_chat_flutter/lib/src/attachment/handler/stream_attachment_handler_html.dart
bitgandtter: Done in 272ba9f, both parameters.

packages/stream_chat_flutter/CHANGELOG.md
bitgandtter: Rebased onto current `master`. The entries now live under `## Upcoming`, with a new `🛑️ Breaking` section for the async `PlatformFileX` change (including the note that the `pickFiles()` call site is being rewritten anyway). I also added the pre-3.44 Android template line about `subprojects { project.evaluationDependsOn(":app") }` and the `cannot find symbol: class FilePickerPlugin` failure.

packages/stream_chat_flutter/CHANGELOG.md
bitgandtter: Applied in 272ba9f. The deprecation line also names `StreamFilePicker`, since its flags are deprecated too.

xsahil03x (approved):

packages/stream_chat_flutter/test/src/attachment/attachment_handler_pick_file_test.dart
```diff
@@ -0,0 +1,112 @@
...
+  Uri get uri => Uri.file('/picked/$name');
+
+  @override
+  XFile get xFile => XFile.fromData(_bytes, name: name, path: uri.toFilePath());
+
+  @override
+  int? lengthSync() => _bytes.length;
+
+  @override
+  Future<int?> length() async => _bytes.length;
```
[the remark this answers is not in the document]
xsahil03x: `length()` now returns `Future<int>` in a07cd4113 — it overrides the non-nullable 12.x signature and still satisfies the nullable 13.x one.

Keeping `@override` on `lengthSync()`: CI resolves `file_picker` 13.1.0, where it does override, and dropping the annotation trips `annotate_overrides` under `dart analyze --fatal-infos`.

PR #2929, opened 2026-08-26
chore(repo): prefer relative imports inside a package's own lib/

xsahil03x:
### 🎯 Goal

Move the repo from `always_use_package_imports` to `prefer_relative_imports` for imports within a package's own `lib/`, and migrate every existing directive accordingly.

Two reasons:

- **The OpenAPI generator emits relative imports.** Every generated model does `import '../models.dart'`, so the generated client trips `always_use_package_imports` — 494 infos, which is a red `melos run analyze` under `--fatal-infos`. Silencing that meant either excluding the whole generated tree from analysis (which also hides real errors) or patching an `ignore_for_file` line into every generated file on every regeneration. Adopting the generator's convention makes both unnecessary.
- **Both sibling repos already do this.** `stream_core_flutter` and `stream_feeds_flutter` each disable `always_use_package_imports` with the comment *"conflicts with `prefer_relative_imports`"*. This aligns chat with core and feeds.

### 📝 Summary

- `analysis_options.yaml`: `always_use_package_imports` → `prefer_relative_imports`. `avoid_relative_lib_imports` stays enabled — it guards against reaching into a `lib/` from outside it, which is still an error.
- Mechanical migration via `dart fix --apply --code=prefer_relative_imports` across **all** packages, then `--code=directives_ordering` to re-sort the sections (`package:` imports first, then relative, each alphabetical).
- Three groups `dart fix` cannot reach, done by hand:
  - `lib/scrollable_positioned_list/**` — vendored and excluded in `analysis_options.yaml`, so no lint fires there.
  - two `export 'package:stream_chat_flutter/…'` directives (`poll_interactor_theme.dart`, `stream_attachment_picker_controller.dart`) — no lint covers `export`.
- `CLAUDE.md` and `STYLE_GUIDE.md` updated to document the new convention.

**Scope:** every `lib/` in the repo. After this, `grep -rn "package:<own_package>/" packages/*/lib` returns nothing for all five published packages, and all 607 relative directives resolve on disk.

`test/` and `example/` keep their `package:` imports, deliberately:

- A test reaching into its own `lib/` by relative path is an error — `avoid_relative_lib_imports` flags it (*"Can't use a relative path to import a library in 'lib'"*), and Dart would treat it as a separate library from the `package:` one.
- `packages/*/example` are separate packages, so `package:stream_chat_flutter/…` there is a cross-package import and must stay as-is.

No CHANGELOG entries: import style is internal, with no client-facing effect.

### 🧪 Testing

[...]

PR #2913, opened 2026-08-21
refactor(llc): extract the message merge algebra into MessageMerging

VelikovPetar:
# Submit a pull request

Linear: FLU-724

Github Issue: #

## CLA

- [x] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required).
- [x] The code changes follow best practices
- [x] Code changes are tested (add some information if not applicable)

## Description of the pull request

Moves the pure message/pin/live-location merge and removal operations (formerly private on `ChannelClientState`) verbatim out of `channel.dart` into a new unexported static holder `MessageMerging` under `lib/src/core/util/`, following the `MessageRules` precedent. The two message predicates (`isShownInChannel` and `hasValidPin`, formerly the top-level `_pinIsValid` and inline filters) live in a separate unexported `MessagePredicates` extension on `Message`.

- Call sites in `channel.dart` are requalified; the stateful orchestrators and persistence writes stay put.
- The thread orchestrators preserve their exact write/no-write emission behavior via an `identical()` guard.
- Adds pure unit-test suites (41 tests across `message_merging_test.dart` and `message_predicates_test.dart`) covering the merge semantics directly, including previously untested pin-expiry filtering and the thread phantom-guard / empty-thread-pruning rules.
- Zero public API change (`lib/stream_chat.dart` untouched); no CHANGELOG entry since there is no observable behavior change.

**Review tip:** `git diff --color-moved=dimmed-zebra` shows the moved bodies as relocated text.

**Test instructions:** `cd packages/stream_chat && dart test` — full package suite passes (1,665 tests, including the existing through-state merge characterization); downstream `stream_chat_flutter_core`, `stream_chat_persistence`, and `stream_chat_flutter` suites verified.

Groundwork for the v11 shared merge-semantics rail.

## Screenshots / Videos

No UI changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2301, opened 2025-07-06
refactor(llc): introduce event controller, resolver

xsahil03x:
## Summary by CodeRabbit

* **New Features**
  * Enhanced event handling to emit more specific poll answer events in chat streams.

* **Refactor**
  * Redesigned event stream management with a new controller supporting conditional event resolution.

* **Tests**
  * Updated test utilities to support the refined event handling mechanism.

PR #2930, opened 2026-08-26
refactor(llc): split `channel.dart` into focused files

VelikovPetar:
# Submit a pull request

Linear: FLU-749

Github Issue: #

## CLA

Groundwork for [FLU-481](https://linear.app/stream/issue/FLU-481/split-channel-into-focused-service-objects), which asks for two things as a first step: move responsibilities out of `Channel` **while keeping the public API unchanged**, and **move state out of `channel.dart`**. This does the second, with zero public API change.

`channel.dart` declared five top-level things, four of which had nothing to do with the `Channel` class itself. They move into a new `src/client/channel/` directory, which `channel.dart` itself joins:

| File | Contents | Lines |
| --- | --- | --- |
| `channel/channel.dart` | `class Channel` and nothing else | 2,468 |
| `channel/channel_client_state.dart` | `class ChannelClientState` + the private `_pinIsValid` (used only by it) | 2,155 |
| `channel/channel_capability_check.dart` | `extension ChannelCapabilityCheck on Channel` | 235 |
| `channel/channel_read_helper.dart` | `extension ChannelReadHelper on ChannelClientState` | 52 |

**`channel.dart`: 4,896 → 2,468 lines.** `channel_delivery_reporter.dart` and `retry_queue.dart` stay at `src/client/` — they exist on `master`, so moving them would be a rename with the cost described below, for no benefit here.

### Cost of moving `channel.dart`, and why it is accepted

Moving `channel.dart` is **not** a breaking change: `lib/src/` is package-private by Dart convention, this repo enables the `implementation_imports` lint, and the barrel's exported namespace is unchanged (proven below). But it does have one real, measured cost.

Because this PR also cuts `channel.dart` from 4,896 to 2,468 lines, the file is only ~50% similar to its `master` counterpart — right at git's default rename-detection threshold. Verified empirically: `git diff-tree -M` reports `delete mode` + `create mode`, **not** a rename. So for an open PR that modifies `channel.dart`, merging this produces a modify/delete conflict.

Measured against the three open PRs that touch it:

| PR | Changed lines in `channel.dart` | In the retained `Channel` class | In the moved-out block |
| --- | --- | --- | --- |
| #2847 offline reactions | 15 | **15** | 0 |
| #2863 location timers | 36 | 1 | 35 |
| #2871 unread banners | 6 | 0 | 6 |

Only **#2847** is newly affected — its 15 lines merge cleanly today and will now need hand-re-applying. #2863 and #2871 already have to be re-derived regardless, because their hunks land inside the block that moved to `channel_client_state.dart`.

### Why this is not a breaking change

The four names are exported from the public barrel directly, alongside every other public type in the package:

```dart
export 'src/client/channel/channel.dart';
export 'src/client/channel/channel_capability_check.dart';
export 'src/client/channel/channel_client_state.dart';
export 'src/client/channel/channel_read_helper.dart';
```

Before, `channel.dart` *declared* exactly four public names: `Channel`, `ChannelClientState`, `ChannelReadHelper`, `ChannelCapabilityCheck`, and the barrel exported that one file. Now the barrel exports four files declaring the same four names. **A consumer importing `package:stream_chat/stream_chat.dart` cannot tell the difference** — `show`, `hide`, prefixes, implicit extension application and explicit extension application all resolve identically either way.

An earlier revision instead had `channel.dart` re-export the three new files, so that a deep `package:stream_chat/src/client/channel.dart` import kept resolving all four. That became pointless once `channel.dart` itself moved into `channel/`, so the re-export block is gone and the barrel is now the single explicit list of public API.

Verified with a probe that imports **only the barrel** — no deep imports — and exercises:

- all four names in value positions, including implicit member access (`channel.canSendMessage`) and explicit extension application (`ChannelCapabilityCheck(channel).canSendReply`, `ChannelReadHelper(state).readsOf(...)`)
- `ChannelClientState` in type positions: local, `List<>`, typedef, and a hand-written `implements` mock
- `usesLocalUnreadCount`, the one moved line whose body changed

It analyzes clean under `--fatal-infos`. Downstream, `stream_chat_flutter`, `stream_chat_flutter_core` and `stream_chat_persistence` all analyze clean against the new barrel, and their suites pass (362 and 302 respectively).

Negative control confirms the barrel exports are load-bearing rather than incidentally redundant: with the three stripped, the probe fails with 9 errors; restored, 0.

One internal file did depend on the old re-export and the analyzer caught it: `core/util/message_rules.dart` deep-imports `channel.dart` and uses the capability extension, so it now imports `channel/channel_capability_check.dart` explicitly.

### The one API addition

Moving the state class out of the library broke 25 private cross-accesses that only compiled because both classes shared a file. All were fixed by requalifying onto existing public equivalents:

- **21 ×** `state!._channelState` → `state!.channelState` — the private getter was a byte-identical duplicate of the public one (`=> _channelStateController.value`), so this is a compiler-verified rename
- **4 ×** `state?._retryQueue.add([msg])` → `state?.scheduleRetry(msg)`
- **4 ×** `_channel._client` / `_client` → the public `client` getter, which returns the same field

That leaves exactly one API delta: **`@internal ChannelClientState.scheduleRetry(Message)`**, whose body is the identical `_retryQueue.add([message])`.

It is unavoidable — `Channel` is the caller, so injection isn't available, and the only public alternative, `retryFailedMessages()`, is argument-less and rescans state, so it can't carry a specific message. It is also the mildest possible form of addition:

- Adding a member is additive. It's only breaking for code that hand-writes `implements` with every member spelled out, which is a routine minor-version change in this SDK.
- `@internal` keeps it out of the documented surface: external callers get `invalid_use_of_internal_member`, a warning, not an error.
- `ClientState` — the barrel-exported sibling class — already carries **six** `@internal` members for exactly this purpose, so this is the established pattern rather than a new one.
- Mocks are unaffected; `implements` + `noSuchMethod` and mocktail's `extends Mock implements` both compile clean (covered by the matrix above).

### Test suite mirrors the split

`channel_test.dart` 12,356 → 6,687, with the groups that exercise the moved code relocated to files matching the new sources: `channel_client_state_test.dart` (5,124), `channel_capability_check_test.dart` (410), `channel_read_helper_test.dart` (319). Which groups moved was decided by measuring each group's state-vs-channel orientation; genuinely mixed groups were left whole rather than split internally. Each new file carries a private copy of the two fixtures it used from `main()`, as `channel_delivery_reporter_test.dart` already does.

Measuring the extracted capability suite in isolation also exposed three members that only ever had **incidental** coverage from elsewhere in the package, so they gain direct tests (+8):

- `usesLocalUnreadCount` — the full `isLocalUnreadCountEnabled` × read-receipts matrix. Previously reached only via `Channel.markRead`, and it is also the one line of moved code that changed, so it was the riskiest thing here and had no direct test.
- `canUseDeliveryReceipts` — absent from the parameterized capability list entirely.
- `canUseReadReceipts` — covered only through its *deprecated* alias `canReceiveReadEvents`, so coverage would have silently vanished when that alias is removed.

Both new source files are now at 100% line coverage from their own suites, and the parameterized entries follow declaration order.

### Merged up to `master` (`c66577890`)

`master` moved three commits into `channel.dart` after this branch was cut, two of which land **inside the block that moved out** — exactly the changes a hand-resolved conflict would drop silently. Rather than resolve the conflict by hand, the split was **regenerated from `master`'s `channel.dart`**: the file was re-sliced at the same four declaration boundaries and only the requalifications listed above were reapplied, each under an exact-occurrence-count assertion so any drift would abort rather than mis-patch.

What that carried into `channel_client_state.dart`:

- `#2933` — `_listenChannelMessageCount` → `_listenChannelCounts`, now also applying `Event.channelMemberCount`
- `#2935` — `.distinct()` on `watcherCountStream`

Plus, in `channel.dart`, all 17 `.distinct()` additions from `#2935`, and in `channel_test.dart`, the 209 new lines of `memberCount` tests.

The new files also adopt the relative-import convention from `#2929`, which is now lint-enforced via `prefer_relative_imports`.

### Verification

- **Behaviour proven by reconstruction:** diffing each of the four files against its exact line range in `master`'s `channel.dart` yields **70 changed lines in total** — `channel.dart` 54 (the import header, 21 `channelState`, 4 `scheduleRetry`), `channel_client_state.dart` 10 (3 `client`, 4 declaring `scheduleRetry`), `channel_read_helper.dart` 4 (a `[msg]` → `[message]` doc fix), `channel_capability_check.dart` 2 (1 `client`). Zero unexplained changes; no logic line from `master` is absent.
- **Public API diff vs master:** top-level declarations identical; `Channel` public members unchanged; `ChannelCapabilityCheck` unchanged; `ChannelReadHelper` unchanged; `ChannelClientState` gains only `scheduleRetry`. Nothing removed or renamed anywhere.
- **Tests:** every non-trivial code line of `master`'s `channel_test.dart` (8,326 lines) is present across the four test files with at least the same multiplicity — **0 missing**. All 375 `test`/`group` declarations from `master` are present; the only additions are the 5 declarations of the new `usesLocalUnreadCount` group.
- **Whole-tree diff vs `master`** is 16 files: the 8 moved into `channel/`, plus 8 whose only change is an import or export path (`stream_chat.dart`, `client.dart`, `query_channels_result.dart`, `channel_delivery_reporter.dart`, `message_rules.dart`, `matchers.dart`, `mocks.dart`). Every other file in the repo is byte-identical to `master`.
- `melos run analyze` clean across all packages; `dart format` clean.
- `stream_chat` 1,639 · `stream_chat_flutter_core` 362 · `stream_chat_persistence` 302 — all passing.

No CHANGELOG entry: internal restructuring with no observable behaviour change, matching the precedent set by `refactor(llc): introduce event controller, resolver` (#2301).

### The in-flight refactors have been re-derived

99% of #2911's and #2913's `channel.dart` diffs landed inside the block that moved here, so neither could be rebased through this. All three dependent PRs have been closed and re-derived on top of this branch, in merge order:

| New | Replaces | Content |
| --- | --- | --- |
| #2942 | #2905 | characterization tests, moved into `channel_client_state_test.dart` |
| #2943 | #2911 | event handler / state mutations extraction |
| #2944 | #2913 | `MessageMerging` extraction |

That cost was accepted deliberately, so that the file move happens once, up front, and both extractions land directly in their final home. Public member sets of `Channel` and `ChannelClientState` are identical across the whole stack, and `stream_chat` goes 1,639 → 1,682 → 1,797 → 1,838 tests, all passing.

## Screenshots / Videos

No UI changes.

## Summary by CodeRabbit

- **New Features**
  - Added comprehensive channel state management for messages, threads, members, drafts, reactions, polls, reminders, locations, pins, and unread counts.
  - Added capability checks for messaging, moderation, attachments, reactions, read receipts, typing indicators, polls, and other channel actions.
  - Added helpers and streams for tracking message read and delivery status.

- **Bug Fixes**
  - Improved retry handling for failed message sends, updates, partial updates, and deletions.
  - Improved synchronization between online and offline channel state.


xsahil03x (approved):
LGTM, Should we move the whole channel related code in a separate /channel directory?

PR #2934, opened 2026-09-03
perf(llc): skip replaying oversized `/sync` payloads

xsahil03x:
Replaying a large `/sync` payload through `handleEvent` holds local persistence and state updates long enough to slow down the regular requests that need them. Payloads over 250 events are no longer replayed — the synced channels are re-queried in their place.

Implements the [Sync events limit](https://app.notion.com/p/stream-wiki/Sync-events-limit-3666a5d7f9f6801f97ecc1461fc62049) spec, matching `stream-chat-swift` (`SyncRepository`) and `stream-chat-android` (`SyncManager`), which use the same 250-event threshold.

Ticket: [FLU-756](https://linear.app/stream/issue/FLU-756/sync-events-limit-skip-replaying-oversized-sync-payloads)

### What changed

- `StreamChatClient.sync` skips event replay when a `/sync` response carries more than 250 events.
- On reconnect the synced channels are re-queried with `queryChannels` in their place, one 30-channel page at a time (the server returns at most 30 channels per request, so a single query would silently cover only the first page).
- `lastSyncAt` advances only once that refresh succeeded. Dropping the events is safe when their state has been re-fetched; advancing past a failed refresh would lose them.
- Mark-all-read events from a skipped payload are still applied, since a channel refresh does not carry read state.
- `lastSyncAt` fallbacks now use `DateTime.timestamp()` instead of local time.

### Behavior change worth a second look

On reconnect, the re-query after a skipped replay runs **even when `recoverStateOnReconnect` is `false`** (which `StreamChatCore` sets for every app built on `stream_chat_flutter`). Skipping replay drops in-memory channel state, so without it there is nothing to repair the currently open channel. When the refresh does run, the flag's own re-query is skipped so the channels are not queried twice.

A direct `client.sync()` call issues no queries — only the reconnect path opts into the refresh.

### Verification

`dart analyze --fatal-infos`, `dart format`, and the full `stream_chat` suite (1632 tests) pass. New tests cover the skip, the re-query in its place, page batching past 30 channels, `lastSyncAt` being kept when the refresh fails, and mark-all-read surviving the skip.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reconnect recovery by refreshing active channels when synchronization data is too large to replay.
  - Preserved read-status updates during oversized synchronization responses.
  - Ensured synchronization progress continues after recovery attempts, while retaining the previous position if channel refresh fails.
  - Reduced unnecessary `memberCountStream` updates by emitting only when the member count changes.
  - Refreshed active channels in batches to improve recovery reliability.


xsahil03x:
Superseded by #2945, which contains this commit plus the FLU-756 follow-ups it enabled: the cid cap, the sync-age pre-check, the refresh-before-advance on a refused window, and the extraction into `SyncManager`.

The changelog conflict on this branch is resolved and pushed, so it is mergeable if you would rather land it on its own first.

PR #2943, opened 2026-09-07
refactor(llc): extract channel event handling into handler and state mutations

VelikovPetar:
# Submit a pull request

Linear: FLU-723

Github Issue: #

## CLA

> [!IMPORTANT]
> **#2930 and #2942 are merged; this PR now sits directly on `master`** and its diff is only the six files below. Merge before #2944, which builds on it.
>
> Replaces #2911, which could not be rebased: 99% of its `channel.dart` diff (924 of 925 lines) was inside the block #2930 moved into `channel_client_state.dart`.
>
> Everything this PR touches or adds lives under `src/client/channel/` — including `channel_event_handler.dart` and `channel_state_mutations.dart`, moved there in their own commit so they sit with the state class they serve.

Moves the channel event handling out of `ChannelClientState` into two new internal classes, with no public API changes and one behavior change (below):

- **`ChannelEventHandler`** — validates and routes each WS event from a single subscription (replacing the 36 per-event subscriptions), preserving the original per-subscription execution order via five named dispatch stages. Also owns the side effects an event triggers outside the channel state: member refresh on ban/unban, persisted-message cleanup on truncation, and delivery reconciliation.
- **`ChannelStateMutations`** — owns the state writes, one semantic method per event (`onMemberRemoved`, `onPollVoteCasted`, …). The few writes that previously went through private state (typing events, watcher removal, member refresh, user message deletion) stay private on `ChannelClientState` and are injected as tear-offs, so the state class gains no new members.

### One behavior change

Routing every event through a single subscription means each dispatch stage is now wrapped in a guard. Previously a handler that threw left an unhandled error in the root zone, which crash reporters listening on `PlatformDispatcher.onError` report as fatal; it is now caught and logged at `WARNING`, and the later stages still run for the same event. The four `async` handlers (`channel.truncated`, `user.banned`, `user.unbanned`, `user.messages.deleted`) are outside that guarantee — their futures are discarded, so a failure after the first `await` still escapes. Both facts are documented on `handleEvent`, and the containment is the one line this PR adds to the CHANGELOG.

### Why these classes exist

Previously, every event listener interleaved three unrelated responsibilities: deciding whether an event applies (payload/identity/cid guards), computing the resulting state (list surgery, poll merges, unread math), and performing side effects (persistence, delivery reconciliation, member re-fetch). A backend payload change and a state-logic change would land in the same method, and none of it could be tested without a full channel lifecycle.

The split separates those along the same lines as the feeds SDK (our newest state architecture — thin event handlers that guard and route, with all mutation logic owned by semantic methods on the state side):

- Each class now has one reason to change: event-shape concerns live in the handler; domain state rules live in the mutations.
- The channel state's mutation surface is explicit for the first time — a reviewable list of named methods instead of logic scattered across listener bodies.
- Write access is enforced structurally: the handler holds no state reference and cannot mutate anything; only the mutations object holds the write capability, including the five injected private paths.
- Each layer is unit-testable in isolation: routing/guards against mocked mutations, state-write logic against a mocked state.

### How this flows into v11

This is the largest subset of the v11 channel refactor achievable without breaking changes, and each piece maps forward:

- **`ChannelStateMutations` is the embryonic write side** of v11's read-only/mutable state split — its method list *is* the mutation contract the mutable state owner needs, discovered and test-pinned now. The five tear-offs mark, by name, exactly which writes must become first-class members of it.
- **`ChannelEventHandler` is the embryonic event-bus subscriber.** The three-block string-typed dispatch exists only to preserve the legacy subscription order; with v11's sealed domain events, the payload guards migrate into the typed event mapping and the ordering constraint can be consciously re-evaluated.
- **The unified REST/WS write path becomes a local change.** The mutation methods take domain payloads (`Message`, `Poll` + `PollVote`, …), not raw events, so routing API responses through the same semantic methods — feeds' single-write-path design, the structural fix for the WS-vs-API dual-write races — only needs new plumbing, not another logic move.
- The side effects deliberately kept in the handler (persistence cleanup, delivery reconciliation, member refresh) are exactly what becomes independent bus subscribers in v11, so they stay clearly marked in the routing layer rather than buried inside state mutations.

Net effect: v11's breaking release is left with visibility moves (hiding mutators, exposing read-only state, file split) instead of logic untangling.

**Testing:** the event coverage from #2942 (written against the old implementation) passes unchanged against the new one; this PR adds dedicated unit tests for the handler (80 — routing, guards, delegation, error isolation) and the mutations (54 — state-write logic), including new cases for the partial-count behavior, plus one end-to-end isolation test in `channel_client_state_test.dart`. That is **135 new tests**, taking the `stream_chat` suite to **1,828**. Known latent issues are deliberately preserved, not fixed (e.g. `member.added` not deduping, the unguarded `lastReadAt!` in `notification.mark_unread`).

The handler tests mock `ChannelStateMutations` and the mutations tests mock `ChannelClientState` — collaborators rather than the outside-world boundaries the style guide points at. That is deliberate: isolating those two seams is what the split exists to make possible, and handler↔mutations agreement is still covered end-to-end by the #2942 event tests running against the real pair.

## Screenshots / Videos

No UI changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when processing channel events, so an issue with one event no longer prevents unrelated updates from being applied.
  * Improved synchronization for messages, unread counts, reactions, polls, read and delivery status, members, watchers, reminders, shared locations, drafts, and notification preferences.
  * Correctly filters messages that should not appear in a channel and preserves relevant user reactions, poll responses, and delivery information during updates.
  * Fixed duplicate live-location expiration events and WebAssembly platform initialization issues.
  * Channel-event processing errors are now logged as warnings instead of propagating unexpectedly.

### DIFF
--- a/packages/stream_chat/lib/src/client/channel/channel_event_handler.dart
+++ b/packages/stream_chat/lib/src/client/channel/channel_event_handler.dart
@@ -1,4 +1,4 @@
-import '../../stream_chat.dart';
+import '../../../stream_chat.dart';
 import 'channel_state_mutations.dart';
 
 /// Routes channel events to the matching [ChannelStateMutations] methods.
--- a/packages/stream_chat/lib/src/client/channel/channel_state_mutations.dart
+++ b/packages/stream_chat/lib/src/client/channel/channel_state_mutations.dart
@@ -2,7 +2,7 @@ import 'dart:math' as math;
 
 import 'package:collection/collection.dart';
 
-import '../../stream_chat.dart';
+import '../../../stream_chat.dart';
 
 /// Applies channel event payloads as [ChannelClientState] mutations.
 ///
--- a/.github/actions/pana/action.yml
+++ b/.github/actions/pana/action.yml
@@ -31,12 +31,7 @@ runs:
           yq eval '.dependencies.stream_chat = {"path": "../stream_chat"}' -i packages/stream_chat_persistence/pubspec.yaml
 
     - name: Install Flutter
-      uses: subosito/flutter-action@v2
-      with:
-        cache: true
-        channel: stable
-        flutter-version: "3.x"
-        cache-key: flutter-:os:-:channel:-:version:-:arch:-:hash:-${{ hashFiles('**/pubspec.lock') }}
+      uses: ./.github/actions/setup-flutter
 
     - name: Install Pana
       working-directory: ${{ inputs.working_directory }}
--- a/.github/actions/setup-flutter/action.yml
+++ b/.github/actions/setup-flutter/action.yml
@@ -1,12 +1,26 @@
 name: 'Setup Flutter'
 description: 'Install Flutter with dependency caching'
+inputs:
+  flutter-version:
+    description: 'Flutter version to install. Empty selects the newest on the channel.'
+    required: false
+    default: '3.x'
+  channel:
+    description: 'Flutter release channel'
+    required: false
+    default: 'stable'
+  cache-sdk:
+    description: 'Cache the Flutter SDK. Turn off for jobs that run too rarely to hit it.'
+    required: false
+    default: 'true'
 runs:
   using: "composite"
   steps:
     - uses: subosito/flutter-action@v2
       with:
-        flutter-version: "3.x"
-        channel: stable
-        cache: true
+        flutter-version: ${{ inputs.flutter-version }}
+        channel: ${{ inputs.channel }}
+        cache: ${{ inputs.cache-sdk }}
+        pub-cache: 'true'
         cache-key: "flutter-:os:-:channel:-:version:-:arch:"
         pub-cache-key: "flutter-pub-:os:-${{ hashFiles('melos.yaml', '**/pubspec.yaml') }}"
--- a/.github/workflows/beta_version_analyze.yml
+++ b/.github/workflows/beta_version_analyze.yml
@@ -21,11 +21,16 @@ jobs:
           fetch-depth: 0
 
       - name: "Install Flutter"
-        uses: subosito/flutter-action@v2
+        uses: ./.github/actions/setup-flutter
         with:
           channel: beta
-          cache: true
-          cache-key: flutter-:os:-:channel:-:version:-:arch:-:hash:-${{ hashFiles('**/pubspec.lock') }}
+          # Must stay empty. Inheriting the action's 3.x default would hold this
+          # job on 3.x betas, so it would stop catching the next major early —
+          # which is the only thing it exists to do.
+          flutter-version: ''
+          # Weekly, and a new beta invalidates the key, so an SDK entry would
+          # almost never be restored before it expires.
+          cache-sdk: 'false'
 
       - name: 📊 Analyze and test packages
         uses: ./.github/actions/package_analysis
--- a/.github/workflows/legacy_version_analyze.yml
+++ b/.github/workflows/legacy_version_analyze.yml
@@ -65,11 +65,9 @@ jobs:
           fetch-depth: 0
 
       - name: "Install Flutter"
-        uses: subosito/flutter-action@v2
+        uses: ./.github/actions/setup-flutter
         with:
           flutter-version: ${{ env.flutter_version }}
-          channel: stable
-          cache-key: flutter-:os:-:channel:-:version:-:arch:-:hash:-${{ hashFiles('**/pubspec.lock') }}
 
       - name: 📊 Analyze and test packages
         uses: ./.github/actions/package_analysis
--- a/.github/workflows/release_publish.yml
+++ b/.github/workflows/release_publish.yml
@@ -29,7 +29,7 @@ jobs:
         uses: dart-lang/setup-dart@v1
 
       - name: 🐦 Install Flutter
-        uses: subosito/flutter-action@v2
+        uses: ./.github/actions/setup-flutter
 
       - name: 📦 Install Tools
         run: flutter pub global activate melos
--- a/.github/workflows/stream_flutter_workflow.yml
+++ b/.github/workflows/stream_flutter_workflow.yml
@@ -177,6 +177,8 @@ jobs:
         with:
           fetch-depth: 0
       - uses: ./.github/actions/setup-flutter
+      - uses: ./.github/actions/cache-gradle
+        if: matrix.platform == 'android'
       - uses: maxim-lobanov/setup-xcode@v1
         if: matrix.platform == 'ios'
         with:
--- a/.github/workflows/update_goldens.yml
+++ b/.github/workflows/update_goldens.yml
@@ -27,11 +27,7 @@ jobs:
         uses: actions/checkout@v7
 
       - name: 🐦 Install Flutter
-        uses: subosito/flutter-action@v2
-        with:
-          flutter-version: "3.x"
-          channel: stable
-          cache-key: flutter-:os:-:channel:-:version:-:arch:-:hash:-${{ hashFiles('**/pubspec.lock') }}
+        uses: ./.github/actions/setup-flutter
 
       - name: 📦 Install Tools
         run: flutter pub global activate melos
@@ -76,11 +72,7 @@ jobs:
         uses: actions/checkout@v7
 
       - name: 🐦 Install Flutter
-        uses: subosito/flutter-action@v2
-        with:
-          flutter-version: "3.x"
-          channel: stable
-          cache-key: flutter-:os:-:channel:-:version:-:arch:-:hash:-${{ hashFiles('**/pubspec.lock') }}
+        uses: ./.github/actions/setup-flutter
 
       - name: 📦 Install Tools
         run: flutter pub global activate melos
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -140,6 +140,12 @@ After modifying any package, update its `CHANGELOG.md`.
 
 UI designs for this SDK are in the [Chat SDK Design System](https://www.figma.com/design/Us73erK1xFNcB5EH3hyq6Y/Chat-SDK-Design-System) Figma project. Use the Figma MCP server to look up designs when implementing or updating UI components.
 
+## Sibling SDK Alignment
+
+Stream ships the same product on other platforms — **stream-chat-swift**, **stream-chat-android**, **stream-chat-react-native**. Check how they name and shape an equivalent API before inventing one here, and prefer their vocabulary over a name derived only from Flutter code or design tokens.
+
+That alignment belongs in the **PR description**, never in the SDK's own docs or code comments. A comment that justifies a decision with "the SwiftUI and React Native SDKs do it this way" tells a reader nothing about what this code guarantees, and goes stale the moment those SDKs change — an integrator reading the dartdoc cannot verify it and has no reason to care. Describe the behavior and the trade-off directly; put the cross-platform argument, where it is genuinely persuasive, in the PR. See [`STYLE_GUIDE.md`](STYLE_GUIDE.md#no-cross-framework-justification-in-comments).
+
 ## `stream_core_flutter` (external sibling repo)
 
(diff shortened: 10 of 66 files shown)


packages/stream_chat/lib/src/client/channel/channel_event_handler.dart
```diff
@@ -0,0 +1,525 @@
...
+
+    return _mutations.onMemberUpdated(member);
+  }
+
+  Future<void> _onMemberBanned(Event event) async {
+    // Filters channel ban from app ban.
+    if (event.cid == null) return;
+
+    final user = event.user!;
+    final member = await _channel.queryMembers(filter: Filter.equal('id', user.id)).then((it) => it.members.first);
```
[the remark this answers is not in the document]
VelikovPetar: Good catch on the mechanics, but this one is unchanged from master, so I'd rather not fix it here.                                                                              
                                                                                                                                                                              
The four handlers you list are `async`, which means a synchronous throw inside them gets packaged into the returned Future rather than thrown — so the `try/catch` in `handleEvent` was never going to see it. Before this PR they were bare `.listen((event) async { ... })` callbacks and the `Future` was discarded there too, so a failing `queryMembers` ended up as an unhandled async error either way.

Please re-check the change against the pre-refactor behaviour: it should be identical - or correct me if I am wrong.

packages/stream_chat/lib/src/client/channel/channel_event_handler.dart
```diff
@@ -0,0 +1,525 @@
...
+
+  /// Handles the given channel [event].
+  ///
+  /// Routes the event by its type through three dispatch blocks, running
+  /// [_onChannelCounts] between the first and second block and
+  /// [_onMemberUserUpdated] between the second and third; those two observe
+  /// every event regardless of its type.
+  ///
+  /// Each of those five regions is contained: one that throws is logged and
+  /// the remaining regions still run for the same event.
```
xsahil03x: `_onChannelTruncated`, `_onMemberBanned`, `_onMemberUnbanned` and `_onUserMessagesDeleted` are `async`, and the switch cases discard their futures — so a rejected `queryMembers`, `.first` on an empty member list, or a failed persistence delete never reaches this `try`. It surfaces as an unhandled async error instead of a logged warning.

Not a regression; the old listeners were `async` too. But the `error isolation` group only exercises sync throws, so nothing marks where the guarantee actually stops, and this doc says it doesn't stop anywhere. Worth either a test pinning the gap or a clause here saying the `async` handlers are outside it — whichever you'd rather carry into v11.

packages/stream_chat/lib/src/client/channel/channel_event_handler.dart
```diff
@@ -0,0 +1,525 @@
...
+        case EventType.messageDelivered:
+          _onMessageDelivered(event);
+        // channel events
+        case EventType.channelTruncated:
+        case EventType.notificationChannelTruncated:
+          _onChannelTruncated(event);
+        case EventType.channelUpdated:
+          _onChannelUpdated(event);
+      }
+    } catch (error, stackTrace) {
```
xsahil03x: These six lines appear five times, identically (here and at 108, 125, 136, 180). A helper would cut ~25 lines and make the stage sequence readable without scrolling:

```dart
void _guard(Event event, void Function() region) {
  try {
    region();
  } catch (error, stackTrace) {
    _client.logger.warning('Error handling ${event.type} event', error, stackTrace);
  }
}
```

`handleEvent` then reads as five `_guard(event, () { … })` calls. Not blocking — and it'd want the whole method re-indented, so it's your call whether it's worth the diff noise here.

packages/stream_chat/lib/src/client/channel/channel_event_handler.dart
```diff
@@ -0,0 +1,525 @@
...
+      _client.logger.warning(
+        'Error handling ${event.type} event',
+        error,
+        stackTrace,
+      );
+    }
+
+    // Merges the event user into the member list on any event carrying one.
+    try {
+      _onMemberUserUpdated(event);
```
xsahil03x: This ordering is load-bearing and nothing pins it. `member.updated` carries both `member` and `user`, where `user` is the actor who made the change — so when someone updates their own membership, both this and `_onMemberUpdated` touch the same member. Running the merge first means `event.member` wins; swapping them would leave `event.user` copied over it.

Worth a test, otherwise a future reorder is silent. (I read this off the event payload rather than running it, so double-check me.)

packages/stream_chat/lib/src/client/channel/channel_event_handler.dart
```diff
@@ -0,0 +1,525 @@
...
+
+  void _onPollVoteRemoved(Event event) {
+    final (eventPoll, eventPollVote) = (event.poll, event.pollVote);
+    if (eventPoll == null || eventPollVote == null) return;
+
+    return _mutations.onPollVoteRemoved(eventPoll, eventPollVote);
+  }
+
+  void _onMessageRead(Event event) {
+    // Skip handling the event if delivered for a thread
```
xsahil03x: Reads, not deliveries. Carried over from the old code so no rush, but it sits next to a guard people will reason about.

```suggestion
    // Skip handling the event if the read is for a thread.
```

packages/stream_chat/lib/src/client/channel/channel_event_handler.dart
```diff
@@ -0,0 +1,525 @@
...
+  Future<void> _onUserMessagesDeleted(Event event) async {
+    final user = event.user;
+    if (user == null) return;
+
+    return _mutations.onUserMessagesDeleted(
+      userId: user.id,
+      hardDelete: event.hardDelete ?? false,
+      deletedAt: event.createdAt,
+    );
+  }
```
xsahil03x: No `await` in here, so the `async` only turns sync throws into a discarded rejected future. Same quirk in the old listener, but worth dropping while it's moving anyway. Neither `discarded_futures` nor `unawaited_futures` is enabled, so this analyzes clean.

```suggestion
  void _onUserMessagesDeleted(Event event) {
    final user = event.user;
    if (user == null) return;

    _mutations.onUserMessagesDeleted(
      userId: user.id,
      hardDelete: event.hardDelete ?? false,
      deletedAt: event.createdAt,
    );
  }
```

packages/stream_chat/lib/src/client/channel/channel_state_mutations.dart
```diff
@@ -0,0 +1,585 @@
+import 'dart:math' as math;
+
+import 'package:collection/collection.dart';
+
+import '../../../stream_chat.dart';
+
+/// Applies channel event payloads as [ChannelClientState] mutations.
+///
+/// Each method computes the new state from the current one and the given
+/// payload, and writes it to the state.
```
xsahil03x: The doc says what this does but not why it's a separate class rather than methods on `ChannelClientState` — which was the first thing I wondered, and the answer isn't reconstructible from the code.

```suggestion
/// Applies channel event payloads as [ChannelClientState] mutations.
///
/// Each method computes the new state from the current one and the given
/// payload, and writes it to the state.
///
/// Kept off [ChannelClientState] itself because that class is exported, and
/// v11 shrinks its write surface rather than widening it.
```

packages/stream_chat/lib/src/client/channel/channel_state_mutations.dart
```diff
@@ -0,0 +1,585 @@
...
+    _state.addNewMessage(message);
+
+    if (watcherCount != null) {
+      _state.updateChannelState(
+        _state.channelState.copyWith(watcherCount: watcherCount),
+      );
+    }
+  }
+
+  /// Marks the [message] as deleted.
```
xsahil03x: Doesn't mention that it also decrements the locally-tracked unread count for hard deletes — the only arithmetic in the method, and the reason it needs `_channel`.

```suggestion
  /// Marks the [message] as deleted, decrementing the locally-tracked unread
  /// count when a hard delete removes a message that counted as unread.
```

With that in the doc, the five-line `//` block below it is mostly restating the same thing; it could come down to the one fact the doc can't carry — that server-driven channels get corrected counts from read events instead.

xsahil03x (commented):
Went through this as a move-refactor — diffed every old `_listen*` body against its new handler/mutation pair rather than reading the new files on their own. All 41 event types map 1:1, including the poll merge maps and the `copyWith(memberCount:, messageCount:)` expression, which are character-identical.

Three things I went looking for and was glad not to find: the ordering is preserved exactly (including that the generic user-merge subscription used to be registered *before* `member.updated`), `_channel.on()` still applies the same cid filter and still lands in `_subscriptions`, and `_removeWatcher` keeps the direct `_channelState = …` write instead of being "cleaned up" into `updateChannelState`. That last one was the easiest thing in the whole diff to break silently.

Nothing blocking. Two comments below worth a look — the untested ordering constraint on `member.updated` is the one I'd actually act on.

On the tests: the fixtures are good — a real `Logger` capturing records, `Fake`s confined to `registerFallbackValue`, real `Message`/`Member`/`User` payloads, `verifyNever` on every drop path. The factoring is where I'd push back. Several tests in `channel_event_handler_test.dart` bundle independent routing rules — `'poll vote events delegate the poll and vote'` covers five event types with five assertions, and `'poll.updated and poll.closed delegate the poll'` and `'reminder.created and reminder.updated delegate the reminder'` each cover two. TESTING.md asks for one behavior per test precisely so a failure names itself; right now if `poll.vote_changed` routing breaks you get a red line saying "poll vote events delegate the poll and vote" and have to open the file. Same in the four `'… ignore events without a …'` tests, which fire two events each. Cheap to split, and a routing table is exactly what you want failing one row at a time.

Also worth naming: the tests mock `ChannelStateMutations` and `ChannelClientState`, which are collaborators rather than the outside-world boundaries the style guide points at. Defensible — isolating those seams is the point of the split, and the #2942 event tests still cover handler↔mutations agreement — but better acknowledged than left to wonder about.

Skipping the CHANGELOG looks right to me — the policy gates on "changes package behavior" and this is a move. The only thing that nudges it is that errors which used to escape to the zone are now caught and logged at WARNING. Your call whether that's observable enough for a `🔄 Internal / Non-breaking` line.

packages/stream_chat/lib/src/client/channel/channel_event_handler.dart
VelikovPetar: I would keep the current behaviour and update the docs - if that is OK with you. Since most of this will undergo changes in `v11`, I would keep the behaviour unchanged for `v10`.

packages/stream_chat/lib/src/client/channel/channel_event_handler.dart
VelikovPetar: Sounds like a good suggestion, I will re-work it!

packages/stream_chat/lib/src/client/channel/channel_event_handler.dart
VelikovPetar: Writing a test to pin the behaviour!

xsahil03x (approved):

packages/stream_chat/lib/src/client/channel/channel_event_handler.dart
```diff
@@ -0,0 +1,501 @@
...
+
+  // Runs the given [stage] of [event] handling, containing a synchronous
+  // failure so the remaining stages still run for the same event.
+  void _guard(Event event, String stage, void Function() region) {
+    try {
+      region();
+    } catch (error, stackTrace) {
+      _client.logger.warning('Error handling ${event.type} event in $stage', error, stackTrace);
+    }
+  }
```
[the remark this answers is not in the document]
VelikovPetar: The gap is deliberate, and documented two lines below the sentence you're quoting: "The four async handlers are outside that guarantee — their futures are discarded here, so a failure after the first await surfaces as an unhandled async error instead of a logged warning."

This is also the same behaviour as before the refactor - please re-evaluate your finding and correct me if I am wrong

Issue #12, opened 2020-04-06
MessageInput grows indefinitely


nosmirck:
Message Input grows with message content but doesn't stop, reaching an overflow.

Image attached for reference:

![Simulator Screen Shot - iPhone 11 Pro Max - 2020-04-06 at 11 47 25](https://user-images.githubusercontent.com/6027290/78578802-05bea480-77fe-11ea-8916-43817ee58d62.png)

If there's no one working on this one, I can PR a fix today.

imtoori:
We'll fix it as soon as possible, however feel free to contribute

nosmirck:
@imtoori  PR is up! :)

Issue #11, opened 2020-04-06
Dark Theme


nosmirck:
Dark Theme seems to be not properly implemented, despite the advertised images on the website.

Using the system theme mode dark (after setting the themes correctly) like this:
```
Widget build(BuildContext context) {
    return MaterialApp(
      theme: ThemeData.light(),
      darkTheme: ThemeData.dark(),
      themeMode: ThemeMode.system,
      home: Container(
        child: StreamChat(
          client: client,
          child: ChannelListPage(),
        ),
      ),
    );
  }
```
causes this:

![Simulator Screen Shot - iPhone 11 Pro Max - 2020-04-06 at 11 50 36](https://user-images.githubusercontent.com/6027290/78578345-639ebc80-77fd-11ea-9158-31d5604b9478.png)

As seen in the image, nothing is dark themed, plus, the message box has white text over white background.
If there's no one working on this, I can fix it and PR, please let me know.

PR #2944, opened 2026-09-07
refactor(llc): extract the message merge algebra into MessageMerging

VelikovPetar:
# Submit a pull request

Linear: FLU-724

Github Issue: #

## CLA

> [!IMPORTANT]
> **#2930, #2942 and #2943 are all merged.** This PR now sits directly on `master`, with no remaining dependency.
>
> Two review follow-ups widen the diff past the extraction itself: the new test suite is split into four files per `TESTING.md`, and one line each in `CHANGELOG.md` and `channel_event_handler.dart` narrows #2943's containment claim to synchronous failures (`async void` state writes still reach the zone).
>
> Replaces #2913, which could not be rebased: 99% of its `channel.dart` diff (281 of 282 lines) was inside the block #2930 moved into `channel_client_state.dart`.
>
> `channel_client_state.dart` now lives at `src/client/channel/`. `message_merging.dart` and `message_predicates.dart` stay under `src/core/util/`, since they are channel-agnostic.

Moves the pure message/pin/live-location merge and removal operations (formerly private on `ChannelClientState`) verbatim into a new unexported static holder `MessageMerging` under `lib/src/core/util/`, following the `MessageRules` precedent. The two message predicates (`isShownInChannel` and `hasValidPin`, formerly the top-level `_pinIsValid` and inline filters) live in a separate unexported `MessagePredicates` extension on `Message`.

- Call sites in `channel_client_state.dart` are requalified (16 `MessageMerging.*` calls, 4 `MessagePredicates` usages); the stateful orchestrators and persistence writes stay put.
- The thread orchestrators preserve their exact write/no-write emission behavior via an `identical()` guard.
- Adds pure unit-test suites (45 tests across `message_merging_{messages,threads,pinned,locations}_test.dart` and `message_predicates_test.dart`) covering the merge semantics directly, including previously untested pin-expiry filtering and the thread phantom-guard / empty-thread-pruning rules.
- Zero public API change (`lib/stream_chat.dart` untouched). The extraction has no observable behavior change; the only CHANGELOG edit narrows the entry #2943 added.

`ChannelClientState` drops from 1,272 to 1,032 lines, and all nine of the private merge/removal helpers are gone.

**Test instructions:** `cd packages/stream_chat && dart test` — full package suite passes (1,873 tests), including the existing through-state merge characterization and #2942's event coverage.

Groundwork for the v11 shared merge-semantics rail.

### Follow-up worth considering

`MessagePredicates` is now the natural home for other message predicates that remain inline in the channel state. Not done here to keep the diff a faithful re-derivation.

## Screenshots / Videos

No UI changes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved channel message visibility handling, especially for threaded replies.
  - Improved message updates and removals while preserving shared locations and quoted-message references.
  - Improved sorting and consistency when messages are added, updated, deleted, or moved.
  - Expired, deleted, or invalid pinned messages and live locations are now filtered more reliably.
  - Improved thread and watcher state updates, including removal of empty threads and inactive locations.
  - Synchronous channel event-handling errors are now logged as warnings instead of reaching the root zone.
- **Tests**
  - Added coverage for message visibility, pins, locations, threads, sorting, and message merging.

### DIFF
--- a/packages/stream_chat/lib/src/client/channel/channel_event_handler.dart
+++ b/packages/stream_chat/lib/src/client/channel/channel_event_handler.dart
@@ -1,4 +1,4 @@
-import '../../stream_chat.dart';
+import '../../../stream_chat.dart';
 import 'channel_state_mutations.dart';
 
 /// Routes channel events to the matching [ChannelStateMutations] methods.
--- a/packages/stream_chat/lib/src/core/util/message_merging.dart
+++ b/packages/stream_chat/lib/src/core/util/message_merging.dart
@@ -1,0 +1,302 @@
+import '../models/location.dart';
+import '../models/message.dart';
+import 'list_extensions.dart';
+import 'message_predicates.dart';
+
+/// Provides the merge and removal operations reconciling incoming messages
+/// with the locally-held channel state collections.
+///
+/// Every operation is pure: it computes a new collection from the given
+/// inputs without reading or writing any state.
+class MessageMerging {
+  const MessageMerging._();
+
+  /// The default `update` strategy for the merge operations: merges the
+  /// incoming [updated] into the locally-known [original] via
+  /// [Message.updateWith], preserving enrichment the server may strip on
+  /// partial payloads.
+  static Message mergeUpdate(Message original, Message updated) => original.updateWith(updated);
+
+  /// The replacing `update` strategy: takes the incoming [updated] as-is.
+  /// Used by local rollback paths.
+  static Message replaceUpdate(Message _, Message updated) => updated;
+
+  /// Compares [a] and [b] by their [Message.createdAt].
+  static int sortByCreatedAt(Message a, Message b) => a.createdAt.compareTo(b.createdAt);
+
+  /// Merges the live locations carried by [toMerge] into [existing].
+  ///
+  /// Locations are keyed by (userId, channelCid, createdByDeviceId); an
+  /// incoming message's [Message.sharedLocation] replaces the existing entry
+  /// with the same key. Locations that are expired or whose attached message
+  /// is deleted are dropped from the result.
+  static Iterable<Location> mergeActiveLocations({
+    required Iterable<Location> existing,
+    required Iterable<Message> toMerge,
+  }) {
+    if (toMerge.isEmpty) return existing;
+
+    final mergedLocations = existing.mergeFrom(
+      toMerge,
+      key: (it) => (it.userId, it.channelCid, it.createdByDeviceId),
+      value: (message) => message.sharedLocation,
+      update: (original, updated) => updated,
+    );
+
+    final toUpdateMap = {for (final m in toMerge) m.id: m};
+    final updatedLocations = mergedLocations.where((it) {
+      // Remove the location if it's expired.
+      if (it.isExpired) return false;
+
+      final updatedMessage = toUpdateMap[it.messageId];
+      // Remove the location if the attached message is deleted.
+      if (updatedMessage?.isDeleted == true) return false;
+
+      return true;
+    });
+
+    return updatedLocations;
+  }
+
+  /// Merges [toMerge] into the [existing] pinned messages, keeping only
+  /// messages that are still valid pins (see [MessagePredicates.hasValidPin]).
+  static Iterable<Message> mergePinnedMessages({
+    required Iterable<Message> existing,
+    required Iterable<Message> toMerge,
+    Message Function(Message original, Message updated) update = mergeUpdate,
+  }) {
+    return mergeMessages(
+      existing: existing,
+      toMerge: toMerge,
+      update: update,
+    ).where((it) => it.hasValidPin);
+  }
+
+  /// Merges [toMerge] into [existing], returning a list sorted by
+  /// [Message.createdAt].
+  ///
+  /// [update] decides whether each pair is reconciled (default — see
+  /// [mergeUpdate]) or replaced ([replaceUpdate], used by local rollback
+  /// paths that don't want enrichment fallback to keep optimistic values).
+  ///
+  /// [upsert] controls whether ids not already in [existing] are inserted.
+  /// Event-driven paths (`message.updated`, `message.deleted` soft) pass
+  /// `upsert: false` so an out-of-window message isn't dropped into a gap
+  /// between the loaded slice and history the client hasn't paged in yet.
+  static Iterable<Message> mergeMessages({
+    required Iterable<Message> existing,
+    required Iterable<Message> toMerge,
+    Message Function(Message original, Message updated) update = mergeUpdate,
+    bool upsert = true,
+  }) {
+    if (toMerge.isEmpty) return existing;
+
+    final existingList = existing is List<Message> ? existing : existing.toList();
+    var toMergeList = toMerge is List<Message> ? toMerge : toMerge.toList();
+
+    // Single-message fast path. The hot ingest path (server echoes, edits,
+    // reactions, read receipts) always lands here, and `lastIndexWhere` +
+    // `sortedUpsertAt` skips the O(N) keymap build that the two-pointer
+    // merge would otherwise do up front.
+    if (toMergeList.length == 1) {
+      final message = toMergeList.first;
+      final oldIndex = existingList.lastIndexWhere((it) => it.id == message.id);
+
+      // upsert: false — skip update if message is not loaded
+      if (oldIndex == -1 && !upsert) return existingList;
+
+      final resolved = oldIndex == -1 ? message : update(existingList[oldIndex], message);
+
+      final mergedMessages = existingList.sortedUpsertAt(
+        oldIndex,
+        resolved,
+        update: update,
+        compare: sortByCreatedAt,
+      );
+
+      // Non-delete updates can't change what embedded quotedMessage copies
+      // should display, so we can skip the rewrite entirely.
+      if (!resolved.isDeleted) return mergedMessages;
+
+      return mergedMessages.updateIf(
+        (it) => it.quotedMessageId == resolved.id,
+        (it) => it.copyWith(quotedMessage: resolved),
+      );
+    }
+
+    // upsert: false - skip messages not loaded in the window
+    if (!upsert) {
+      final existingIds = {for (final m in existingList) m.id};
+      toMergeList = toMergeList.where((m) => existingIds.contains(m.id)).toList();
+      if (toMergeList.isEmpty) return existingList;
+    }
+
+    // Batch path: receiver (`existingList`) is maintained sorted as a
+    // state invariant; `mergeSorted` sorts `toMergeList` internally and
+    // returns a sorted result.
+    final mergedMessages = existingList.mergeSorted(
+      toMergeList,
+      key: (message) => message.id,
+      update: update,
+      compare: sortByCreatedAt,
+    );
+
+    // Refresh embedded `quotedMessage` refs only for messages quoting an
+    // incoming message that is now deleted. `updateIf` returns the same
+    // list reference when nothing matches, so steady-state allocates
+    // nothing for this step.
+    final deletedIds = toMergeList.where((m) => m.isDeleted).map((m) => m.id).toSet();
+    if (deletedIds.isEmpty) return mergedMessages;
+
+    final mergedById = {for (final m in mergedMessages) m.id: m};
+    return mergedMessages.updateIf(
+      (it) => deletedIds.contains(it.quotedMessageId),
+      (it) => it.copyWith(quotedMessage: mergedById[it.quotedMessageId]),
+    );
+  }
+
+  /// Merges [toMerge] into the [existing] threads map, returning the updated
+  /// map, or [existing] as-is when [toMerge] carries no thread replies.
+  ///
+  /// Replies are grouped by their parent id so each thread merge only sees
+  /// its own messages. With [upsert] `false`, replies to threads not present
+  /// in [existing] are dropped instead of creating the thread entry.
+  static Map<String, List<Message>> mergeThreadMessages({
+    required Map<String, List<Message>> existing,
+    required Iterable<Message> toMerge,
+    Message Function(Message original, Message updated) update = mergeUpdate,
+    bool upsert = true,
+  }) {
+    if (toMerge.isEmpty) return existing;
+
+    // Group messages by parentId so each thread merge only sees its own
+    // replies — passing the full batch to every thread would leak replies
+    // across thread boundaries (the merge dedups by id, not by parentId).
+    final messagesByThread = <String, List<Message>>{};
+    for (final m in toMerge) {
+      if (m.parentId case final parentId?) (messagesByThread[parentId] ??= []).add(m);
+    }
+
+    // If there are no affected threads, return early.
+    if (messagesByThread.isEmpty) return existing;
+
+    final updatedThreads = {...existing};
+    for (final MapEntry(key: thread, :value) in messagesByThread.entries) {
+      final existingThreadMessages = updatedThreads[thread];
+
+      // Don't create a phantom entry for a thread that wasn't loaded: with
+      // `upsert: false` an out-of-window reply is dropped, so there's nothing
+      // to merge. Writing it back would make `threads.containsKey(parentId)`
+      // report a thread that was never paged in.
+      if (existingThreadMessages == null && !upsert) continue;
+
+      final threadMessages = existingThreadMessages ?? <Message>[];
+      final updatedThreadMessages = mergeMessages(
+        existing: threadMessages,
+        toMerge: value,
+        update: update,
+        upsert: upsert,
+      );
+
+      // Update the thread with the modified message list.
+      updatedThreads[thread] = updatedThreadMessages.toList();
+    }
+
+    return updatedThreads;
+  }
+
+  /// Removes from [existing] the locations attached to any message in
+  /// [toRemove].
+  static Iterable<Location> removeActiveLocations({
+    required Iterable<Location> existing,
+    required Iterable<Message> toRemove,
+  }) {
+    if (toRemove.isEmpty) return existing;
+
+    final toRemoveIds = toRemove.map((m) => m.id).toSet();
+    final updatedLocations = existing.where(
+      // Remove the location if its attached message is in the toRemove list.
+      (it) => !toRemoveIds.contains(it.messageId),
+    );
+
+    return updatedLocations;
+  }
+
+  /// Removes [toRemove] from the [existing] pinned messages, keeping only
+  /// messages that are still valid pins (see [MessagePredicates.hasValidPin]).
+  static Iterable<Message> removePinnedMessages({
+    required Iterable<Message> existing,
+    required Iterable<Message> toRemove,
+  }) {
+    return removeMessages(
+      existing: existing,
+      toRemove: toRemove,
+    ).where((it) => it.hasValidPin);
+  }
+
+  /// Removes [toRemove] from [existing], clearing the quoted-message
+  /// reference of any remaining message that quotes a removed one.
+  static Iterable<Message> removeMessages({
+    required Iterable<Message> existing,
+    required Iterable<Message> toRemove,
+  }) {
+    if (toRemove.isEmpty) return existing;
+
+    final toRemoveIds = toRemove.map((m) => m.id).toSet();
+    final updatedMessages = existing
+        .where((it) {
+          // Remove the message if it's in the toRemove list.
+          return !toRemoveIds.contains(it.id);
+        })
+        .map((it) {
+          // Continue if the message doesn't quote any of the deleted messages.
+          if (!toRemoveIds.contains(it.quotedMessageId)) return it;
+
+          // Setting it to null will remove the quoted message from the message.
+          return it.copyWith(quotedMessageId: null, quotedMessage: null);
+        });
+
+    return updatedMessages;
+  }
+
+  /// Removes [toRemove] from the [existing] threads map, returning the
+  /// updated map, or [existing] as-is when [toRemove] carries no thread
+  /// replies.
+  ///
+  /// Thread entries left with no messages are dropped from the map.
+  static Map<String, List<Message>> removeThreadMessages({
+    required Map<String, List<Message>> existing,
+    required Iterable<Message> toRemove,
+  }) {
+    if (toRemove.isEmpty) return existing;
+
+    final affectedThreads = {...toRemove.map((it) => it.parentId).nonNulls};
+    // If there are no affected threads, return early.
+    if (affectedThreads.isEmpty) return existing;
+
+    final updatedThreads = {...existing};
+    for (final thread in affectedThreads) {
+      final threadMessages = updatedThreads[thread];
+      // Continue if the thread doesn't exist.
+      if (threadMessages == null) continue;
+
+      // Remove the deleted message from the thread messages and reference from
+      // other messages quoting it.
+      final updatedThreadMessages = removeMessages(
+        existing: threadMessages,
+        toRemove: toRemove,
+      );
+
+      // If there are no more messages in the thread, remove the thread entry.
+      if (updatedThreadMessages.isEmpty) {
+        updatedThreads.remove(thread);
+        continue;
+      }
+
+      // Otherwise, update the thread with the modified message list.
+      updatedThreads[thread] = updatedThreadMessages.toList();
+    }
+
+    return updatedThreads;
+  }
+}
--- a/packages/stream_chat/lib/src/core/util/message_merging.dart
+++ b/packages/stream_chat/lib/src/core/util/message_merging.dart
@@ -156,7 +156,7 @@ class MessageMerging {
   }
 
   /// Merges [toMerge] into the [existing] threads map, returning the updated
-  /// map, or [existing] as-is when [toMerge] carries no thread replies.
+  /// map.
   ///
   /// Replies are grouped by their parent id so each thread merge only sees
   /// its own messages. With [upsert] `false`, replies to threads not present
@@ -260,8 +260,7 @@ class MessageMerging {
   }
 
   /// Removes [toRemove] from the [existing] threads map, returning the
-  /// updated map, or [existing] as-is when [toRemove] carries no thread
-  /// replies.
+  /// updated map.
   ///
   /// Thread entries left with no messages are dropped from the map.
   static Map<String, List<Message>> removeThreadMessages({
--- a/packages/stream_chat/lib/src/core/util/message_predicates.dart
+++ b/packages/stream_chat/lib/src/core/util/message_predicates.dart
@@ -1,0 +1,35 @@
+import '../models/message.dart';
+
+/// Predicates over a [Message] used by the channel state management.
+extension MessagePredicates on Message {
+  /// Whether the message is shown in the channel message list.
+  ///
+  /// Non-thread messages always are; thread replies only when explicitly
+  /// marked to also show in the channel.
+  bool get isShownInChannel {
+    // Non-thread messages are always shown in the channel.
+    if (parentId == null) return true;
+
+    // Thread messages are only shown if explicitly marked.
+    return showInChannel == true;
+  }
+
+  /// Whether the message represents a currently valid pin.
+  ///
+  /// Returns `false` if the message is deleted, not pinned, or its
+  /// [Message.pinExpires] has passed.
+  bool get hasValidPin {
+    // If the message is deleted, the pin is not valid.
+    if (isDeleted) return false;
+
+    // If the message is not pinned, it's not valid.
+    if (pinned != true) return false;
+
+    // If there's no expiration, the pin is valid.
+    final expiresAt = pinExpires;
+    if (expiresAt == null) return true;
+
+    // If there's an expiration, check if it's still valid.
+    return expiresAt.isAfter(DateTime.now());
+  }
+}
--- a/packages/stream_chat/test/src/core/util/message_merging_test.dart
+++ b/packages/stream_chat/test/src/core/util/message_merging_test.dart
@@ -1,0 +1,489 @@
+import 'package:stream_chat/src/core/util/message_merging.dart';
+import 'package:stream_chat/stream_chat.dart';
+import 'package:test/test.dart';
+
+Message _message(
+  String id, {
+  DateTime? createdAt,
+  String? text,
+  String? parentId,
+  bool? showInChannel,
+  String? quotedMessageId,
+  Message? quotedMessage,
+  bool pinned = false,
+  DateTime? pinExpires,
+  String type = 'regular',
+  Location? sharedLocation,
+}) {
+  return Message(
+    id: id,
+    createdAt: createdAt ?? DateTime(2024),
+    text: text,
+    parentId: parentId,
+    showInChannel: showInChannel,
+    quotedMessageId: quotedMessageId,
+    quotedMessage: quotedMessage,
+    pinned: pinned,
+    pinExpires: pinExpires,
+    type: type,
+    sharedLocation: sharedLocation,
+  );
+}
+
+Location _sharedLocation({
+  String? messageId,
+  String? userId = 'user-id',
+  String? channelCid = 'messaging:channel-id',
+  String? createdByDeviceId = 'device-id',
+  DateTime? endAt,
+  double latitude = 0,
+  double longitude = 0,
+}) {
+  return Location(
+    messageId: messageId,
+    userId: userId,
+    channelCid: channelCid,
+    createdByDeviceId: createdByDeviceId,
+    endAt: endAt,
+    latitude: latitude,
+    longitude: longitude,
+  );
+}
+
+Iterable<String?> _ids(Iterable<Message> messages) => messages.map((it) => it.id);
+
+void main() {
+  group('MessageMerging.mergeUpdate', () {
+    test('merges the incoming message into the original, preserving enrichment', () {
+      final location = _sharedLocation(messageId: 'm1', endAt: DateTime.now().add(const Duration(hours: 1)));
+      final original = _message('m1', text: 'old', sharedLocation: location);
+      final updated = _message('m1', text: 'new');
+
+      final result = MessageMerging.mergeUpdate(original, updated);
+
+      expect(result.text, 'new');
+      expect(result.sharedLocation, location);
+    });
+  });
+
+  group('MessageMerging.replaceUpdate', () {
+    test('takes the incoming message as-is, dropping enrichment', () {
+      final location = _sharedLocation(messageId: 'm1', endAt: DateTime.now().add(const Duration(hours: 1)));
+      final original = _message('m1', text: 'old', sharedLocation: location);
+      final updated = _message('m1', text: 'new');
+
+      final result = MessageMerging.replaceUpdate(original, updated);
+
+      expect(result, same(updated));
+      expect(result.sharedLocation, isNull);
+    });
+  });
+
+  group('MessageMerging.sortByCreatedAt', () {
+    test('orders messages by their creation time', () {
+      final earlier = _message('m1', createdAt: DateTime(2024));
+      final later = _message('m2', createdAt: DateTime(2024, 2));
+
+      expect(MessageMerging.sortByCreatedAt(earlier, later), isNegative);
+      expect(MessageMerging.sortByCreatedAt(later, earlier), isPositive);
+      expect(MessageMerging.sortByCreatedAt(earlier, earlier), isZero);
+    });
+  });
+
+  group('MessageMerging.mergeMessages', () {
+    test('returns the existing messages untouched when there is nothing to merge', () {
+      final existing = [_message('m1')];
+
+      final result = MessageMerging.mergeMessages(existing: existing, toMerge: const []);
+
+      expect(result, same(existing));
+    });
+
+    test('inserts a single unknown message in sorted position', () {
+      final existing = [
+        _message('m1', createdAt: DateTime(2024)),
+        _message('m3', createdAt: DateTime(2024, 3)),
+      ];
+      final incoming = _message('m2', createdAt: DateTime(2024, 2));
+
+      final result = MessageMerging.mergeMessages(existing: existing, toMerge: [incoming]);
+
+      expect(_ids(result), ['m1', 'm2', 'm3']);
+    });
+
+    test('updates a single existing message via the default merge strategy', () {
+      final location = _sharedLocation(messageId: 'm1', endAt: DateTime.now().add(const Duration(hours: 1)));
+      final existing = [_message('m1', text: 'old', sharedLocation: location)];
+      final incoming = _message('m1', text: 'new');
+
+      final result = MessageMerging.mergeMessages(existing: existing, toMerge: [incoming]);
+
+      expect(result, hasLength(1));
+      expect(result.first.text, 'new');
+      expect(result.first.sharedLocation, location, reason: 'enrichment should survive a stripped payload');
+    });
+
+    test('upsert: false skips a single message that is not loaded', () {
+      final existing = [_message('m1')];
+      final incoming = _message('m2');
+
+      final result = MessageMerging.mergeMessages(existing: existing, toMerge: [incoming], upsert: false);
+
+      expect(result, same(existing));
+    });
+
+    test('rewrites the embedded quote on quoters when the incoming message is deleted', () {
+      final quoted = _message('m1', createdAt: DateTime(2024), text: 'quoted');
+      final quoter = _message(
+        'm2',
+        createdAt: DateTime(2024, 2),
+        quotedMessageId: 'm1',
+        quotedMessage: quoted,
+      );
+      final deleted = _message('m1', createdAt: DateTime(2024), type: MessageType.deleted);
+
+      final result = MessageMerging.mergeMessages(existing: [quoted, quoter], toMerge: [deleted]);
+
+      final updatedQuoter = result.singleWhere((it) => it.id == 'm2');
+      expect(updatedQuoter.quotedMessage?.isDeleted, isTrue);
+    });
+
+    test('does not rewrite the embedded quote on quoters for a non-delete update', () {
+      final quoted = _message('m1', createdAt: DateTime(2024), text: 'quoted');
+      final quoter = _message(
+        'm2',
+        createdAt: DateTime(2024, 2),
+        quotedMessageId: 'm1',
+        quotedMessage: quoted,
+      );
+      final edited = _message('m1', createdAt: DateTime(2024), text: 'edited');
+
+      final result = MessageMerging.mergeMessages(existing: [quoted, quoter], toMerge: [edited]);
+
+      final updatedQuoter = result.singleWhere((it) => it.id == 'm2');
+      expect(updatedQuoter.quotedMessage?.text, 'quoted');
+    });
+
+    test('interleaves a batch of unknown messages in sorted order', () {
+      final existing = [
+        _message('m1', createdAt: DateTime(2024)),
+        _message('m3', createdAt: DateTime(2024, 3)),
+      ];
+      final incoming = [
+        _message('m4', createdAt: DateTime(2024, 4)),
+        _message('m2', createdAt: DateTime(2024, 2)),
+      ];
+
+      final result = MessageMerging.mergeMessages(existing: existing, toMerge: incoming);
+
+      expect(_ids(result), ['m1', 'm2', 'm3', 'm4']);
+    });
+
+    test('upsert: false only applies the batch entries that are already loaded', () {
+      final existing = [_message('m1', text: 'old')];
+      final incoming = [
+        _message('m1', text: 'new'),
+        _message('m2'),
+        _message('m3'),
+      ];
+
+      final result = MessageMerging.mergeMessages(existing: existing, toMerge: incoming, upsert: false);
+
+      expect(_ids(result), ['m1']);
+      expect(result.first.text, 'new');
+    });
+
+    test('honors a replacing update strategy for a batch', () {
+      final location = _sharedLocation(messageId: 'm1', endAt: DateTime.now().add(const Duration(hours: 1)));
+      final existing = [
+        _message('m1', text: 'old', sharedLocation: location),
+        _message('m2', createdAt: DateTime(2024, 2)),
+      ];
+      final incoming = [
+        _message('m1', text: 'new'),
+        _message('m2', createdAt: DateTime(2024, 2)),
+      ];
+
+      final result = MessageMerging.mergeMessages(
+        existing: existing,
+        toMerge: incoming,
+        update: MessageMerging.replaceUpdate,
+      );
+
+      final replaced = result.singleWhere((it) => it.id == 'm1');
+      expect(replaced.text, 'new');
+      expect(replaced.sharedLocation, isNull);
+    });
+
+    test('rewrites the embedded quote on quoters when a batch entry is deleted', () {
+      final quoted = _message('m1', createdAt: DateTime(2024), text: 'quoted');
+      final quoter = _message(
+        'm3',
+        createdAt: DateTime(2024, 3),
+        quotedMessageId: 'm1',
+        quotedMessage: quoted,
+      );
+      final incoming = [
+        _message('m1', createdAt: DateTime(2024), type: MessageType.deleted),
+        _message('m2', createdAt: DateTime(2024, 2)),
+      ];
+
+      final result = MessageMerging.mergeMessages(existing: [quoted, quoter], toMerge: incoming);
+
+      final updatedQuoter = result.singleWhere((it) => it.id == 'm3');
+      expect(updatedQuoter.quotedMessage?.isDeleted, isTrue);
+    });
+  });
+
+  group('MessageMerging.mergeThreadMessages', () {
+    test('returns the existing threads untouched when nothing targets a thread', () {
+      final existing = {
+        'p1': [_message('m1', parentId: 'p1')],
+      };
+
+      final result = MessageMerging.mergeThreadMessages(existing: existing, toMerge: [_message('m2')]);
+
+      expect(result, same(existing));
+    });
+
+    test('groups replies into their own threads', () {
+      final existing = {
+        'p1': [_message('m1', parentId: 'p1', createdAt: DateTime(2024))],
+        'p2': [_message('m2', parentId: 'p2', createdAt: DateTime(2024))],
+      };
+      final incoming = [
+        _message('m3', parentId: 'p1', createdAt: DateTime(2024, 2)),
+        _message('m4', parentId: 'p2', createdAt: DateTime(2024, 2)),
+      ];
+
+      final result = MessageMerging.mergeThreadMessages(existing: existing, toMerge: incoming);
+
+      expect(_ids(result['p1']!), ['m1', 'm3']);
+      expect(_ids(result['p2']!), ['m2', 'm4']);
+    });
+
+    test('creates the thread entry for a reply to a new thread', () {
+      final result = MessageMerging.mergeThreadMessages(
+        existing: const {},
+        toMerge: [_message('m1', parentId: 'p1')],
+      );
+
+      expect(_ids(result['p1']!), ['m1']);
+    });
+
+    test('upsert: false does not create an entry for a thread that was never loaded', () {
+      final result = MessageMerging.mergeThreadMessages(
+        existing: const {},
+        toMerge: [_message('m1', parentId: 'p1')],
+        upsert: false,
+      );
+
+      expect(result, isEmpty);
+    });
+
+    test('upsert: false still updates a reply in a loaded thread', () {
+      final existing = {
+        'p1': [_message('m1', parentId: 'p1', text: 'old')],
+      };
+
+      final result = MessageMerging.mergeThreadMessages(
+        existing: existing,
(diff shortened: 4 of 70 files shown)


packages/stream_chat/lib/src/core/util/message_merging.dart
```diff
@@ -0,0 +1,301 @@
+import '../models/location.dart';
+import '../models/message.dart';
+import 'list_extensions.dart';
+import 'message_predicates.dart';
+
+/// Provides the merge and removal operations reconciling incoming messages
+/// with the locally-held channel state collections.
+///
+/// Every operation is pure: it computes a new collection from the given
+/// inputs without reading or writing any state.
```
xsahil03x: **The purity claim isn't true.**

Three of the nine operations read the wall clock: `mergePinnedMessages` / `removePinnedMessages` via `hasValidPin` → `expiresAt.isAfter(DateTime.now())`, and `mergeActiveLocations` via `Location.isExpired` → `isActive` → `endAt.isAfter(DateTime.now())` (`location.dart:87-95`).

What makes this more than a wording nit: all three return lazy `.where(...)` iterables, so the clock read happens when the caller drains the result, not when the operation is called. A doc promising purity is misleading about exactly the part that isn't.

```suggestion
/// No operation reads or writes channel state: each computes a new
/// collection from the given inputs.
///
/// The pin and live-location filters additionally consult the current time,
/// and the returned iterables are lazy — those filters run when the result is
/// iterated, not when the operation is called.
```

packages/stream_chat/lib/src/core/util/message_merging.dart
```diff
@@ -0,0 +1,301 @@
...
+class MessageMerging {
+  const MessageMerging._();
+
+  /// The default `update` strategy for the merge operations: merges the
+  /// incoming [updated] into the locally-known [original] via
+  /// [Message.updateWith], preserving enrichment the server may strip on
+  /// partial payloads.
```
xsahil03x: **First dartdoc paragraph should be one short sentence.**

STYLE_GUIDE: *"The first paragraph of any dartdoc section must be a short, self-contained sentence explaining the purpose of the item… Avoid having the first paragraph contain multiple sentences (it gets extracted for tables of contents)."* This is a four-line multi-clause sentence.

```suggestion
  /// The default `update` strategy for the merge operations.
  ///
  /// Merges the incoming [updated] into the locally-known [original] via
  /// [Message.updateWith], preserving enrichment the server may strip on
  /// partial payloads.
```

packages/stream_chat/lib/src/core/util/message_merging.dart
```diff
@@ -0,0 +1,301 @@
...
+  static Message mergeUpdate(Message original, Message updated) => original.updateWith(updated);
+
+  /// The replacing `update` strategy: takes the incoming [updated] as-is.
+  /// Used by local rollback paths.
+  static Message replaceUpdate(Message _, Message updated) => updated;
+
+  /// Compares [a] and [b] by their [Message.createdAt].
```
xsahil03x: **Doc you could have written from the name alone.**

*"Compares [a] and [b] by their [Message.createdAt]"* is the *"Avoid useless documentation"* case verbatim. Either drop it (nothing in `lib/src` requires it) or make it say something the signature doesn't — that this is the ordering every merge result comes back in, which is the useful fact.

packages/stream_chat/lib/src/core/util/message_merging.dart
```diff
@@ -0,0 +1,301 @@
...
+  }) {
+    return mergeMessages(
+      existing: existing,
+      toMerge: toMerge,
+      update: update,
+    ).where((it) => it.hasValidPin);
+  }
+
+  /// Merges [toMerge] into [existing], returning a list sorted by
+  /// [Message.createdAt].
```
xsahil03x: **It doesn't return a list, and it doesn't always sort.**

Return type is `Iterable<Message>`, and on `toMerge.isEmpty` it hands back `existing` untouched — unsorted if the caller's was. Same identity contract as the thread functions, worth stating here too.

```suggestion
  /// Merges [toMerge] into [existing].
  ///
  /// The result is ordered by [Message.createdAt]. Returns [existing]
  /// unchanged (same reference) when [toMerge] is empty, or when [upsert] is
  /// `false` and none of [toMerge] is already loaded.
```

packages/stream_chat/lib/src/core/util/message_merging.dart
```diff
@@ -0,0 +1,301 @@
...
+
+    final mergedById = {for (final m in mergedMessages) m.id: m};
+    return mergedMessages.updateIf(
+      (it) => deletedIds.contains(it.quotedMessageId),
+      (it) => it.copyWith(quotedMessage: mergedById[it.quotedMessageId]),
+    );
+  }
+
+  /// Merges [toMerge] into the [existing] threads map, returning the updated
+  /// map.
```
xsahil03x: **This doc line is load-bearing, and `d9cb260` removed it.**

That commit dropped `"or [existing] as-is when [toMerge] carries no thread replies"` from both thread-merge docs — and this same PR adds the call site that depends on precisely that, via `identical()` in `_updateThreadMessages`.

To be clear, the contract *is* locked by tests (`expect(result, same(existing))` exists for both), so this isn't an untested landmine. The problem is docs contradicting a tested, load-bearing contract. If someone later tidies the early return to `return {...existing}`, the tests do catch it — but they'll have read the doc first and concluded the identity return was incidental. Cost of the drift is a spurious `threadsStream` emission plus a debounced persistence write (`channel_client_state.dart:633-636`) on every batch that touches no thread.

`list_extensions.dart`, same directory, documents this exact kind of contract: *"Returns the receiver unchanged (same reference) when [other] is null, empty, or identical to this iterable."* Worth matching.

```suggestion
  /// Merges [toMerge] into the [existing] threads map, returning the updated
  /// map, or [existing] unchanged (same reference) when [toMerge] carries no
  /// thread replies.
```

packages/stream_chat/lib/src/core/util/message_merging.dart
```diff
@@ -0,0 +1,301 @@
...
+    required Iterable<Location> existing,
+    required Iterable<Message> toRemove,
+  }) {
+    if (toRemove.isEmpty) return existing;
+
+    final toRemoveIds = toRemove.map((m) => m.id).toSet();
+    final updatedLocations = existing.where(
+      // Remove the location if its attached message is in the toRemove list.
+      (it) => !toRemoveIds.contains(it.messageId),
+    );
```
xsahil03x: **Duplicate of the doc, and it costs a line wrap.**

The doc two lines up already says *"Removes from [existing] the locations attached to any message in [toRemove]."* The comment is the same sentence, and it's the only reason this call is three lines instead of one.

```suggestion
    final updatedLocations = existing.where((it) => !toRemoveIds.contains(it.messageId));
```

packages/stream_chat/lib/src/core/util/message_merging.dart
```diff
@@ -0,0 +1,301 @@
...
+          // Remove the message if it's in the toRemove list.
+          return !toRemoveIds.contains(it.id);
+        })
+        .map((it) {
+          // Continue if the message doesn't quote any of the deleted messages.
+          if (!toRemoveIds.contains(it.quotedMessageId)) return it;
+
+          // Setting it to null will remove the quoted message from the message.
+          return it.copyWith(quotedMessageId: null, quotedMessage: null);
+        });
```
xsahil03x: **A comment forcing a block body.**

`// Remove the message if it's in the toRemove list.` is the only thing keeping that `where` from being an arrow, and the two `.map` comments restate the doc's *"clearing the quoted-message reference of any remaining message that quotes a removed one."*

```suggestion
    final updatedMessages = existing
        .where((it) => !toRemoveIds.contains(it.id))
        .map((it) {
          if (!toRemoveIds.contains(it.quotedMessageId)) return it;
          return it.copyWith(quotedMessageId: null, quotedMessage: null);
        });
```

Worth a `melos run format` after — I wrote that by hand.

Two more of the same kind I didn't write suggestions for, since you'll spot them faster than I can quote them:

- **L180 and L273** — `// If there are no affected threads, return early.` twice, both directly above `if (…isEmpty) return existing;`.
- **L279, L289, L295** — `// Continue if the thread doesn't exist.`, `// If there are no more messages in the thread, remove the thread entry.`, `// Otherwise, update the thread with the modified message list.` The middle one is already in the doc as *"Thread entries left with no messages are dropped from the map."*

And to be explicit about what I'm **not** asking you to touch — these earn their place and should stay exactly as they are, because they explain *why* rather than *what*:

- L97–L100, the single-message fast-path rationale (`lastIndexWhere` + `sortedUpsertAt` skipping the keymap build)
- L117–L118, why a non-delete update can skip the quote rewrite
- L134–L136, the "receiver is maintained sorted as a state invariant" note
- L144–L147, the `updateIf` steady-state allocation note
- L172–L174, the cross-thread reply-leak explanation
- L187–L190, the phantom-entry guard

packages/stream_chat/lib/src/core/util/message_merging.dart
```diff
@@ -0,0 +1,301 @@
...
+
+    return updatedMessages;
+  }
+
+  /// Removes [toRemove] from the [existing] threads map, returning the
+  /// updated map.
```
xsahil03x: **Same as #5, removal side.**

```suggestion
  /// Removes [toRemove] from the [existing] threads map, returning the
  /// updated map, or [existing] unchanged (same reference) when [toRemove]
  /// carries no thread replies.
```

packages/stream_chat/lib/src/core/util/message_predicates.dart
```diff
@@ -0,0 +1,35 @@
+import '../models/message.dart';
+
+/// Predicates over a [Message] used by the channel state management.
```
xsahil03x: **Say why this isn't in `message.dart`.**

Minor, but `message.dart` already exports a `MessageVisibility` extension that is about message visibility, and `isShownInChannel` is also about message visibility while living somewhere else and not being exported. Someone who finds `MessageVisibility` will reasonably assume it's the whole story. I think the placement is right — these encode channel-state rules, not properties of the message — but the doc should say so.

```suggestion
/// Predicates over a [Message] used by the channel state management.
///
/// Kept out of the exported extensions in `message.dart` on purpose: these
/// encode channel-state rules rather than properties of the message itself.
```

packages/stream_chat/lib/src/core/util/message_predicates.dart
```diff
@@ -0,0 +1,35 @@
...
+    // If the message is not pinned, it's not valid.
+    if (pinned != true) return false;
+
+    // If there's no expiration, the pin is valid.
+    final expiresAt = pinExpires;
+    if (expiresAt == null) return true;
+
+    // If there's an expiration, check if it's still valid.
+    return expiresAt.isAfter(DateTime.now());
+  }
```
xsahil03x: **The new docs and the old comments say the same thing twice.**

35 lines, 15 of them comments, for two getters that are three and six lines of actual code. Every `//` in here restates the `///` directly above it:

| doc says | comment says |
|---|---|
| "Non-thread messages always are" | `// Non-thread messages are always shown in the channel.` |
| "thread replies only when explicitly marked" | `// Thread messages are only shown if explicitly marked.` |
| "Returns `false` if the message is deleted…" | `// If the message is deleted, the pin is not valid.` |
| "…not pinned…" | `// If the message is not pinned, it's not valid.` |
| "…or its [Message.pinExpires] has passed" | `// If there's no expiration, the pin is valid.` + `// If there's an expiration, check if it's still valid.` |

I know the `//` half is pre-existing — it rode along from `_isShownInChannel` / `_pinIsValid`, and the guide is clear that pre-existing comments shouldn't be swept up in unrelated changes. But this PR authored the `///` half, which is what created the duplication, so cleaning it up is in scope. `hasValidPin` as it stands is close to the literal BAD example under *"Minimize inline comments"* — four comments narrating four consecutive guard clauses. The doc is the better half; keep it and drop the narration:

```suggestion
  bool get isShownInChannel {
    if (parentId == null) return true;
    return showInChannel == true;
  }

  /// Whether the message represents a currently valid pin.
  ///
  /// Returns `false` if the message is deleted, not pinned, or its
  /// [Message.pinExpires] has passed.
  bool get hasValidPin {
    if (isDeleted) return false;
    if (pinned != true) return false;

    final expiresAt = pinExpires;
    if (expiresAt == null) return true;

    return expiresAt.isAfter(DateTime.now());
  }
```

packages/stream_chat/test/src/core/util/message_merging_test.dart
```diff
@@ -0,0 +1,489 @@
...
+  String id, {
+  DateTime? createdAt,
+  String? text,
+  String? parentId,
+  bool? showInChannel,
+  String? quotedMessageId,
+  Message? quotedMessage,
+  bool pinned = false,
+  DateTime? pinExpires,
+  String type = 'regular',
```
xsahil03x: **Magic string where a constant exists.**

`MessageType.regular` is what `Message`'s own default uses (`message.dart:36`), and the tests already reach for `MessageType.deleted`.

```suggestion
  String type = MessageType.regular,
```

xsahil03x (commented):
_Superseded — this review's comments were reposted together in [the review below](https://github.com/GetStream/stream-chat-flutter/pull/2944#pullrequestreview-5200142216) so they sit under one entry. Nothing here changed in substance._

packages/stream_chat/test/src/core/util/message_merging_test.dart
```diff
@@ -0,0 +1,489 @@
...
+    endAt: endAt,
+    latitude: latitude,
+    longitude: longitude,
+  );
+}
+
+Iterable<String?> _ids(Iterable<Message> messages) => messages.map((it) => it.id);
+
+void main() {
+  group('MessageMerging.mergeUpdate', () {
```
xsahil03x: **`group` used to organise the file by method.**

TESTING.md is explicit: *"Do not use `group` to organize a file by method or class — that's what the file itself is for. If a `group` is doing the work a separate file should be doing, split the file instead."* This is 489 lines with eleven groups, every one named `MessageMerging.<method>`.

The individual test names are good — behaviour-first, one behaviour each — so this is purely about the containers. Two ways out, your pick:

- Split into `message_merging_messages_test.dart` / `_threads_test.dart` / `_pinned_test.dart` / `_locations_test.dart`. Also satisfies *"Prefer more test files, avoid long test files"*.
- Keep one file and rename groups to the preconditions they actually cluster on. The `upsert: false` cases are a genuine shared precondition and would group well:

```dart
group('when the message is outside the loaded window', () {
  test('skips a single message that is not loaded', () { ... });
  test('only applies the batch entries that are already loaded', () { ... });
  test('does not create an entry for a thread that was never loaded', () { ... });
});
```

I'd lean toward the split, but either satisfies the guide.

packages/stream_chat/test/src/core/util/message_merging_test.dart
```diff
@@ -0,0 +1,489 @@
...
+
+    test('upsert: false does not create an entry for a thread that was never loaded', () {
+      final result = MessageMerging.mergeThreadMessages(
+        existing: const {},
+        toMerge: [_message('m1', parentId: 'p1')],
+        upsert: false,
+      );
+
+      expect(result, isEmpty);
+    });
```
xsahil03x: **This test can't fail the way you need it to.**

It passes `existing: const {}` and asserts only `isEmpty`, so it passes whether or not a fresh map comes back. But the fresh-map return is exactly what preserves the old write behaviour in row 3 of the table in my summary — if `mergeThreadMessages` started returning `existing` by identity here, `_updateThreadMessages` would silently stop emitting on `threadsStream` and stop scheduling the persistence write, and nothing in this file would notice.

Adding the non-empty case pins it down:

```suggestion
    test('upsert: false does not create an entry for a thread that was never loaded', () {
      final result = MessageMerging.mergeThreadMessages(
        existing: const {},
        toMerge: [_message('m1', parentId: 'p1')],
        upsert: false,
      );

      expect(result, isEmpty);
    });

    test('upsert: false returns a new map when a reply targeted an unloaded thread', () {
      final existing = {
        'p1': [_message('m1', parentId: 'p1')],
      };

      final result = MessageMerging.mergeThreadMessages(
        existing: existing,
        toMerge: [_message('m2', parentId: 'p2')],
        upsert: false,
      );

      expect(result.keys, ['p1']);
      expect(result, isNot(same(existing)), reason: 'the caller writes back whenever a thread was targeted');
    });
```

Three more gaps I noticed, worth one pass if you agree: duplicate ids in `existing` (`lastIndexWhere` deliberately takes the last), the single-message path when `createdAt` changes and the message re-sorts, and a custom `update` on `mergePinnedMessages`.

packages/stream_chat/test/src/core/util/message_predicates_test.dart
```diff
@@ -0,0 +1,66 @@
...
+import 'package:stream_chat/stream_chat.dart';
+import 'package:test/test.dart';
+
+Message _message(
+  String id, {
+  String? parentId,
+  bool? showInChannel,
+  bool pinned = false,
+  DateTime? pinExpires,
+  String type = 'regular',
```
xsahil03x: **Same magic string as #11.**

```suggestion
  String type = MessageType.regular,
```

packages/stream_chat/test/src/core/util/message_predicates_test.dart
```diff
@@ -0,0 +1,66 @@
...
+      final message = _message('m1', pinned: true, pinExpires: DateTime.now().add(const Duration(hours: 1)));
+      expect(message.hasValidPin, isTrue);
+    });
+
+    test('is false once the pin expiration has passed', () {
+      final message = _message('m1', pinned: true, pinExpires: DateTime.now().subtract(const Duration(hours: 1)));
+      expect(message.hasValidPin, isFalse);
+    });
+  });
+}
```
xsahil03x: **Same `group` issue as #12, small enough to just show.**

Two groups for eight short tests; folding the subject into the test names reads better in `dart test` output and drops a level of nesting.

```suggestion
void main() {
  test('isShownInChannel is true for a non-thread message', () {
    expect(_message('m1').isShownInChannel, isTrue);
  });

  test('isShownInChannel is true for a thread reply marked to show in the channel', () {
    final reply = _message('m1', parentId: 'p1', showInChannel: true);
    expect(reply.isShownInChannel, isTrue);
  });

  test('isShownInChannel is false for a thread-only reply', () {
    final reply = _message('m1', parentId: 'p1');
    expect(reply.isShownInChannel, isFalse);
  });

  test('hasValidPin is false for a deleted message', () {
    final message = _message('m1', pinned: true, type: MessageType.deleted);
    expect(message.hasValidPin, isFalse);
  });

  test('hasValidPin is false for an unpinned message', () {
    final message = _message('m1');
    expect(message.hasValidPin, isFalse);
  });

  test('hasValidPin is true for a pinned message without expiration', () {
    final message = _message('m1', pinned: true);
    expect(message.hasValidPin, isTrue);
  });

  test('hasValidPin is true while the pin expiration is in the future', () {
    final message = _message('m1', pinned: true, pinExpires: DateTime.now().add(const Duration(hours: 1)));
    expect(message.hasValidPin, isTrue);
  });

  test('hasValidPin is false once the pin expiration has passed', () {
    final message = _message('m1', pinned: true, pinExpires: DateTime.now().subtract(const Duration(hours: 1)));
    expect(message.hasValidPin, isFalse);
  });
}
```

xsahil03x (commented):
Folded into my review above — see the updated comments on `message_merging.dart` L14–17 and L75–76. Nothing new here.

packages/stream_chat/lib/src/core/util/message_merging.dart
```diff
@@ -0,0 +1,301 @@
...
+
+  /// The default `update` strategy for the merge operations: merges the
+  /// incoming [updated] into the locally-known [original] via
+  /// [Message.updateWith], preserving enrichment the server may strip on
+  /// partial payloads.
+  static Message mergeUpdate(Message original, Message updated) => original.updateWith(updated);
+
+  /// The replacing `update` strategy: takes the incoming [updated] as-is.
+  /// Used by local rollback paths.
+  static Message replaceUpdate(Message _, Message updated) => updated;
```
xsahil03x: **Supersedes my suggestion on L14–L17 — take this one instead.** It folds in the first-paragraph split I asked for there.

L16–17's "preserving enrichment the server may strip on partial payloads" describes server response behavior; the contract is just that absent fields fall back. `ChannelClientState.updateMessage` L228–231 already carries the rationale and names the fields.

L21's "Used by local rollback paths" is a caller inventory, and `replaceMessage` L239–244 already documents that path properly.

(Note the first parameter is a wildcard `_`, so `replaceUpdate`'s doc can't reference it by name.)

```suggestion
  /// The default `update` strategy for the merge operations.
  ///
  /// Merges the incoming [updated] into the locally-known [original] via
  /// [Message.updateWith]: fields absent from [updated] keep their value
  /// from [original].
  static Message mergeUpdate(Message original, Message updated) => original.updateWith(updated);

  /// The replacing `update` strategy: takes the incoming [updated] as-is,
  /// ignoring the locally-known message.
  static Message replaceUpdate(Message _, Message updated) => updated;
```

packages/stream_chat/lib/src/core/util/message_merging.dart
```diff
@@ -0,0 +1,301 @@
...
+  /// [Message.createdAt].
+  ///
+  /// [update] decides whether each pair is reconciled (default — see
+  /// [mergeUpdate]) or replaced ([replaceUpdate], used by local rollback
+  /// paths that don't want enrichment fallback to keep optimistic values).
+  ///
+  /// [upsert] controls whether ids not already in [existing] are inserted.
+  /// Event-driven paths (`message.updated`, `message.deleted` soft) pass
+  /// `upsert: false` so an out-of-window message isn't dropped into a gap
+  /// between the loaded slice and history the client hasn't paged in yet.
```
xsahil03x: **Supersedes my suggestion on L75–L76 — take this one instead.** Same summary rewrite, plus the param docs, as one block so the two don't fight over shifting line numbers.

This is the clearest instance. L83–85 names `message.updated` and `message.deleted` (server WS event types) and then describes channel pagination — "out-of-window", "the loaded slice and history the client hasn't paged in yet" — in the doc for a function that takes two `Iterable<Message>` and knows nothing about either. L79–80 repeats the rollback rationale from L21.

Both parameters stay documented; only the caller inventory and the pagination prose go. L82 was already the contract sentence and is kept verbatim.

```suggestion
  /// Merges [toMerge] into [existing].
  ///
  /// The result is ordered by [Message.createdAt]. Returns [existing]
  /// unchanged (same reference) when [toMerge] is empty, or when [upsert] is
  /// `false` and none of [toMerge] is already loaded.
  ///
  /// [update] reconciles a pair whose id is already in [existing]; defaults
  /// to [mergeUpdate], with [replaceUpdate] for a strict overwrite.
  ///
  /// [upsert] controls whether ids not already in [existing] are inserted.
```

packages/stream_chat/lib/src/core/util/message_merging.dart
```diff
@@ -0,0 +1,301 @@
...
+  const MessageMerging._();
+
+  /// The default `update` strategy for the merge operations: merges the
+  /// incoming [updated] into the locally-known [original] via
+  /// [Message.updateWith], preserving enrichment the server may strip on
+  /// partial payloads.
+  static Message mergeUpdate(Message original, Message updated) => original.updateWith(updated);
+
+  /// The replacing `update` strategy: takes the incoming [updated] as-is.
+  /// Used by local rollback paths.
```
xsahil03x: **Caller inventory, not contract.**

"Used by local rollback paths." tells the reader who calls this, not what it does. `ChannelClientState.replaceMessage` L239–244 already documents that path in full — *"Useful for local rollbacks of an optimistic update, where the caller has the full prior snapshot and doesn't want the merge falling back to the optimistic values."* Nothing is lost by dropping it here.

```suggestion
  /// The replacing `update` strategy: takes the incoming [updated] as-is,
  /// ignoring the locally-known message.
```

(The first parameter is a wildcard `_`, so the doc can't reference it by name.)

packages/stream_chat/lib/src/core/util/message_merging.dart
xsahil03x: **The clearest leak in the file.**

L83–85 names `message.updated` and `message.deleted` — server WS event types — then describes the channel's pagination window: "out-of-window", "the loaded slice and history the client hasn't paged in yet". This function takes two `Iterable<Message>` and knows about neither. L79–80 repeats the rollback rationale from L21.

It lands harder here than elsewhere because of your own reason for the file's location: these sit in `core/util/` rather than beside `channel_client_state.dart` *because they're channel-agnostic*. Either the module is, or the docs are.

Nothing is lost by cutting — `ChannelClientState.updateMessage` L233–236 already states the `upsert` contract without either. Both params stay documented:

```suggestion
  /// [update] reconciles a pair whose id is already in [existing]; defaults
  /// to [mergeUpdate], with [replaceUpdate] for a strict overwrite.
  ///
  /// [upsert] controls whether ids not already in [existing] are inserted.
```

packages/stream_chat/lib/src/core/util/message_merging.dart
```diff
@@ -0,0 +1,301 @@
...
+      (it) => it.copyWith(quotedMessage: mergedById[it.quotedMessageId]),
+    );
+  }
+
+  /// Merges [toMerge] into the [existing] threads map, returning the updated
+  /// map.
+  ///
+  /// Replies are grouped by their parent id so each thread merge only sees
+  /// its own messages. With [upsert] `false`, replies to threads not present
+  /// in [existing] are dropped instead of creating the thread entry.
```
xsahil03x: **Describes the algorithm rather than the guarantee.**

"Replies are grouped by their parent id so each thread merge only sees its own messages" is how the function is built — the grouping pass, and the fact that it runs one merge per thread. What a caller needs to know is that a reply lands in its own parent's list and nowhere else.

The `[upsert]` sentence after it is already contract, so it survives verbatim:

```suggestion
  /// Each reply is merged into the list for its [Message.parentId], leaving
  /// other threads untouched. With [upsert] `false`, replies to threads not
  /// present in [existing] are dropped instead of creating the thread entry.
```

packages/stream_chat/lib/src/core/util/message_merging.dart
xsahil03x: **Two things in one block: a four-line opening paragraph, and detail that's already documented upstream.**

STYLE_GUIDE: *"The first paragraph of any dartdoc section must be a short, self-contained sentence explaining the purpose of the item… Avoid having the first paragraph contain multiple sentences (it gets extracted for tables of contents)."* L14–17 is a single four-line multi-clause sentence.

The leaks — this corrects the "clean" verdict I originally gave in the review body:

- **L16–17** — "preserving enrichment the server may strip on partial payloads" describes server response behavior, not what the function guarantees. `ChannelClientState.updateMessage` L228–231 already carries that rationale, and names the actual fields (`poll`, `sharedLocation`, `ownReactions`, nested `quotedMessage`).
- **L21** — "Used by local rollback paths." is a caller inventory. `ChannelClientState.replaceMessage` L239–244 already documents that path in full.

Nothing is lost by cutting either; the surviving copy is the more specific one.

```suggestion
  /// The default `update` strategy for the merge operations.
  ///
  /// Merges the incoming [updated] into the locally-known [original] via
  /// [Message.updateWith]: fields absent from [updated] keep their value
  /// from [original].
  static Message mergeUpdate(Message original, Message updated) => original.updateWith(updated);

  /// The replacing `update` strategy: takes the incoming [updated] as-is,
  /// ignoring the locally-known message.
  static Message replaceUpdate(Message _, Message updated) => updated;
```

(`replaceUpdate`'s first parameter is a wildcard `_`, so its doc can't reference it by name.)

packages/stream_chat/lib/src/core/util/message_merging.dart
xsahil03x: **The summary overstates the return, and the param docs are the clearest leak in the file.**

**L75–76** says "returning a list sorted by [Message.createdAt]", but the return type is `Iterable<Message>`, and on `toMerge.isEmpty` it hands back `existing` untouched — unsorted if the caller's was. Worth stating the identity return here too, same as the thread functions.

**L83–85** names `message.updated` and `message.deleted` — server WS event types — then describes the channel's pagination window: "out-of-window", "the loaded slice and history the client hasn't paged in yet". This function takes two `Iterable<Message>` and knows about neither. **L79–80** repeats the rollback rationale from L21.

It lands harder here than elsewhere because of your own reason for the file's location: these sit in `core/util/` rather than beside `channel_client_state.dart` *because they're channel-agnostic*. Either the module is, or the docs are.

And nothing is lost by cutting — `ChannelClientState.updateMessage` L233–236 already states the `upsert` contract without either. Both params stay documented:

```suggestion
  /// Merges [toMerge] into [existing].
  ///
  /// The result is ordered by [Message.createdAt]. Returns [existing]
  /// unchanged (same reference) when [toMerge] is empty, or when [upsert] is
  /// `false` and none of [toMerge] is already loaded.
  ///
  /// [update] reconciles a pair whose id is already in [existing]; defaults
  /// to [mergeUpdate], with [replaceUpdate] for a strict overwrite.
  ///
  /// [upsert] controls whether ids not already in [existing] are inserted.
```

packages/stream_chat/lib/src/core/util/message_merging.dart
xsahil03x: **A deleted contract line that this PR made load-bearing, and an algorithm description where the guarantee should be.**

**L158–159** — `d9cb260` dropped `"or [existing] as-is when [toMerge] carries no thread replies"`, and this same PR adds the call site that depends on precisely that, via `identical()` in `_updateThreadMessages`.

To be clear, the contract *is* locked by tests (`expect(result, same(existing))` exists for both thread functions), so this isn't an untested landmine. The problem is docs contradicting a tested, load-bearing contract: someone tidying the early return to `return {...existing}` reads the doc first and concludes the identity return was incidental. Cost of the drift is a spurious `threadsStream` emission plus a debounced persistence write (`channel_client_state.dart:633-636`) on every batch that touches no thread. `list_extensions.dart`, same directory, documents this exact kind of contract: *"Returns the receiver unchanged (same reference) when [other] is null, empty, or identical to this iterable."*

**L161–162** — "Replies are grouped by their parent id so each thread merge only sees its own messages" is how the function is built: the grouping pass, and the fact that it runs one merge per thread. What a caller needs is that a reply lands in its own parent's list and nowhere else. The `[upsert]` sentence is already contract and survives verbatim.

```suggestion
  /// Merges [toMerge] into the [existing] threads map, returning the updated
  /// map, or [existing] unchanged (same reference) when [toMerge] carries no
  /// thread replies.
  ///
  /// Each reply is merged into the list for its [Message.parentId], leaving
  /// other threads untouched. With [upsert] `false`, replies to threads not
  /// present in [existing] are dropped instead of creating the thread entry.
```

packages/stream_chat/lib/src/core/util/message_merging.dart
xsahil03x: **"Every operation is pure" isn't true — and my first replacement here traded one leak for another.**

Three of the nine operations read the wall clock: `mergePinnedMessages` / `removePinnedMessages` via `hasValidPin` → `expiresAt.isAfter(DateTime.now())`, and `mergeActiveLocations` via `Location.isExpired` → `isActive` → `endAt.isAfter(DateTime.now())` (`location.dart:87-95`). So the claim is wrong on its face.

But the version I first suggested described *when evaluation happens* — "resolved when the returned collection is read" — which is the same category of thing the rule is about: *"describe what values are produced and when, not the … mechanics."* Lazy-vs-eager is how it's built. What a caller needs is that inputs aren't mutated, and that pin and expiry results move with the clock:

```suggestion
/// Operations leave their inputs untouched and return new collections. Pin
/// validity and live-location expiry are relative to the current time.
```

For what it's worth the laziness never escapes this class — every call site in `channel_client_state.dart` materialises immediately (`.toList()` at L659, L921, L935, L986, L1000, L1014). So there's nothing for a caller to trip over and nothing to document.

packages/stream_chat/lib/src/core/util/message_merging.dart
xsahil03x: **Two things in one block: a four-line opening paragraph, and detail that's already documented upstream.**

STYLE_GUIDE: *"The first paragraph of any dartdoc section must be a short, self-contained sentence explaining the purpose of the item… Avoid having the first paragraph contain multiple sentences (it gets extracted for tables of contents)."* L14–17 is a single four-line multi-clause sentence.

The leaks — this corrects the "clean" verdict I originally gave in the review body:

- **L16–17** — "preserving enrichment the server may strip on partial payloads" describes server response behavior, not what the function guarantees. `ChannelClientState.updateMessage` L228–231 already carries that rationale, and names the actual fields (`poll`, `sharedLocation`, `ownReactions`, nested `quotedMessage`).
- **L21** — "Used by local rollback paths." is a caller inventory. `ChannelClientState.replaceMessage` L239–244 already documents that path in full.

Nothing is lost by cutting either; the surviving copy is the more specific one.

```suggestion
  /// The default `update` strategy for the merge operations.
  ///
  /// Merges the incoming [updated] into the locally-known [original] via
  /// [Message.updateWith]: fields absent from [updated] keep their value
  /// from [original].
  static Message mergeUpdate(Message original, Message updated) => original.updateWith(updated);

  /// The replacing `update` strategy: takes the incoming [updated] as-is,
  /// ignoring the locally-known message.
  static Message replaceUpdate(Message _, Message updated) => updated;
```

(`replaceUpdate`'s first parameter is a wildcard `_`, so its doc can't reference it by name.)

packages/stream_chat/lib/src/core/util/message_merging.dart
xsahil03x: **Doc you could have written from the name alone.**

*"Compares [a] and [b] by their [Message.createdAt]"* restates the signature — the *"Avoid useless documentation"* case. The fact worth writing down is that this is the canonical ordering, not just a comparator that happens to exist.

Effective Dart also wants function docs to open with a third-person verb, which a bare noun phrase wouldn't:

```suggestion
  /// Orders messages by [Message.createdAt], the order every merge result is
  /// returned in.
```

packages/stream_chat/lib/src/core/util/message_merging.dart
xsahil03x: **The summary overstates the return, and the param docs are the clearest leak in the file.**

**L75–76** says "returning a list sorted by [Message.createdAt]", but the return type is `Iterable<Message>`, and on `toMerge.isEmpty` it hands back `existing` untouched — unsorted if the caller's was. Worth stating the identity return here too, same as the thread functions.

**L83–85** names `message.updated` and `message.deleted` — server WS event types — then describes the channel's pagination window: "out-of-window", "the loaded slice and history the client hasn't paged in yet". This function takes two `Iterable<Message>` and knows about neither. **L79–80** repeats the rollback rationale from L21.

It lands harder here than elsewhere because of your own reason for the file's location: these sit in `core/util/` rather than beside `channel_client_state.dart` *because they're channel-agnostic*. Either the module is, or the docs are.

And nothing is lost by cutting — `ChannelClientState.updateMessage` L233–236 already states the `upsert` contract without either. Both params stay documented, in terms of the two collections the function actually receives:

```suggestion
  /// Merges [toMerge] into [existing].
  ///
  /// The result is ordered by [Message.createdAt]. Returns [existing]
  /// unchanged (same reference) when [toMerge] is empty, or when [upsert] is
  /// `false` and no id in [toMerge] is present in [existing].
  ///
  /// [update] reconciles a pair whose id is already in [existing]; defaults
  /// to [mergeUpdate], with [replaceUpdate] for a strict overwrite.
  ///
  /// [upsert] controls whether ids not already in [existing] are inserted.
```

packages/stream_chat/lib/src/core/util/message_merging.dart
xsahil03x: **A deleted contract line that this PR made load-bearing, and an algorithm description where the guarantee should be.**

**L158–159** — `d9cb260` dropped `"or [existing] as-is when [toMerge] carries no thread replies"`, and this same PR adds the call site that depends on precisely that, via `identical()` in `_updateThreadMessages`.

To be clear, the contract *is* locked by tests (`expect(result, same(existing))` exists for both thread functions), so this isn't an untested landmine. The problem is docs contradicting a tested, load-bearing contract: someone tidying the early return to `return {...existing}` reads the doc first and concludes the identity return was incidental. Cost of the drift is a spurious `threadsStream` emission plus a debounced persistence write (`channel_client_state.dart:633-636`) on every batch that touches no thread. `list_extensions.dart`, same directory, documents this exact kind of contract: *"Returns the receiver unchanged (same reference) when [other] is null, empty, or identical to this iterable."*

**L161–162** — "Replies are grouped by their parent id so each thread merge only sees its own messages" is how the function is built: the grouping pass, and the fact that it runs one merge per thread. What a caller needs is that a reply lands in its own parent's list and nowhere else. The `[upsert]` sentence is already contract and survives verbatim.

```suggestion
  /// Merges [toMerge] into the [existing] threads map, returning the updated
  /// map, or [existing] unchanged (same reference) when [toMerge] carries no
  /// thread replies.
  ///
  /// Each reply is merged into the list for its [Message.parentId], leaving
  /// other threads untouched. With [upsert] `false`, replies to threads not
  /// present in [existing] are dropped instead of creating the thread entry.
```

packages/stream_chat/lib/src/core/util/message_merging.dart
xsahil03x: **Same as #5, removal side.**

```suggestion
  /// Removes [toRemove] from the [existing] threads map, returning the
  /// updated map, or [existing] unchanged (same reference) when [toRemove]
  /// carries no thread replies.
```

xsahil03x (commented):
I reviewed the five files this PR adds on top of #2943 (`git diff origin/refactor/FLU-723_extract_channel_event_handler..HEAD`) — `channel_event_handler.dart` and friends belong to the other PR, so I left them alone. CI is green on the branch; I didn't run the suite locally, so everything below comes from reading the before/after rather than executing it.

The extraction reads well and `MessageRules` is the right precedent to follow. One thing worth calling out because the description doesn't: eight of the nine helpers did move verbatim, but the **thread orchestrators didn't** — `_updateThreadMessages` and `_removeThreadMessages` gained an `identical()` guard and a restructured early return. That's the only place in the diff where behavior could have shifted, so I traced it. It holds in all four cases:

| | old | new |
|---|---|---|
| `messages.isEmpty` | early return, no write | same guard, still first |
| nothing targets a thread | `messagesByThread.isEmpty` → no write | returns `existing` by identity → `identical()` → no write |
| threads targeted, all skipped by the phantom guard | wrote `{...threads}` | returns fresh `{...existing}` → not identical → writes |
| normal merge | writes | writes |

Capturing `final currentThreads = threads;` into a local before the call was right — `threads` is a getter, and comparing against a second invocation would have been fragile.

**Two I'd like fixed before merge**, both doc-only: the class doc claims purity that three operations don't have, and `d9cb260` deleted a doc line that this PR made load-bearing. **Then** the test-file organisation, one test that can't fail the way it needs to, and a batch of terseness nits. Nothing blocks.

On the specific things I was asked to check:

- **No internal-impl / server-side detail in public docs** — four lines want trimming, flagged on L14–17 and L75–76. `mergeUpdate` L16–17, `replaceUpdate` L21 and `mergeMessages` L79–80 / L83–85 describe server response behavior, WS event type names and the channel's pagination window rather than what these functions guarantee. Each already exists upstream at the seam and better written (`ChannelClientState.updateMessage` L228–236, `replaceMessage` L239–244), so the ask is delete, not relocate. Worth weighing against your own reason for the file's location: `core/util/` rather than beside `channel_client_state.dart` *because they're channel-agnostic*. Either the module is, or the docs are.
- **`///` vs `//`** — correct. That rule targets `_`-prefixed members; these are public members of an unexported class, same as `MessageRules`.
- **Effective Dart** — largely clean: third-person verbs, `Whether…` on boolean getters, `[]` for params that are also properties. Two hits, flagged inline.
- **TESTING.md** — one citable violation, flagged inline.
- **Terseness** — density is fine and I checked rather than eyeballed: `message_merging.dart` is 47 doc + 36 inline comment lines over 301, against `message_rules.dart` at 39 + 27 over 209. Same ratio. What isn't fine is **duplication** — the new `///` docs restate `//` comments that rode along verbatim from `channel_client_state.dart`, so several blocks now say the same thing twice. The `//` half is pre-existing and I'd normally leave it alone, but this PR authored the `///` half, which is what created the overlap. Specific pairs flagged inline, along with an explicit list of the comments I think should stay.

packages/stream_chat/lib/src/core/util/message_predicates.dart
xsahil03x, line 3: **Say what these are, not where they aren't.**

`message.dart` already exports a `MessageVisibility` extension that is about message visibility, and `isShownInChannel` is also about message visibility while living somewhere else. Someone who finds `MessageVisibility` will reasonably assume it's the whole story, so a line distinguishing the two earns its place.

Keeping it to what the predicates *are*. My first version here ended with "…which is why they live apart from the extensions on [Message]" — that's a note about how the codebase is arranged, and it goes stale the moment either file moves. The distinction itself does the work without it:

```suggestion
/// Predicates over a [Message] used by the channel state management.
///
/// These encode channel-state rules rather than properties of the message
/// itself.
```

packages/stream_chat/test/src/core/util/message_merging_test.dart
xsahil03x: **Magic string where a constant exists.**

`MessageType.regular` is what `Message`'s own default uses (`message.dart:36`), and the tests already reach for `MessageType.deleted`.

```suggestion
  String type = MessageType.regular,
```

packages/stream_chat/test/src/core/util/message_merging_test.dart
xsahil03x: **`group` used to organise the file by method.**

TESTING.md is explicit: *"Do not use `group` to organize a file by method or class — that's what the file itself is for. If a `group` is doing the work a separate file should be doing, split the file instead."* This is 489 lines with eleven groups, every one named `MessageMerging.<method>`.

The individual test names are good — behaviour-first, one behaviour each — so this is purely about the containers. Two ways out, your pick:

- Split into `message_merging_messages_test.dart` / `_threads_test.dart` / `_pinned_test.dart` / `_locations_test.dart`. Also satisfies *"Prefer more test files, avoid long test files"*.
- Keep one file and rename groups to the preconditions they actually cluster on. The `upsert: false` cases are a genuine shared precondition and would group well:

```dart
group('when the message is outside the loaded window', () {
  test('skips a single message that is not loaded', () { ... });
  test('only applies the batch entries that are already loaded', () { ... });
  test('does not create an entry for a thread that was never loaded', () { ... });
});
```

I'd lean toward the split, but either satisfies the guide.

packages/stream_chat/test/src/core/util/message_merging_test.dart
xsahil03x: **This test can't fail the way you need it to.**

It passes `existing: const {}` and asserts only `isEmpty`, so it passes whether or not a fresh map comes back. But the fresh-map return is exactly what preserves the old write behaviour in row 3 of the table in my summary — if `mergeThreadMessages` started returning `existing` by identity here, `_updateThreadMessages` would silently stop emitting on `threadsStream` and stop scheduling the persistence write, and nothing in this file would notice.

Adding the non-empty case pins it down:

```suggestion
    test('upsert: false does not create an entry for a thread that was never loaded', () {
      final result = MessageMerging.mergeThreadMessages(
        existing: const {},
        toMerge: [_message('m1', parentId: 'p1')],
        upsert: false,
      );

      expect(result, isEmpty);
    });

    test('upsert: false returns a new map when a reply targeted an unloaded thread', () {
      final existing = {
        'p1': [_message('m1', parentId: 'p1')],
      };

      final result = MessageMerging.mergeThreadMessages(
        existing: existing,
        toMerge: [_message('m2', parentId: 'p2')],
        upsert: false,
      );

      expect(result.keys, ['p1']);
      expect(result, isNot(same(existing)), reason: 'the caller writes back whenever a thread was targeted');
    });
```

Three more gaps I noticed, worth one pass if you agree: duplicate ids in `existing` (`lastIndexWhere` deliberately takes the last), the single-message path when `createdAt` changes and the message re-sorts, and a custom `update` on `mergePinnedMessages`.

packages/stream_chat/test/src/core/util/message_predicates_test.dart
xsahil03x: **Same magic string as #11.**

```suggestion
  String type = MessageType.regular,
```

packages/stream_chat/test/src/core/util/message_predicates_test.dart
xsahil03x, line 62: **Same `group` issue as #12, small enough to just show.**

Two groups for eight short tests; folding the subject into the test names reads better in `dart test` output and drops a level of nesting.

```suggestion
void main() {
  test('isShownInChannel is true for a non-thread message', () {
    expect(_message('m1').isShownInChannel, isTrue);
  });

  test('isShownInChannel is true for a thread reply marked to show in the channel', () {
    final reply = _message('m1', parentId: 'p1', showInChannel: true);
    expect(reply.isShownInChannel, isTrue);
  });

  test('isShownInChannel is false for a thread-only reply', () {
    final reply = _message('m1', parentId: 'p1');
    expect(reply.isShownInChannel, isFalse);
  });

  test('hasValidPin is false for a deleted message', () {
    final message = _message('m1', pinned: true, type: MessageType.deleted);
    expect(message.hasValidPin, isFalse);
  });

  test('hasValidPin is false for an unpinned message', () {
    final message = _message('m1');
    expect(message.hasValidPin, isFalse);
  });

  test('hasValidPin is true for a pinned message without expiration', () {
    final message = _message('m1', pinned: true);
    expect(message.hasValidPin, isTrue);
  });

  test('hasValidPin is true while the pin expiration is in the future', () {
    final message = _message('m1', pinned: true, pinExpires: DateTime.now().add(const Duration(hours: 1)));
    expect(message.hasValidPin, isTrue);
  });

  test('hasValidPin is false once the pin expiration has passed', () {
    final message = _message('m1', pinned: true, pinExpires: DateTime.now().subtract(const Duration(hours: 1)));
    expect(message.hasValidPin, isFalse);
  });
}
```

packages/stream_chat/lib/src/core/util/message_merging.dart
VelikovPetar: Addressed in 8e68896 — took your wording: the class doc now says operations leave their inputs untouched, and that pin validity and live-location expiry are relative to the current time.

packages/stream_chat/lib/src/core/util/message_merging.dart
VelikovPetar: Addressed in 8e68896 — `mergeUpdate` now opens with a one-sentence summary and drops the server-payload rationale; `replaceUpdate` drops the caller inventory. Used your text for both.

packages/stream_chat/lib/src/core/util/message_merging.dart
VelikovPetar: Addressed in 8e68896 — the doc now names this as the order every merge result comes back in, instead of restating the signature.

packages/stream_chat/lib/src/core/util/message_merging.dart
VelikovPetar: Addressed in 8e68896 — the summary no longer promises a sorted list, the identity return is spelled out for both the empty and `upsert: false` cases, and the WS event names and pagination-window prose are gone.

packages/stream_chat/lib/src/core/util/message_merging.dart
VelikovPetar: Addressed in 8e68896 — the identity-return line `d9cb260` dropped is back, and the grouping mechanics gave way to the per-thread guarantee.

packages/stream_chat/lib/src/core/util/message_merging.dart
VelikovPetar: Addressed in 8e68896 — same identity-return line restored on the removal side.

packages/stream_chat/lib/src/core/util/message_predicates.dart
VelikovPetar: Addressed in 8e68896 — the doc now separates these from `MessageVisibility`: they encode channel-state rules rather than properties of the message itself.

packages/stream_chat/test/src/core/util/message_merging_test.dart
VelikovPetar: Addressed in 8e68896 — the fixture defaults to `MessageType.regular` now; it moved to `message_merging_fixtures.dart`, shared by the four files the split produced.

packages/stream_chat/test/src/core/util/message_merging_test.dart
VelikovPetar: Addressed in 8e68896 — went with the split: `message_merging_messages_test.dart` (18 tests), `_threads_test.dart` (10), `_pinned_test.dart` (4), `_locations_test.dart` (5). No `group` left in any of them, and the shared helpers moved to `message_merging_fixtures.dart`.

packages/stream_chat/test/src/core/util/message_merging_test.dart
VelikovPetar: Addressed in 8e68896 — added your non-empty case pinning the fresh-map return, plus the three gaps you listed: duplicate ids in `existing`, the single-message path when `createdAt` re-sorts, and a custom `update` on `mergePinnedMessages`. That last one uses an idempotent callback on purpose — a non-idempotent one fails today, since `mergeMessages` applies `update` twice on the existing-entry fast path (it hands the already-resolved message *and* `update` to `sortedUpsertAt`). Unchanged from master, so I left it for a follow-up.

packages/stream_chat/test/src/core/util/message_predicates_test.dart
VelikovPetar: Addressed in 8e68896 — the local fixture defaults to `MessageType.regular` now.

packages/stream_chat/test/src/core/util/message_predicates_test.dart
VelikovPetar: Addressed in 8e68896 — both groups are gone and the eight tests carry the subject in their names, as you laid out.

packages/stream_chat/CHANGELOG.md
```diff
@@ -5,6 +5,10 @@
 - Fixed `CurrentPlatform` throwing `UnimplementedError` on WebAssembly builds.
 - Fixed live location expiry emitting repeated `location.expired` events for the same expired location.
 
+🔄 Internal / Non-breaking
+
+- Errors thrown while handling a channel event are now logged as warnings instead of reaching the root zone, where crash reporters report them as fatal.
```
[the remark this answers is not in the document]
VelikovPetar: Good catch, this one holds. `deleteDraft` is `async void` and `draft.deleted` reaches it through two sync hops, so the error goes straight to the zone - `_guard` never sees it. That's a fifth escape route the doc didn't cover.

Fixed in a862506: the changelog now says errors thrown *synchronously*, and the doc states the rule rather than listing the four async handlers. Behaviour is unchanged from master, so this was a docs defect, not a regression.

packages/stream_chat/lib/src/client/channel/channel_event_handler.dart
```diff
@@ -0,0 +1,501 @@
...
+
+  // Runs the given [stage] of [event] handling, containing a synchronous
+  // failure so the remaining stages still run for the same event.
+  void _guard(Event event, String stage, void Function() region) {
+    try {
+      region();
+    } catch (error, stackTrace) {
+      _client.logger.warning('Error handling ${event.type} event in $stage', error, stackTrace);
+    }
+  }
```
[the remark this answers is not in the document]
VelikovPetar: There's no `FutureOr` in this file - `_guard` takes `void Function()` and always has. The change you're describing was never made, so there's nothing to be ineffective here.

This file also belongs to #2943, which is already merged to master; this PR's diff is the `message_merging` / `message_predicates` files. The real point - failures after an `await` escape the guard - is the same one you raised and withdrew on #2943, and it's documented on `handleEvent` now.

packages/stream_chat/lib/src/core/util/message_merging.dart
```diff
@@ -0,0 +1,301 @@
...
+      if (oldIndex == -1 && !upsert) return existingList;
+
+      final resolved = oldIndex == -1 ? message : update(existingList[oldIndex], message);
+
+      final mergedMessages = existingList.sortedUpsertAt(
+        oldIndex,
+        resolved,
+        update: update,
+        compare: sortByCreatedAt,
+      );
```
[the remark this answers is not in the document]
VelikovPetar: Confirmed, the callback does run twice on that path. Not fixing it here though: the block is identical to master, `MessageMerging` is unexported, and every `update` parameter is on a private method - the public surface is `updateMessage` (default) and `replaceMessage` (hardcodes `replaceUpdate`), both idempotent. So no caller can pass a non-idempotent callback today.

Tracking it as a follow-up.

xsahil03x (approved):

PR #2945, opened 2026-09-07
refactor(llc): rework the reconnect catch-up into SyncManager

xsahil03x:
Implements the [Sync events limit](https://app.notion.com/p/stream-wiki/Sync-events-limit-3666a5d7f9f6801f97ecc1461fc62049) spec, matching `stream-chat-swift` (`SyncRepository`) and `stream-chat-android` (`SyncManager`), and takes the follow-ups that implementing it turned up.

Ticket: [FLU-756](https://linear.app/stream/issue/FLU-756/sync-events-limit-skip-replaying-oversized-sync-payloads) · supersedes #2934, whose commit is included here.

### The spec

Replaying a large `/sync` payload through `handleEvent` holds local persistence and state updates long enough to slow down the requests that need them. Payloads over 250 events are no longer replayed; the channels they covered are re-queried with `queryChannels` instead, and `lastSyncAt` advances only once that refresh succeeded — dropping the events is safe when their state has been re-fetched, and advancing past a failed refresh would lose them.

### The follow-ups

- **`/sync` was sent every known channel id**, which the endpoint rejects past 255. Accounts over that many never caught up on missed events at all — every reconnect was refused, and the refusal was read as stale local state and wiped the entire offline database. Now capped at 100, deduplicated first (the server counts duplicates before collapsing them).
- **The recovery refresh silently covered only its first page.** The backend caps a channels response at 30 (`DefaultChannelsPagerMaxLimit`), so a single `queryChannels` for 300 cids refreshed 30 of them. It now pages by 30.
- **A refused window is the server saying local state is too far behind to reconcile**, whether it is older than 30 days or holds more events than the endpoint will return. Both are handled the same way — drop the store, repopulate from the refresh, advance. The client no longer predicts either limit.
- **Replaying an event runs the whole event pipeline**, so a listener can throw part way through a window. The replay now stops there and keeps `lastSyncAt`, since advancing past a partly-applied window would lose the remainder.
- **The flow moved out of `client.dart` into a library-private `SyncManager`**, taking the client and a fetch callback and reading the clock through `package:clock`. That is what makes the paths above testable at all.

The caps are read from the backend rather than guessed: `channel_cids` is validated `min=1,max=255`, and the 2000-event limit is shared across the whole request rather than applied per channel (`lib/chat/controller/v1/sync.go`).

### Verification

`dart analyze --fatal-infos` and `dart format` clean. The `stream_chat` suite passes (1735 tests), 34 of them new in `sync_manager_test.dart` covering the cid cap, the replay limit on both sides of the boundary, both give-up paths, partial-refresh accounting, and the never-throws contract on `sync` and `recoverState`.

## Summary by CodeRabbit

* **New Features**
  * Added public access to the reconnect state recovery setting.
  * Improved reconnect recovery to replay missed events and refresh affected channels as needed.

* **Bug Fixes**
  * Large offline event backlogs now trigger a cache reset and channel refresh instead of excessive replay.
  * Reconnecting with more than 255 channels no longer clears the offline cache or skips missed events.
  * Recovery now refreshes all applicable active channels instead of only the first 30.


packages/stream_chat/lib/src/client/sync_manager.dart
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    for (final event in events) {
+      logger?.fine('Syncing event: ${event.type}');
+      client.handleEvent(event);
+    }
+
+    await _advanceLastSyncAt(nextSyncAt);
+    return const <String>{};
+  }
```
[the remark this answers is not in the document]
xsahil03x: Valid finding, and the diagnosis is right — a checkpoint advanced past channels that were never asked about does lose their events, and a direct `sync()` has no refresh to fall back on.

Not paging the request, though. Paging scales the wrong thing: it would triple the reconnect cost for large accounts to keep syncing channels nobody is looking at. The spec this implements ([Sync events limit](https://app.notion.com/p/stream-wiki/Sync-events-limit-3666a5d7f9f6801f97ecc1461fc62049)) scopes the fallback to *actively watched* channels, and the reason we cap arbitrarily today is that the LLC has no such registry — `state.channels.keys` holds every channel any query has paged through, not what is on screen. Swift uses `activeChannelControllers`, Android `stateRegistry.getTrackedWatchedChannels()`.

Tracked as the highest-priority follow-up on [FLU-756](https://linear.app/stream/issue/FLU-756/sync-events-limit-skip-replaying-oversized-sync-payloads). With that registry the set is the handful of channels being viewed, so it sits far below the cap and the truncation stops being reachable in practice. Fixing the scoping is the change that makes this correct; paging would entrench the thing that makes it wrong.

packages/stream_chat/lib/src/client/client.dart
```diff
@@ -599,46 +606,13 @@ class StreamChatClient {
...
-
-      handleEvent(
-        Event(
-          type: EventType.connectionRecovered,
-          online: true,
-        ),
-      );
-    }
+    await _syncManager.recoverState();
+    handleEvent(Event(type: EventType.connectionRecovered, online: true));
```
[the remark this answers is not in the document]
xsahil03x: Good catch — fixed in 79db2922f, though guarding at the call site would have masked the contract rather than restored it.

`recoverState` documents that it never throws, and `sync` does too, so the guard belongs in them. Checking the escape routes turned up a second one you did not flag: `client.handleEvent` in the replay loop runs the whole event pipeline, so a listener can throw part way through a window and escape `sync()` as well.

Both are now closed. `recoverState` guards its body, covering the `persistenceEnabled` route you identified. The replay loop stops on a throwing listener and **leaves `lastSyncAt` where it was** — advancing past a window that was only partly applied would lose the un-applied remainder, so the next catch-up asks for it again.

Two tests added: `sync` completes when applying an event throws, and `lastSyncAt` is unmoved after a partial replay.

packages/stream_chat/lib/src/client/sync_manager.dart
```diff
@@ -0,0 +1,267 @@
...
+    // refresh below still runs — it needs the network, not the local store.
+    // Guarded so the contract above holds: `persistenceEnabled` reads a
+    // user-supplied persistence client, and replaying an event runs the whole
+    // event pipeline. A throw here would stop the caller announcing recovery.
+    try {
+      var refreshed = const <String>{};
+      if (client.persistenceEnabled) refreshed = await sync(cids: cids);
+
+      if (client.recoverStateOnReconnect) {
+        final stale = cids.whereNot(refreshed.contains).toList();
```
VelikovPetar: Bit curious about this part: Are we sure that the `/sync` will always return all events required to sync the state for the channel? If for example we do not receive all events for a given `cid` (not sure if poll-related events are replayed) - if we omit the `queryChannels`, we might end up with a channel with outdated state. 
Additionally, if we skip re-querying the synced channels, wouldn't they remain unwatched after the reconnect?

packages/stream_chat/lib/src/client/sync_manager.dart
```diff
@@ -0,0 +1,267 @@
...
+    });
+  }
+
+  /// Recovers the state of the channels that were active before the connection
+  /// was lost, re-querying the ones the replay did not already refresh.
+  ///
+  /// Best-effort and never throws: the connection can drop again while this is
+  /// in flight, and what did not recover is left for the next reconnect.
+  Future<void> recoverState() async {
+    final cids = client.state.channels.keys.toList();
```
VelikovPetar: Would it make sense to sort the keys here by recency (matching the path when they are retrieved from the cache)? Otherwise we might end up syncing (and dropping) arbitrary channels?

VelikovPetar (approved):

renefloor (commented):
The extraction itself is a clear improvement — the invariants are stronger than the inlined version's, the failure handling is deliberate, and the test suite is genuinely behavioural rather than mock-coupled. `dart analyze --fatal-infos` and `dart format` are clean, the 247 tests pass, `clock` matches the constraint already in `melos.yaml`, and there's no lock re-entrancy in `recoverState → sync → _refreshPages`. `_sortActiveCidsByRecency` reading recency off `channelState` instead of `Channel`'s throwing date getters is exactly right, and the recency order genuinely matches `ChannelDao.cids`.

**Three things I'd resolve before merging.** All are scoped to apps that wire up `stream_chat_persistence` (the replay path is gated on `persistenceEnabled`) — which is the audience this PR is for. Detail is inline on the relevant lines; this is just the map:

1. **An oversized window flushes the whole store but repopulates at most 100 channels** — `sync_manager.dart:273`. New trigger, and not rare.
2. **The 100-cid cap drops events permanently** — `sync_manager.dart:218`. Regression for the 101–255 band, which used to be fully covered.
3. **After a flush, the checkpoint write is silently discarded** — `sync_manager.dart:72`. Pre-existing persistence bug, but the new rollback logic is built entirely on that write, and two green tests assert behaviour production can't produce.

Two pieces of SDK context that turn the first two from theoretical into user-visible, since both inline notes lean on them:

- `stream_chat_flutter_core` sets `client.recoverStateOnReconnect = false` for **every** Flutter SDK app (`stream_chat_core.dart:235`). The mitigation `_maxSyncCids` relies on — "channels past the cap are left to `recoverState`" — therefore never runs in a Flutter app. It only helps Dart-only consumers.
- The default channel-list refresh on `connectionRecovered` covers only its first page: `defaultChannelPagedLimit = 10` × `defaultInitialPagedLimitMultiplier = 3` = 30 channels. And nothing re-queries an already-materialised channel — `StreamChannel._maybeInitChannel` only watches `if (channel.state == null)` (`stream_channel.dart:847`), and `message_list_core` has no reconnect handling. So stale channel state is visible and persists until the list happens to re-query that page.

packages/stream_chat/lib/src/client/sync_manager.dart
```diff
@@ -0,0 +1,310 @@
...
+      logger?.warning(
+        'Putting lastSyncAt back: store dropped: $flushed, '
+        'refreshed ${refreshed.length} of ${cids.length} channels',
+      );
+      await _recordLastSyncAt(from);
+      return refreshed;
+    }
+
+    await _recordLastSyncAt(to);
+    return refreshed;
```
renefloor, line 289: **Must fix — the destruction scope doesn't match the repopulation scope.**

`_flushStore()` drops *every* table, but `_refreshPages(cids)` only ever restores the <=100 `cappedCids` at `messageLimit ?? 25` (`client.dart:857`). Everything else in the store — messages, members, reads, drafts, reminders, pinned messages, and failed sends that only lived on disk — is deleted with nothing repopulating it.

This is a new trigger: before this PR a full flush only happened on a server 400, and a 300-event window just replayed. It also isn't rare — `/sync` returns up to 2000 events across the capped channels, so a busy account crosses 250 after a fairly short spell offline, making a whole-DB wipe a routine reconnect outcome.

Two bits of SDK context make it user-visible rather than self-healing: `stream_chat_flutter_core` sets `recoverStateOnReconnect = false` for every Flutter app, so the `stale` branch in `recoverState` never runs; and the default channel-list refresh on `connectionRecovered` covers only its first page (`defaultChannelPagedLimit = 10` x `defaultInitialPagedLimitMultiplier = 3` = 30 channels), while nothing re-queries an already-materialised channel (`StreamChannel._maybeInitChannel` only watches `if (channel.state == null)`).

`deleteChannels(cids)` / `deleteMessageByCids(cids)` are already on `ChatPersistenceClient` and would scope the drop to what the refresh actually replaces.

packages/stream_chat/lib/src/client/sync_manager.dart
```diff
@@ -0,0 +1,310 @@
...
+    return (refreshed, failure);
+  }
+
+  // Asks for the window [lastSyncAt] opens on [cids] and applies what comes
+  // back. [sync] has already decided there is a window worth asking for, and
+  // holds the lock while this runs.
+  Future<Set<String>> _performSync(List<String> cids, DateTime lastSyncAt) async {
+    // Deduplicated before capping: the endpoint counts duplicates against its
+    // own limit, so leaving them in would spend slots on nothing.
+    final cappedCids = cids.toSet().take(_maxSyncCids).toList();
```
renefloor, line 220: **Must fix — the cap drops events permanently, and regresses the 101-255 band.**

This caps to 100 cids, and `:262` then advances the checkpoint to the newest event of that capped window. `lastSyncAt` is a single global value (`ChatPersistenceClient` has no per-channel checkpoint), so once it moves, the events belonging to cids 101+ are never requested again.

Worth flagging that this band used to work: pre-PR all cids went to `/sync` uncapped and the endpoint accepts up to 255, so an account with 150 channels was fully covered. Now 50 of them silently lose their window. `ChannelDao.cids` returns up to 250, so the direct `client.sync()` path reaches this readily.

Nothing compensates in a Flutter app — `recoverStateOnReconnect` is always `false` there, the list refresh covers 30, and opening a channel doesn't re-query it. The symptom is stale previews and unread counts outside the first 30 list entries, healing only when the list happens to re-query that page.

Chunking `cids.slices(_maxSyncCids)` sequentially and advancing the checkpoint only after the last chunk lands keeps the request-size fix without the loss — and plays well with the shared 2000-event ceiling this comment reasons about.

packages/stream_chat/lib/src/client/sync_manager.dart
```diff
@@ -0,0 +1,310 @@
...
+  // Swallows a write failure: there is nowhere else to keep the checkpoint, and
+  // a catch-up that applied its events but could not write it down is not a
+  // failed catch-up.
+  Future<void> _recordLastSyncAt(DateTime to) async {
+    try {
+      await _store?.updateLastSyncAt(to);
+    } catch (error, stk) {
+      logger?.warning('Failed to record lastSyncAt as $to', error, stk);
+    }
+  }
```
renefloor, line 78: **Must fix — after a flush, this write is silently discarded.**

`ConnectionEventDao.updateLastSyncAt` is `UPDATE connectionEvents ... WHERE id = 1` (`connection_event_dao.dart:40`), and `flush()` deletes every table including `ConnectionEvents` (`drift_chat_database.dart:83`). A post-flush write therefore matches **zero rows and does not throw** — the catch never fires and nothing is logged. Only `updateConnectionEvent` recreates the row, seeding `lastSyncAt: connectionInfo?.lastSyncAt`, i.e. `null`.

This is a pre-existing persistence bug (the old 400 path had it too), but the new design leans on this write, so both discard paths are affected:

- `_discardOversizedWindow`'s rollback at `:284` can only *matter* when the flush succeeded — precisely when the write cannot land. `lastSyncAt` becomes `null`, the next `sync()` takes the fresh-start branch at `:124`, and the window is never re-requested. The comment at `:270-272` isn't achievable with the shipped client.
- `_discardRefusedWindow`'s "advances either way" at `:295-298` is likewise a no-op; the checkpoint goes to `null`, not forward.

It's also timing-dependent: health checks call `updateConnectionInfo` every ~20s, re-inserting row `id: 1`. Same code, two outcomes depending on whether one falls between the flush and the write.

Making `updateLastSyncAt` an upsert (mirroring `updateConnectionEvent`) is probably the smallest fix. Separately, this method can't distinguish a lost write from a successful one — the DAO returns rows-affected but the interface discards it.

renefloor (commented):
Moved the three blocking notes onto the relevant lines so they are easier to act on. The summary comment has been trimmed to match.

packages/stream_chat/lib/src/client/sync_manager.dart
xsahil03x: Fixed in 2461cc90c — `updateLastSyncAt` upserts now, seeding the row when a reset took it and carrying the other columns over when it didn't.

Confirmed the diagnosis before changing anything: against the old DAO the write reports success and affects **0 rows**, and `lastSyncAt` reads back `null`. The new tests fail on the old implementation and pass on the new one.

One wrinkle worth flagging: `type` is non-nullable, so a checkpoint-only row has to put something there. It stores `EventType.any`, and `connectionEvent` filters that row back out. Without the filter `getConnectionInfo()` would start returning a row after a reset, and `connectUser`'s offline fallback (`client.dart:461`) would return the user where it used to rethrow. The alternative — making `type` nullable — is a schema migration this PR has no business carrying.

packages/stream_chat/lib/src/client/sync_manager.dart
xsahil03x: Fixed in 7cb885340 — the cap is 255 now.

The regression is real, and worse than I'd have guessed: `ChannelDao.cids` has been capped at 250 since #2880, which is already under the endpoint's 255, so the persisted path could never be refused in the first place. The 101–250 band lost coverage in exchange for nothing. At 255 it isn't truncated at all, and it still costs one request.

The event-budget reasoning I picked 100 for doesn't survive a second look either: the replay is skipped at 250 events, long before the server's 2000, and the cids are recency-ordered, so the channels the cap dropped were the quietest ones. What a cap is actually for is the size refusal — the one refusal that doesn't clear itself, since the next catch-up asks about the same channels. That's what the comment says now.

Not chunking: `recoverState`'s in-memory list is the only unbounded source left, and scoping it to actually-watched channels ([FLU-756](https://linear.app/stream/issue/FLU-756/sync-events-limit-skip-replaying-oversized-sync-payloads)) makes the cap unreachable rather than entrenching it.

One correction on the symptom: `PagedValueNotifier.refresh()` resets to the initial value and re-runs the initial load, so the list keeps page 1 and refetches later pages from the server as you scroll — previews come back fresh. What does persist is the offline store missing those events, and a materialised channel outside the cap opening stale.

renefloor (approved):

PR #2951, opened 2026-09-10
chore(samples): xcode 27, no cocoapods, SPM re-enabled

renefloor:
### 🎯 Goal

Get every sample in the repo onto Xcode 27 — `sample_app` and the five per-package example apps. While checking, CocoaPods turned out to be removable entirely, and the SPM signing failure from #2672 turned out to be fixable rather than permanent.

### 🛠 Implementation details

- **Solves #2672 / #2675.** `FLUTTER_XCODE_CODE_SIGN_IDENTITY` / `..._PROVISIONING_PROFILE_SPECIFIER` reach xcodebuild as *command-line* settings, so they land on every SPM package bundle too → `Signing for "X" requires a development team` (40+ targets). CocoaPods was immune only because `podhelper.rb` sets `CODE_SIGNING_ALLOWED = NO` on pod resource bundles. `build_ipa` now writes both settings into `ios/Flutter/Signing.xcconfig`, which only the Runner target reads. SPM stays on; the `Disable Swift Package Manager` steps are gone.
- **CocoaPods removed from iOS and macOS.** `media_kit_video` was the last pod-only plugin and was dead weight — its desktop branch is gated on Windows/Linux, which `sample_app` has no runner directories for, so it always fell through to `DefaultStreamVideoPlayer`. Dropped it, deintegrated pods, deleted both Podfiles and the `cocoapods` gem.
- Crashlytics `upload-symbols` now globbed from the SPM checkout instead of `ios/Pods/`.
- macOS deployment target 11.5 → 12.0 — Flutter 3.47's `FlutterMacOS.podspec` floor, below which `pod install` refused to resolve.
- **The example apps too.** They pinned iOS 11/12 and macOS 10.11–11.0, all below Flutter 3.47's floors, so none of them built at all; four Podfiles also declared a `RunnerTests` target no project defines. Targets raised, stanza dropped, CocoaPods deintegrated there as well — every plugin they use is already a Swift Package. Flutter's pending project migrations (`@main`, `FlutterImplicitEngineDelegate`, `UIApplicationSceneManifest`) came along with the builds.
- CI: iOS jobs on the `xcode-27` image behind a new shared `.github/actions/setup-xcode` (one place for the version); `select_xcode` dropped from the Fastfile since it can't express "latest" and was overriding the workflows with a pinned 26.2 the image lacks; e2e jobs pin 26.2 explicitly; pod cache action deleted.

### ☑️ Verification

`distribute_internal` is green on this branch — [run 34455350823](https://github.com/GetStream/stream-chat-flutter/actions/runs/34455350823). Both iOS jobs signed, symbolicated and shipped (Firebase App Distribution + `Successfully uploaded package to App Store Connect`) with zero signing errors and zero `pod install`. Locally `sample_app` also builds *and runs* on an iOS 27 simulator and macOS 26.6.2, and all seven example-app configurations build from a wiped `build/` with no Podfile regenerated and no `pod install`. Note CI never builds the examples — no workflow references them and `melos run analyze` skips them — so those seven builds are their only coverage.

#### General
- [ ] Assigned a person / code owner group (required)
- [ ] Thread with the PR link started in a respective Slack channel (#flutter-team) (required)
- [ ] PR is linked to the GitHub issue it resolves

### ☑️Reviewer Checklist
- [ ] Sample runs & works
- [ ] UI Changes correct (before & after images)
- [ ] Bugs validated (bugfixes)
- [ ] New feature tested and works
- [ ] All code we touched has new or updated Documentation

## Summary by CodeRabbit

- **Build & Platform Updates**
  - Example iOS projects now require iOS 15 or later; macOS examples require macOS 12 or later.
  - Example applications now use Swift Package Manager for Flutter plugin integration.
  - Updated Flutter app lifecycle and scene configuration for current Apple platform tooling.

- **Video Playback**
  - The sample app now uses the SDK’s default video player instead of custom desktop playback support.

- **Release Engineering**
  - Updated Xcode setup and dependency caching for more consistent iOS and macOS builds.


.github/actions/cache-swiftpm/action.yml
```diff
@@ -0,0 +1,24 @@
...
+    default: 'ios'
+runs:
+  using: "composite"
+  steps:
+    - uses: actions/cache@v6
+      with:
+        # Only the downloaded binary artifacts. `repositories` and `checkouts`
+        # add ~590MB compressed on top of this and SwiftPM rebuilds them from
+        # the clones, so they are not worth the repository's cache budget.
+        path: sample_app/build/${{ inputs.platform }}/SourcePackages/artifacts
```
xsahil03x: Do you think hardcoding to `sample_app` is good here? I think we should ask for the app directory then append the path `build/${{ inputs.platform }}/SourcePackages/artifacts`

.gitignore
```diff
@@ -84,6 +84,9 @@ GeneratedPluginRegistrant.*
 **/ios/Flutter/ephemeral/**
 Podfile.lock
 **/xcshareddata/swiftpm/Package.resolved
+# ...except the sample app's, so SwiftPM resolution is pinned and reproducible
+# and CI can key its package cache on it.
+!sample_app/*/Runner.xcworkspace/xcshareddata/swiftpm/Package.resolved
```
xsahil03x: Lets move this under `# Exceptions to iOS rules`

.github/workflows/e2e_test.yml
```diff
@@ -138,6 +138,13 @@ jobs:
     steps:
       - uses: actions/checkout@v7
 
+      - uses: ./.github/actions/setup-xcode
+        with:
+          # Matches the simulator runtimes this job pins: macos-15 ships an iOS
+          # 26.2 runtime but no 26.3, so 'latest' (Xcode 26.3 on that image)
+          # would have no matching runtime.
+          xcode-version: '26.2'
```
xsahil03x: This isn't right: macos-15 carries both Xcode 26.3 and the iOS 26.2 runtime, and runtimes are shared across Xcode installs, so 26.3 would run the 26.2 sim fine.

The real reason the step has to exist is a better one — the image *default* is Xcode 16.4, so without it the job silently drops from 26.2 to 16.4, which is exactly what the Fastfile's `select_xcode` used to prevent. (The dispatched run shows it doing exactly that work: `Switching Xcode to version '26.2'`.) Worth saying that instead, since it also justifies the step:

```suggestion
          # macos-15 defaults to Xcode 16.4, so this has to be pinned or the job
          # silently builds on it. 26.2 keeps what the Fastfile's select_xcode
          # used to select.
          xcode-version: '26.2'
```

.github/workflows/e2e_test_cron.yml
```diff
@@ -106,11 +106,16 @@ jobs:
...
-      XCODE_VERSION: "26.2"
     steps:
       - uses: actions/checkout@v7
 
+      - uses: ./.github/actions/setup-xcode
+        with:
+          # Matches the simulator runtimes this job pins: macos-15 ships an iOS
+          # 26.2 runtime but no 26.3, so 'latest' (Xcode 26.3 on that image)
+          # would have no matching runtime.
+          xcode-version: '26.2'
```
xsahil03x: Same as the one on `e2e_test.yml` — macos-15 has Xcode 26.3 *and* the iOS 26.2 runtime, and runtimes are shared across installs. The pin is needed because the image default is Xcode 16.4, not because 26.3 lacks a runtime:

```suggestion
          # macos-15 defaults to Xcode 16.4, so this has to be pinned or the job
          # silently builds on it. 26.2 keeps what the Fastfile's select_xcode
          # used to select.
          xcode-version: '26.2'
```

xsahil03x (commented):
Only looked at the workflows and the Fastfile — the pbxproj/Podfile churn I'm taking on trust from the seven local builds.

The signing fix is the right shape, and I convinced myself it actually holds: the Runner target's Release config declares neither `CODE_SIGN_IDENTITY` nor `PROVISIONING_PROFILE_SPECIFIER` (and this PR doesn't touch that), so the xcconfig values flow through; Profile shares `Release.xcconfig` as its base but overrides both at target level, so a profile build can't accidentally pick up the match profile. Generated.xcconfig has no signing keys either, so the include order is fine. Nice that `select_xcode` going away also takes the allure-lane exclusion list with it.

Thanks for dispatching the iOS e2e — that was the one path the green `distribute_internal` run didn't cover. Batch 0 is clean: no `pod install`, no Podfile regenerated, `Xcode is set to 26.2.0 (17C52)`, 13 + 27 tests passed.

Dropping the pod cache also costs nothing — `build (ios)` came in at 6m45s here against 10–12m on recent runs.

Two comment fixes and one suggestion inline.

Unrelated to the code: `build (ios)` runs on every PR, so this puts larger-runner minutes on a public repo on the default path — you'd know better than me whether that's within budget.

sample_app/ios/fastlane/Fastfile
```diff
@@ -156,10 +170,18 @@ platform :ios do
...
+      "#{root_path}/build/ios/SourcePackages/checkouts/*/Crashlytics/upload-symbols"
+    ).first
+    UI.user_error!('Could not find Crashlytics upload-symbols in build/ios/SourcePackages') if upload_symbols.nil?
+
     upload_symbols_to_crashlytics(
       dsym_path: dsym_zip,
       gsp_path: "#{root_path}/ios/Runner/GoogleService-Info.plist",
-      binary_path: "#{root_path}/ios/Pods/FirebaseCrashlytics/upload-symbols"
+      binary_path: upload_symbols
     )
```
xsahil03x: `build/ios/SourcePackages` is the Flutter tool's `-clonedSourcePackagesDirPath`, and the `*` absorbs whatever firebase-ios-sdk names its checkout — both can drift under a Flutter or Firebase bump. Because this lane runs *before* `firebase_app_distribution` / `upload_to_testflight`, a drift there hard-fails an already-built release over a symbol upload. dSYMs can be re-uploaded after the fact; a failed release costs a full rerun.

```suggestion
    if upload_symbols.nil?
      UI.important('Could not find Crashlytics upload-symbols under build/ios/SourcePackages; skipping dSYM upload')
      next
    end

    upload_symbols_to_crashlytics(
      dsym_path: dsym_zip,
      gsp_path: "#{root_path}/ios/Runner/GoogleService-Info.plist",
      binary_path: upload_symbols
    )
```

.github/workflows/e2e_test.yml
renefloor: I think it's more about predictability. We could use latest of 26.3, but than the xcode-version would not match the iOS runtime. It's fine, but matching them is probably better. You change reads like it just gives a reason based on history "what the Fastfile's select_xcode used to select."

.github/workflows/e2e_test_cron.yml
renefloor: See also response on other comment, I know that 26.3 has the 26.2 runtime, but we want to match them.

sample_app/ios/fastlane/Fastfile
renefloor: Agreed, and thanks — this was a real bug, not a style call. Confirmed the ordering you describe in all three release lanes: `build_ipa` → `upload_dsyms_to_crashlytics` → `firebase_app_distribution` / `upload_to_testflight`, so a glob miss aborted an already-built, already-signed release. The asymmetry you point at is the deciding argument: dSYMs can be uploaded after the fact, a failed release costs a full rebuild.

Applied in fecd043ba, with one deviation: `UI.error` instead of `UI.important`. Both are non-fatal, but the failure mode of skipping is silent — Crashlytics quietly goes unsymbolicated and nobody notices for months. Red in the log makes that harder to miss. Happy to switch to `UI.important` if you'd rather keep it a warning.

Also took your framing for the comment, since "outside our control" is the durable reason the branch has to exist at all:

```ruby
# Both the SourcePackages directory and the checkout name are outside our
# control, so a miss is possible. This lane runs before the app is
# distributed: symbols can be uploaded later, a failed release cannot.
```

.github/workflows/e2e_test.yml
renefloor: You're right that runtimes are shared across Xcode installs — that part of my comment was wrong. It's already gone: 6ee54122a (pushed after this review) reduced the line to what's on the branch now, at `.github/workflows/e2e_test.yml:143`:

```yaml
          # Pinned to match the simulator runtimes this job selects.
          xcode-version: '26.2'
```

No false claim left, and it states the intent I described — matching the toolchain to the runtime for predictability, rather than justifying the value by what the deleted `select_xcode` used to pick. Resolving on that basis.

.github/workflows/e2e_test_cron.yml
renefloor: You're right that runtimes are shared across Xcode installs — that part of my comment was wrong. It's already gone: 6ee54122a (pushed after this review) reduced the line to what's on the branch now, at `.github/workflows/e2e_test_cron.yml:114`:

```yaml
          # Pinned to match the simulator runtimes this job selects.
          xcode-version: '26.2'
```

No false claim left, and it states the intent I described — matching the toolchain to the runtime for predictability, rather than justifying the value by what the deleted `select_xcode` used to pick. Resolving on that basis.

xsahil03x (approved):

PR #2952, opened 2026-09-10
ci(repo): route every Flutter setup through setup-flutter

renefloor:
### 🎯 Goal

Reclaim ~1.6GB of the repository's 10GB Actions cache allowance, which currently sits at ~9.05GB and therefore evicts caches other jobs need.

### 🛠 Implementation details

- Five call sites configured `subosito/flutter-action` independently and their cache keys had drifted into two shapes — `flutter-:os:-:channel:-:version:-:arch:` in `setup-flutter`, and `…-:hash:-${{ hashFiles('**/pubspec.lock') }}` everywhere else — so the same SDK is stored twice. Both of these are live right now: `flutter-linux-stable-3.47.3-x64` (1640 MiB, master) and `flutter-linux-stable-3.47.3-x64-e8113bf…-b09b477…` (1640 MiB, PR ref).
- `setup-flutter` now takes optional `flutter-version` and `channel` inputs and owns the key shape; `update_goldens` (×2), `release_publish`, `legacy_version_analyze`, `beta_version_analyze` and `actions/pana` pass only what differs. Composite-in-composite is already used by `actions/allure-launch`, so `actions/pana` is not new ground.
- `legacy_version_analyze` still pins Flutter 3.44.0 and so still gets its own SDK entry — correct, not duplication.
- Two deliberate behaviour changes: `release_publish` used a bare `flutter-action` with **no** caching and no version pin, and now matches everything else; `beta_version_analyze` passes an empty `flutter-version` so it keeps tracking the newest beta whatever its major, instead of inheriting the `3.x` default.

### ☑️ Verification

Not yet run — this is a draft. All six touched files parse as YAML, and no direct `subosito/flutter-action` usage remains outside `setup-flutter`. The keys can only be confirmed by landing this and watching the cache list settle to one SDK entry per os/channel/version/arch; `update_goldens`, `release_publish` and `beta_version_analyze` are dispatch/cron/schedule-triggered, so they will not exercise themselves on this PR.

### ⚠️ Also worth a look

Unrelated to this change, `gradle-Linux-…` is 2.92GB — the single largest entry, and a bigger prize than this PR if the cache limit keeps biting.

#### General
- [ ] Assigned a person / code owner group (required)
- [ ] Thread with the PR link started in a respective Slack channel (#flutter-team) (required)
- [ ] PR is linked to the GitHub issue it resolves

### ☑️Reviewer Checklist
- [ ] Sample runs & works
- [ ] UI Changes correct (before & after images)
- [ ] Bugs validated (bugfixes)
- [ ] New feature tested and works
- [ ] All code we touched has new or updated Documentation

## Summary by CodeRabbit

- **Chores**
  - Standardized Flutter setup across analysis, release, and golden-update workflows.
  - Added configurable Flutter version, channel, and SDK caching options for automated workflows.
  - Consolidated Flutter and pub-cache handling to provide more consistent build and validation environments.
  - Beta-version analysis can now disable SDK caching while retaining the appropriate Flutter channel configuration.

### DIFF
--- a/.github/workflows/beta_version_analyze.yml
+++ b/.github/workflows/beta_version_analyze.yml
@@ -21,11 +21,11 @@ jobs:
           fetch-depth: 0
 
       - name: "Install Flutter"
-        uses: subosito/flutter-action@v2
+        uses: ./.github/actions/setup-flutter
         with:
           channel: beta
-          cache: true
-          cache-key: flutter-:os:-:channel:-:version:-:arch:-:hash:-${{ hashFiles('**/pubspec.lock') }}
+          # Empty so the newest beta is used, whatever major it is.
+          flutter-version: ''
 
       - name: 📊 Analyze and test packages
         uses: ./.github/actions/package_analysis
--- a/.github/workflows/beta_version_analyze.yml
+++ b/.github/workflows/beta_version_analyze.yml
@@ -26,6 +26,9 @@ jobs:
           channel: beta
           # Empty so the newest beta is used, whatever major it is.
           flutter-version: ''
+          # Weekly, and a new beta invalidates the key, so an SDK entry would
+          # almost never be restored before it expires.
+          cache-sdk: 'false'
 
       - name: 📊 Analyze and test packages
         uses: ./.github/actions/package_analysis
--- a/.github/workflows/beta_version_analyze.yml
+++ b/.github/workflows/beta_version_analyze.yml
@@ -24,7 +24,9 @@ jobs:
         uses: ./.github/actions/setup-flutter
         with:
           channel: beta
-          # Empty so the newest beta is used, whatever major it is.
+          # Must stay empty. Inheriting the action's 3.x default would hold this
+          # job on 3.x betas, so it would stop catching the next major early —
+          # which is the only thing it exists to do.
           flutter-version: ''
           # Weekly, and a new beta invalidates the key, so an SDK entry would
           # almost never be restored before it expires.
--- a/.github/actions/pana/action.yml
+++ b/.github/actions/pana/action.yml
@@ -31,12 +31,7 @@ runs:
           yq eval '.dependencies.stream_chat = {"path": "../stream_chat"}' -i packages/stream_chat_persistence/pubspec.yaml
 
     - name: Install Flutter
-      uses: subosito/flutter-action@v2
-      with:
-        cache: true
-        channel: stable
-        flutter-version: "3.x"
-        cache-key: flutter-:os:-:channel:-:version:-:arch:-:hash:-${{ hashFiles('**/pubspec.lock') }}
+      uses: ./.github/actions/setup-flutter
 
     - name: Install Pana
       working-directory: ${{ inputs.working_directory }}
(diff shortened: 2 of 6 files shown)


.github/workflows/beta_version_analyze.yml
```diff
@@ -21,11 +21,14 @@ jobs:
...
 
       - name: "Install Flutter"
-        uses: subosito/flutter-action@v2
+        uses: ./.github/actions/setup-flutter
         with:
           channel: beta
-          cache: true
-          cache-key: flutter-:os:-:channel:-:version:-:arch:-:hash:-${{ hashFiles('**/pubspec.lock') }}
+          # Empty so the newest beta is used, whatever major it is.
+          flutter-version: ''
```
xsahil03x, line 30: i don't like passing empty here, better to pass the version or channel directly like we had earlier.

renefloor: We do pass the channel on line 26. This is just to override the default version in the shared action

renefloor: The empty value isn't incidental — it encodes a requirement, and I've pushed 9b1ac12f2 to make the comment say so rather than restate the mechanism:

```yaml
# Must stay empty. Inheriting the action's 3.x default would hold this
# job on 3.x betas, so it would stop catching the next major early —
# which is the only thing it exists to do.
flutter-version: ''
```

On passing the version directly instead: there are 19 `setup-flutter` usages across the workflows and actions, and only 2 pass anything today (this one passes `channel`, `legacy_version_analyze` passes `flutter-version`). Inverting the default so callers supply `3.x` would add a line to 17 call sites and scatter the version policy — bumping to `4.x` later would mean editing 17 places instead of one, which is the drift this PR is trying to remove.

Three workable options, and I'm happy with any of them:

1. **`flutter-version: ''`** (current) — self-maintaining, at the cost of an empty string meaning "newest on channel".
2. **Drop the line and inherit `3.x`** — cleanest call site, no empty string, and closest to what you asked for: only `channel: beta` and `cache-sdk: 'false'` remain. Today it's a no-op, since the beta channel's newest is `3.48.0-0.4.pre`, exactly what `''` resolves to. I'm fine with this: a Flutter 4 release is not something we'd miss.

   One caveat worth recording, because I assumed the opposite and checked: it degrades *silently*, not loudly. The beta release manifest keeps history — 231 beta entries, 120 of them under `3.x` back to `3.1.0` — so once 4.0 betas ship, `3.x` still matches and resolves the newest 3.x beta instead of failing. The job keeps reporting green while testing a stale beta, so nothing prompts the update; it depends on us remembering after we notice the release.
3. **Keep this job on `subosito/flutter-action@v2` directly** — newly viable, because `cache-sdk: 'false'` means it never touches the SDK cache key that motivates the shared action. It would still want `pub-cache-key`, so that string gets duplicated here — a milder duplication than the SDK key, since the pub key carries no channel or version and so barely drifts.

I lean to 1 because duplication is what produced the two divergent key shapes in the first place, but 2 is a fine trade if we'd rather not have an empty string carrying meaning. Tell me which and I'll push it.

xsahil03x (approved):

PR #2956, opened 2026-09-12
refactor(llc, ui, persistence)!: adopt stream_core's foundation

xsahil03x:
Linear: FLU-783

## CLA

### 🎯 Goal

Replace six hand-written subsystems with `stream_core`'s, deleting ~3.5k lines of chat code for ~2.4k of adoption. Each was a near-verbatim fork, and three of them had drifted into bugs core has already fixed.

### 🛠 Implementation details

**Read this one commit by commit** — each is a self-contained subsystem, and the diff makes far more sense that way than as a single blob of this size.

- `InFlightCache` and `SystemEnvironment` — verbatim forks, deleted.
- `UploadState`'s variant classes renamed to `UploadStatePreparing` / `InProgress` / `Success` / `Failed`, freeing the name `Success` for core's `Result`.
- The error layer moves to core's sealed `StreamException` family. `StreamChatException` aliases the root, so `on StreamChatException catch` handles all four kinds.
- The token layer moves to core's `UserToken` / `TokenProvider` / `TokenManager` / `AuthInterceptor`. Core's interceptor fixes three bugs ours had: no retry-loop guard, no user-switch guard, and no `FormData` re-clone before a retried upload.
- Logging moves to `StreamLogger`; `LoggingInterceptor` and `ConnectionIdInterceptor` become core's.
- Anonymous connections now identify as `!anon` rather than a client-generated random id, matching every other Stream SDK.
- The WebSocket connect and reconnect URIs are logged with the user token redacted; they carried it in full at `info`.
- `info` now carries only client and connection lifecycle. Per-operation records are `debug`, per-event and per-timer ones `verbose` — a health check every 20 seconds and a line per frame were all `info`.
- A throwable the SDK does not recognise is reported as a `StreamClientException` naming the load that failed, with the original kept as its `cause` rather than stringified into the message.

### ☑️ Verification

`melos run analyze`, `melos run format` and `melos run test:dart` pass at this commit against hosted `stream_core` 0.5.0, verified in a clean worktree so no local dependency override could mask it.

A live anonymous connect against a real app key was checked by hand, because the `!anon` change alters what the client sends on the WebSocket connect payload.

## Summary by CodeRabbit

* **Breaking Changes**
  * Migrated public errors and retry handling to `StreamChatException` and core error types.
  * Renamed upload-state variants to `UploadStatePreparing`, `UploadStateInProgress`, `UploadStateSuccess`, and `UploadStateFailed`.
  * Replaced legacy token and development-token APIs with `UserToken` and token providers.
  * Replaced `logLevel` and `logHandlerFunction` with `StreamLogConfig`.
  * Flutter error builders and paged-value errors now receive `StreamChatException`.

* **Documentation**
  * Added migration guidance, deferred work tracking, upstream requirements, and updated changelogs.


packages/stream_chat_flutter/lib/src/message_input/stream_message_composer.dart
```diff
@@ -1577,7 +1577,7 @@ class DefaultStreamMessageComposerState extends State<DefaultStreamMessageCompos
           stack: stackTrace,
           library: 'stream_chat_flutter',
           context: ErrorDescription('while sending a message'),
-          silent: error is StreamChatNetworkError,
+          silent: error is StreamChatException,
```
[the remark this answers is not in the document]
xsahil03x: Fixed in `d1919e620`. You are right, and the comment above it was describing the old behaviour: it was `StreamChatNetworkError` before the retype, and `StreamChatException` is the sealed root, so it silenced `StreamClientException` too — the kind whose contract is "the SDK itself failed".

Narrowed to `StreamApiException || StreamNetworkException`, which is what the comment meant: a refusal the server sent, or a connection that failed. Both leave the message failed for the retry queue. An SDK bug now reaches `FlutterError.reportError` again.

core-migration/06-logger.md
```diff
@@ -104,17 +80,101 @@ precisely so a product can namespace them — feeds passes `SF:Ws`, `SF:Http`, `
...
-- [ ] `melos bootstrap && melos run analyze && melos run test:dart && melos run test:flutter`.
-- [ ] `refactor(llc)!:` title, `🛑️ Breaking` CHANGELOG entry, `migrations/v11-migration.md`
-      Symbol Map rows for the `package:logging` re-export and each `logger:` parameter.
-- [ ] Decisions recorded here — **especially the tag prefix** — and status box ticked in
-      `README.md`.
+- [x] `StreamLogger.configure` is called once from `StreamChatClient`, with the priority mapped
+      from `logLevel`.
+- [x] The bridge exists and is tested: a record written through `StreamLogger` reaches a
+      consumer-supplied `Logger` with the right `Level`, `error` and `stackTrace`, and nothing
+      arrives while `stream_core` is unconfigured.
```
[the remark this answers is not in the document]
xsahil03x: Fixed in 36053dc. The first item now reads `logConfig.priority`, and the bridge item is gone along with the "bridge's callers" wording — the bridge was deleted, so the checklist could have certified something that no longer exists.

core-migration/08-query-dsl.md
```diff
@@ -4,7 +4,8 @@
 evaluated client-side as well as sent to the server.
 
 **Size:** ~600 chat LOC deleted against ~1,420 core LOC. The widest public break after phase
-[03](03-errors.md), and the one hard upstream block in the plan.
+[03](03-errors.md). One small upstream ask (), and two operators chat should deprecate
+rather than port.
```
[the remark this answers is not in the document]
xsahil03x: Fixed on the sort rung, where that paragraph was rewritten: the summary now reads "**No upstream `stream_core` work** — and three operators chat should deprecate rather than port". `$nor` is no longer an upstream ask at all; it is dropped, and a preset carrying it decodes through `Filter.raw`.

packages/stream_chat_flutter_core/test/stream_reaction_list_controller_test.dart
```diff
@@ -200,7 +200,7 @@ void main() {
       expect((controller.value as Error).error, equals(chatError));
     });
 
-    test('wraps generic exceptions in StreamChatError', () async {
+    test('wraps generic exceptions in a StreamClientException', () async {
```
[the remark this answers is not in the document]
xsahil03x: Fixed in 36053dc. The test now asserts `isA<StreamClientException>()` before checking the message, so the wrapping contract the test is named for is actually enforced.

packages/stream_chat/lib/src/client/channel_delivery_reporter.dart
```diff
@@ -30,12 +30,10 @@ class ChannelDeliveryReporter {
...
   /// The optional [logger] logs warnings and errors during operation.
   ChannelDeliveryReporter({
-    this._logger,
     required this.onMarkChannelsDelivered,
     this._throttleDuration = const Duration(seconds: 1),
-  });
-
-  final Logger? _logger;
+    String tag = 'SCh:Delivery',
+  }) : _logger = StreamLogger(tag);
```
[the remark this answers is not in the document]
xsahil03x: Fixed in 36053dc. The paragraph now documents `tag` — the parameter the constructor takes — instead of the `logger` it replaced.

packages/stream_chat/test/src/core/http/stream_http_client_test.dart
```diff
@@ -209,8 +187,12 @@ void main() {
...
       await client.get(path);
     } catch (e) {
-      expect(e, isA<StreamChatNetworkError>());
-      expect(e, StreamChatNetworkError.fromDioException(error));
+      expect(e, isA<StreamApiException>());
+      // Assert on the facts the mapper read off the response rather than on
+      // object identity: the exception also carries the DioException as its
+      // cause, which equality would compare.
+      expect((e as StreamApiException).statusCode, 500);
+      expect(e.code, StreamErrorCode.internalError);
```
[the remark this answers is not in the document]
xsahil03x: Fixed in 36053dc. All six method tests and the pre-classified `.request` test now use `await expectLater(..., throwsA(...))` with `.having` matchers on `statusCode` and `code`, so a request that returns instead of throwing fails the test. The one remaining bare `catch` is the logging test at L116, where the outcome of the request is deliberately irrelevant — it asserts on the records written, not on the result.

core-migration/08-query-dsl.md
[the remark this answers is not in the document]
xsahil03x: The original ask is satisfied on this rung — line 7 reads "One small upstream ask (`$nor`)", which is the exact wording you proposed, and it is consistent with the rest of the file at this point in the stack, where `$nor` is still listed as upstream work.

My earlier reply overreached by describing the end state. The `$nor`-is-dropped rewrite belongs to #2957, which is where the `Filter` and `Sort` decisions are actually taken; pulling it down here would put a conclusion in a tree that has not reached it yet. On #2957 the same line reads "**No upstream `stream_core` work** — and three operators chat should deprecate rather than port".

packages/stream_chat/lib/src/client/client.dart
```diff
@@ -1352,9 +1324,6 @@ class StreamChatClient {
     return res;
   }
 
-  /// Get a development token
-  Token devToken(String userId) => Token.development(userId);
```
VelikovPetar, line 1356: Do we no longer support this API?

packages/stream_chat/lib/src/client/retry_queue.dart
```diff
@@ -9,21 +9,22 @@ class RetryQueue {
   /// Instantiate a new RetryQueue object.
   RetryQueue({
     required this.channel,
-    this.logger,
-  }) : client = channel.client {
+    String tag = 'SCh:RetryQueue',
```
VelikovPetar, line 12: Should we maybe still pass the `StreamLogger` here instead of the `tag` - mostly for testing purposes, if we want to stub out the logger? (the same question goes for all the other places where we are passing `tag`)

Edit: I see that we can easily manipulate the loggers in tests, so this probably doesn't make too much sense. Still, it might be a better approach to inject the logger dependency, instead of constructing it internally.

packages/stream_chat/lib/src/core/error/stream_chat_error.dart
```diff
@@ -77,20 +77,13 @@ class StreamWebSocketError extends StreamChatError {
 }
 
 /// An error raised when a network request to Stream fails.
+///
+/// Nothing in the SDK raises this any more: a failed request now reports a
+/// [StreamApiException] or a [StreamNetworkException] from `stream_core`. An
+/// `on StreamChatNetworkError catch` clause therefore still compiles but no
+/// longer matches — catch [StreamChatException] instead.
+@Deprecated('Use StreamApiException or StreamNetworkException instead')
```
VelikovPetar: Can we delete this instead of deprecating, since these changes are anyway targeting `V11`?

packages/stream_chat/lib/src/core/http/stream_http_client.dart
```diff
@@ -45,27 +49,17 @@ class StreamHttpClient {
...
-                        return logger.info(message);
-                      case InterceptStep.response:
-                        return logger.info(message);
-                      case InterceptStep.error:
-                        return logger.severe(message);
-                    }
-                  },
-                ),
+              // provided. What it writes is gated by the configured priority,
+              // so there is no separate on/off switch here.
```
VelikovPetar: I think the new dart doc added here is a bit redundant - it references an old implementation.

packages/stream_chat/lib/stream_chat.dart
```diff
@@ -12,8 +12,37 @@ export 'package:dio/dio.dart'
...
         Options,
         ProgressCallback;
-export 'package:logging/logging.dart' show Logger, Level, LogRecord;
 export 'package:rate_limiter/rate_limiter.dart';
+// Re-exported with a `show` allowlist rather than wholesale: `stream_core`
+// also declares names this barrel defines — `AttachmentFile`, `Filter`,
+// `User` — so a blanket export would not compile.
+export 'package:stream_core/stream_core.dart'
+    show
+        Failure,
```
VelikovPetar, line 21: Question: Do we need to publicly re-export all of these classes? For example, `LoggingInterceptor`, `InterceptStep` `LogPrint` look like SDK internals, I don't think we should expose them transitively. (please correct me if I am missing something)

packages/stream_chat_persistence/CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## Upcoming Beta
+
+🛑️ Breaking
+
+- `StreamChatPersistenceClient` no longer takes `logLevel` or `logHandlerFunction`. It writes through `stream_core`'s logger, configured once via `StreamChatClient`'s `logConfig`.
```
VelikovPetar: One small thing: As the changelog would be customer-facing, referencing `stream_core` here might be a bit confusing - as is general, consumers might not even be aware of the `stream_core` package.

packages/stream_chat/lib/src/client/client.dart
xsahil03x: Correct, and deliberately. `devToken` minted a `devtoken`-signed JWT, which only an app with development tokens enabled accepts — shipping a minter in the SDK is what invites it into production. Tests build their own through `testUserToken` in `test/src/utils.dart`.

It is documented in all three places a break should be: the `🛑️ Breaking` entry in `CHANGELOG.md`, a `removed` row in `migrations/v11-migration.md`'s Symbol Map pointing at backend token generation, and the reasoning in `core-migration/04-token-and-auth.md`.

packages/stream_chat/lib/src/client/retry_queue.dart
xsahil03x: Injecting it would not buy the stubbing. Core's `StreamLogger` is a `final class`, so there is nothing to subclass or mock, and it carries only the tag — every record goes to the process-wide handler. The part that varies in a test is the handler, and that is already settable, which is what you found.

So the two seams are already in the right places: `tag` injects the thing that differs per call site, and `logConfig` injects the thing you stub. A `StreamLogger` parameter would add a knob without adding reach. `StreamLogger.detached(handler, filter)` is there if some case ever needs a genuinely isolated one. Same reasoning for the other three sites that take a tag — `websocket.dart`, `channel_delivery_reporter.dart` and `app_settings_manager.dart`.

packages/stream_chat/lib/src/core/http/stream_http_client.dart
xsahil03x: Agreed — removed in f547e855f. The first sentence restated the `??`, and the second only parsed if you remembered the `logger.level != Level.OFF` guard it replaced.

packages/stream_chat/lib/stream_chat.dart
xsahil03x: You're right that they are internals, and they're dropped in f547e855f.

One thing worth flagging, because it makes this a slightly wider break than the PR set out to make: all three were already public *before* this PR. The barrel carried `export 'src/core/http/interceptor/logging_interceptor.dart';`, and that file declared `LoggingInterceptor`, `enum InterceptStep` and `typedef LogPrint`. So the `show` list was preserving the existing surface rather than adding to it — which is why it did not read as a new exposure when it was written.

Dropping them is still the right call: the interceptor is installed by default and writes through the configured `StreamLogHandler`, so routing its output is a `logConfig` concern and there is nothing left to reach for. The CHANGELOG line now says removed rather than re-exported, and the migration guide's Symbol Map has a `removed` row.

Nothing in `stream_chat_flutter`, `stream_chat_flutter_core`, persistence or the sample app referenced them, and the one test that uses `LoggingInterceptor` already imports it from `stream_core` directly.

packages/stream_chat_persistence/CHANGELOG.md
xsahil03x: Agreed — reworded in f547e855f to "Logging is configured once, on `StreamChatClient`, with `logConfig`." The entry should not point at a package a persistence consumer never adds to their pubspec.

packages/stream_chat/lib/src/core/error/stream_chat_error.dart
xsahil03x: Yes — deleted in 258d5a8d1.

I checked whether anything still needed it first: nothing in any `lib/` references it, in this package or in `stream_chat_flutter`, `stream_chat_flutter_core`, persistence or the sample app. The only references left were two stale test *names* in `channel_test.dart` and a string literal in `stream_chat_default_channel_error_test.dart` asserting the raw type name never reaches the user. `ErrorResponse` stays — it is still `StreamWebSocketError`'s payload.

Worth being explicit that this reverses a written decision rather than just filling one in. `core-migration/03-errors.md` deliberately deferred the deletion to `openapi-migration` group 12, reasoning that endpoints that group had not moved still flow through the hand-written verb facade. That reasoning does not survive this phase: the facade throws a `StreamChatException` now, so nothing raises the type whichever endpoint you call. The same file's risk section already called the deprecation "the one break a consumer can ship without noticing" — an `on StreamChatNetworkError catch` that compiles and matches nothing. Deleting makes it a compile error instead, which is what the major is for.

So the plan's own contract moved with it: the deferral entry in `03-errors.md`, the `DEFERRED.md` row, and the migration guide's call-out, which said the opposite of what now ships. `StreamChatNetworkErrorType` went too.

packages/stream_chat/lib/src/client/retry_queue.dart
VelikovPetar: Fine with me, thanks for the explanation!

VelikovPetar (approved):

PR #2958, opened 2026-09-12
refactor(llc)!: adopt stream_core's Filter for every query

xsahil03x:
Linear: FLU-785

## CLA

### 🎯 Goal

`Filter` was one class with a nullable `operator` string and a `key`. Core models it as a sealed type over a per-model field registry — the same registry that drives sorting — so one declaration serves both the server query and client-side `Filter.matches`.

### 🛠 Implementation details

- `Filter.equal('type', 'messaging')` becomes `ChannelFilter.equal(ChannelFilterField.type, 'messaging')`. A field the SDK does not model is reached with `ChannelFilterField.custom('my_field')`.
- `Filter` is sealed and no longer exposes `key`, `value` or `operator`; read it with `toJson` or pattern match. It also compares by identity now, so compare `toJson()` where you compared filters.
- `Filter.notEqual`, `notIn` and `nor` are removed — the API is withdrawing `$ne`, `$nin` and `$nor`. A caller that excluded someone now excludes them client-side, or shows them. The sample app shows them: its directory and member picker no longer hide the signed-in user or existing members.
- `Filter.empty()` is removed; every `filter` argument is nullable. This matters on `queryThreads`, where the API widens an omitted filter but not an empty one.
- `Filter.custom` is removed in favour of a registry's `custom` factory; `Filter.raw` takes its map positionally and is not validated.
- `stream_core` and `stream_core_flutter` now resolve from a pinned commit rather than a hosted constraint, because this PR needs `Filter.raw`, which is not in published 0.5.0. Pinned to a SHA rather than a branch: `stream_core_flutter` reaches `stream_core` through the same clone, and pub counts a branch ref and a resolved SHA as two sources of one package, which fails to solve. `invalid_dependency` is silenced for the same reason — a git dependency makes these packages temporarily unpublishable, which is true and intended.

### ☑️ Verification

`melos run analyze`, `melos run format` and `melos run test:dart` pass, verified in a clean worktree with no sibling checkout in scope — the same conditions CI runs under.

Filter fields were verified against the backend's own query layer, not the published spec — the spec still advertises operators the API rejects.

Restoring hosted constraints is a v11 release task, tracked in `core-migration/DEFERRED.md`.

## Summary by CodeRabbit

* **Breaking Changes**
  * Query filters and sorting now use type-specific APIs and fields across channels, users, members, messages, polls, reactions, reminders, and threads.
  * Legacy generic filter constructors and string-based field definitions are no longer supported.
  * List controllers and persistence APIs now require their corresponding typed filters and sort lists.

* **New Features**
  * Added channel `muted` and `blocked` state accessors, including change streams.
  * Added typed filtering and sorting support for additional query types.

* **Documentation**
  * Expanded migration guidance, changelogs, and filter registry documentation.

* **Bug Fixes**
  * Improved handling of invalid channel metadata values and nullable persisted filters.


packages/stream_chat_flutter/pubspec.yaml
```diff
@@ -55,7 +55,11 @@ dependencies:
   share_plus: ">=12.0.2 <14.0.0"
   shimmer: ^3.0.0
   stream_chat_flutter_core: ^10.4.0
-  stream_core_flutter: ^0.5.1
+  stream_core_flutter:
+    git:
+      url: https://github.com/GetStream/stream-core-flutter.git
+      ref: ad963968f87797262bd5d232e5670c0f0a8e885a
+      path: packages/stream_core_flutter
```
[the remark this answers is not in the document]
xsahil03x: This is already the case — the pin lives in `melos.yaml` (`command.bootstrap.dependencies`, lines 116 and 121), and what you are reading in `pubspec.yaml` is what `melos bootstrap` writes out of it. Editing the pubspec directly would be overwritten on the next bootstrap.

Worth flagging why it is a SHA and not a branch, since that part is not obvious: `stream_core_flutter` reaches `stream_core` through the same clone, and pub compares git *descriptors* rather than resolved commits — a branch ref on one side and a resolved SHA on the other count as two sources of one package and fail to solve. Both entries naming the same commit is what makes it resolve.

The hosted constraint is restored before the v11 release, together with the `invalid_dependency` suppression in `analysis_options.yaml`; they are one change in two files.

sample_app/lib/widgets/add_members_sheet.dart
```diff
@@ -80,22 +79,11 @@ class _AddMembersSheetState extends State<AddMembersSheet> {
...
-      if (query.isNotEmpty)
-        Filter.or([
-          Filter.autoComplete('name', query),
-          Filter.autoComplete('id', query),
-        ]),
+  UserFilter? _filter({String query = ''}) {
+    if (query.isEmpty) return null;
+    return UserFilter.or([
+      UserFilter.autoComplete(UserFilterField.name, query),
+      UserFilter.autoComplete(UserFilterField.id, query),
```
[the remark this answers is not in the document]
xsahil03x: Deliberate, and the changelog wording that suggested otherwise is fixed in `959bc8dec`.

Dropping `$ne`/`$nin` genuinely removes the exclusion — the server no longer supports expressing it — so the options were to filter client-side or to show everyone. The sample app shows everyone, and the changelog now says that plainly instead of implying the app compensates: "A query that excluded someone … has to do that in the client now, or show them. The sample app shows them."

Filtering the picker client-side is a reasonable follow-up for the sample app, but it is a sample-app UX change rather than part of the filter migration, so I have left it out of this PR.

core-migration/08-query-dsl.md
```diff
@@ -132,148 +132,232 @@ is not a target to match.
...
+An earlier pass extracted these from `mq.TableConfig.Columns` per resource and was wrong
+everywhere: 16 message fields against the spec's 44, 23 channel fields against 29, and
+`QueryMessageFlagsPayload` missed entirely. Reading one `Columns` map misses fields a second
+config contributes, and it does not apply the publication rules. Two properties make the spec the
+right source instead — it is keyed by *endpoint* rather than by table, which is how a caller
+thinks, and it has already applied `isPublishedOperator`, so
+[#15657](https://github.com/GetStream/chat/pull/15657) is baked in: exactly four `$ne` and one
+`$nin` survive, the index-safe exceptions on `QueryUsersPayload`.
+
+Regenerate with `git -C ~/GolandProjects/protocol pull` and re-reading that file; do not
```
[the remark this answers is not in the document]
xsahil03x: Fixed on this rung. The instruction no longer hard-codes one workstation's path — it names the `GetStream/protocol` checkout and takes its location from `$PROTOCOL_DIR`, with the old path as the default.

packages/stream_chat/lib/src/core/models/channel_model.dart
```diff
@@ -244,6 +258,8 @@ class ChannelModel {
     cooldown: cooldown ?? this.cooldown,
     disabled: disabled ?? extraData?['disabled'] as bool? ?? this.disabled,
     hidden: hidden ?? extraData?['hidden'] as bool? ?? this.hidden,
+    muted: muted ?? extraData?['muted'] as bool? ?? this.muted,
+    blocked: blocked ?? extraData?['blocked'] as bool? ?? this.blocked,
```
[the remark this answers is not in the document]
xsahil03x: Fixed in 05cd2e0, extended to `disabled` and `hidden`, which had the same cast. A test pins it: `copyWith(extraData: {'muted': 'yes', 'blocked': 1})` now falls back to the current value instead of throwing, matching what the getters already did.

packages/stream_chat/lib/src/core/models/predefined_filter.dart
```diff
@@ -27,10 +26,11 @@ class PredefinedFilter {
 
   /// Filter conditions as resolved by the server.
   ///
-  /// Wrapped in [Filter.raw] — the SDK does not evaluate filters locally.
-  /// Access the underlying map via [Filter.value] or [Filter.toJson].
+  /// Wrapped in [Filter.raw], since the server authors it and may use an
+  /// operator this package does not model. Read it with [Filter.toJson];
+  /// [Filter.matches] throws for it.
```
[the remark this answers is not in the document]
xsahil03x: Fixed in 05cd2e0. `predefined_filter.dart` imports `channel_state.dart` and not `stream_core`, so `Filter` is not in scope and those three links resolved to nothing. They now name `ChannelFilter`, which is the typedef the file actually uses.

packages/stream_chat/lib/src/core/models/channel_model.dart
```diff
@@ -46,12 +48,14 @@ class ChannelModel {
...
 
-       // For backwards compatibility, set 'disabled', 'hidden'
-       // and 'truncated_at' in [extraData].
+       // For backwards compatibility, set 'disabled', 'hidden', 'muted',
+       // 'blocked' and 'truncated_at' in [extraData].
        extraData = {
          ...extraData,
          if (disabled != null) 'disabled': disabled,
          if (hidden != null) 'hidden': hidden,
+         if (muted != null) 'muted': muted,
```
VelikovPetar, line 57: Shouldn't we be safe to define these fields at root level, as we are targetting `v11`?

packages/stream_chat/lib/src/core/models/channel_model.dart
```diff
@@ -136,6 +140,14 @@ class ChannelModel {
   @JsonKey(includeToJson: false, includeFromJson: false)
   bool? get hidden => extraData['hidden'].safeCast<bool>();
 
+  /// True if the channel is muted by the current user
+  @JsonKey(includeToJson: false, includeFromJson: false)
+  bool? get muted => extraData['muted'].safeCast<bool>();
```
VelikovPetar, line 145: Hmm curious about the `muted` field - do we now receive this info in the `channel` payload? Until recently, I believe the only info about muted channels was received via the `connected` event in the `OwnUser.channel_mutes` - has this changed now?

packages/stream_chat/lib/src/core/models/channel_state.dart
```diff
@@ -128,13 +129,240 @@ class ChannelState {
...
+  /// **Supported operators:** `$eq`, `$in`, `$gt`, `$gte`, `$lt`, `$lte`, `$exists`
+  static final messageCount = ChannelFilterField(
+    'message_count',
+    (it) => it.channel?.messageCount,
+  );
+
+  /// Filters channels by the tags assigned to them.
+  ///
+  /// **Supported operators:** `$eq`, `$in`
+  static final filterTags = ChannelFilterField(
```
VelikovPetar: I am not sure we should be publicly exposing this - as far as I know, the `filter_tags` feature has already been deprecated.

packages/stream_chat/lib/src/core/models/message.dart
```diff
@@ -726,12 +726,145 @@ class Message extends Equatable {
...
+/// Represents a field that a message search can be filtered on.
+class MessageSearchFilterField extends FilterField<Message> {
+  /// Creates a message filter field named [remote] on the wire, reading its
+  /// value off an instance with [value].
+  MessageSearchFilterField(super.remote, super.value);
+
+  /// Creates a field the SDK does not model, read from [Message.extraData].
+  ///
+  /// **Supported operators:** `$eq`, `$in`, `$gt`, `$gte`, `$lt`, `$lte`
+  factory MessageSearchFilterField.custom(String remote) {
```
VelikovPetar, line 750: Question about the `.custom` filters (for other types of filters as well): Should we maybe also allow passing the `local` lambda? We can still default to `extraData[remote]`, but I am thinking whether it is possible that the default is not enough, for example if we have some nested custom data:
- `remote: 'config.priority` => `local: (it) => it.extraData['config']['priority']`

packages/stream_chat/CHANGELOG.md
```diff
@@ -20,6 +20,17 @@
 - `PollVoteSortField.answerText` is removed: the API rejects a sort on `answer_text`.
 - `ChannelSortField.cid` is added, matching the iOS and Android SDKs.
 - `search(sort:)` on the client and channel, and `StreamMessageSearchListController.sort`, are typed `List<MessageSearchSort>` rather than an untyped `SortOrder`. Searching is the only message query the API sorts, so the type is named for it.
+- Filtering is now `stream_core`'s. `Filter.equal('type', 'messaging')` becomes `ChannelFilter.equal(ChannelFilterField.type, 'messaging')` — one `Filter` alias and one field registry per query, matching the sort change. A field the SDK does not model is reached with `ChannelFilterField.custom('my_field')`.
```
VelikovPetar: Similar to what I raised in a different PR, I am a bit worried that referencing `stream_core` might be confusing for integrators, as they would never directly include the package.

packages/stream_chat/lib/src/core/models/channel_model.dart
xsahil03x: Yes, this changed — the channel payload carries it now.

`ChannelResponse` declares both fields, and the spec documents them:

```go
Muted   *bool `json:"muted,omitempty"`    // "Whether this channel is muted or not"
Blocked *bool `json:"blocked,omitempty"`  // "Whether this channel is blocked by current user or not"
```

`OwnUser.channel_mutes` still arrives on the `connected` event, so both routes exist; this is the per-channel one. Both are `*bool` with `omitempty`, which is why the Dart side is `bool?` rather than `bool` — a payload that omits them reads `null` rather than `false`.

Worth separating from the *filter*, which is a different mechanism: `muted` as a filter is a server-computed predicate (`exists (select 1 from mutes_channels where ... user_id = ?)`), resolved against the calling user rather than read off a channel column. Same for `blocked`. So the filter works regardless of whether the response happened to include the field.

packages/stream_chat/lib/src/core/models/message.dart
xsahil03x: It already is possible — `.custom` is the convenience shorthand, not the only way in.

`ChannelFilterField(remote, value)` is a public constructor, so your nested case is:

```dart
ChannelFilterField('config.priority', (it) => it.channel?.extraData['config']?['priority'])
```

`.custom(remote)` exists only to save typing the common case, and its dartdoc says it reads `extraData[remote]`. Adding a `local` parameter to it would just be the plain constructor with a longer name, so I would rather leave the two as they are: `custom` for the flat default, the constructor when the default does not fit. The same holds for every registry — each declares the same pair.

packages/stream_chat/lib/src/core/models/channel_model.dart
xsahil03x: Agreed in principle, and your point is sharper than it looks — `muted` and `blocked` are *new here*, so this PR is adding fields in the backwards-compatible shape rather than inheriting them in it.

Not doing it in this PR though: it is a `ChannelModel` shape decision, and the same one applies to `disabled`, `hidden` and `truncatedAt` already sitting there, plus `User.name` / `image` which came up on #2957. Promoting them one at a time, from whichever PR happens to touch the next flag, is how a model ends up half-promoted.

Recorded in ffe298ab1 against `openapi-migration/11-channels-and-members.md`, the group that owns `ChannelModel`, so it gets answered for the whole model at once rather than rediscovered. All five fields do arrive as root fields on `ChannelResponse`, so nothing on the wire stands in the way.

packages/stream_chat/CHANGELOG.md
xsahil03x: Agreed, and fixed more widely than the entry you flagged — in ffe298ab1 and e35551082, plus the merge into the rungs above. There is no mention of `stream_core` left anywhere in the unreleased section of any package's changelog.

You are right that it explains a change in terms of something the integrator cannot see. The entries now say what moved without naming where it moved from: "Logging is rebuilt", "The token layer is retyped", "Sorting is restructured", "Filtering is restructured", "one of four sealed `StreamException` kinds", "`LocationCoordinates` is renamed `LocationCoordinate`", and so on. Where a type genuinely changed identity — `SystemEnvironment`, `Result` — the entry says that and what stays source-compatible, which is the part a reader needs.

packages/stream_chat/lib/src/core/models/channel_state.dart
xsahil03x: Checked the backend and the spec, and neither shows it as deprecated — so I have left it, but you may well know something the source does not carry.

What I found:

- `filter_tags` is a live column in the channel table config, with `$eq` and `$in` handlers and a dedicated GIN index (`idx_channels_..._gin`). No deprecation marker anywhere in `lib/core/mq/channel/channel.go`.
- The current spec still carries it both ways: `UpdateChannelRequest.add_filter_tags` / `remove_filter_tags`, and `filter_tags` on `ChannelResponse` — both present in the generated client in `lib/open_api/`.

If it is deprecated at the product level rather than in code, that is exactly the kind of thing the registry should not publish, and I will drop `ChannelFilterField.filterTags` — a caller can still reach it through `ChannelFilterField.custom('filter_tags')` if they are already using it. Say the word and I will.

core-migration/02-platform-and-environment.md
```diff
@@ -17,9 +17,9 @@
 | --- | --- | --- |
 | `lib/src/system_environment.dart` (57) | `stream_core` `api/system_environment.dart` | **done** |
 | `lib/src/core/http/system_environment_manager.dart` (111) | `stream_core` `api/system_environment_manager.dart` | **done** |
-| `lib/src/core/platform_detector/` — `platform_detector.dart` (~97) + `_stub` / `_web` / `_io` | `stream_core` `platform/current_platform.dart` + `platform/detector/` | **blocked on a `stream_core` release** |
+| `lib/src/core/platform_detector/` — `platform_detector.dart` (~97) + `_stub` / `_web` / `_io` | `stream_core` `platform/current_platform.dart` + `platform/detector/` | **ready** — the SHA pinned in [08](08-query-dsl.md) carries what this needed |
```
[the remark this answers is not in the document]
xsahil03x: Right, and the note was self-defeating — it exists to stop someone diffing against the wrong copy of core, and it named the wrong copy. It cost me an hour of phantom `Filter.raw` errors on this branch before I remembered to bootstrap.

Corrected: the resolved path is read out of `packages/stream_chat/.dart_tool/package_config.json` rather than assumed, with a pointer that `melos.yaml` pins a git `ref` today so the copy in use is under `~/.pub-cache/git/`, and that it returns to a hosted constraint when core next releases. The release statements further down were already right and are untouched.

packages/stream_chat/lib/src/core/models/predefined_filter.dart
```diff
@@ -40,29 +40,19 @@ class PredefinedFilter {
...
-}
-
-bool _touchesField(Filter filter, String field) {
-  if (filter.key == field) return true;
-  final value = filter.value;
-  if (value is List<Filter>) {
-    return value.any((sub) => _touchesField(sub, field));
+List<ChannelSort> _defaultSortFor(ChannelFilter filter) {
+  final lastMessageAt = ChannelSortField.lastMessageAt;
+  if (_mapTouchesField(filter.toJson(), lastMessageAt.remote)) {
```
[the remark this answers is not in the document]
xsahil03x: Correct, and already tracked — you have independently found what the sort phase recorded.

`core-migration/DEFERRED.md` carries this with the exact condition to match: the server sets its flag only when *every* `last_message_at` node is reachable through `$and` alone **and** at least one narrows by value (`$eq` / `$ne` / `$gt` / `$gte` / `$lt` / `$lte`, or `$exists: true`), per `DetectLastMessageAtFiltering` in `lib/core/mq/channel/modifiers.go`.

It is deferred rather than fixed because closing it is a behaviour change, not a translation: a test currently pins the loose `$or` case to today's answer, so the fix inverts that test. Worth noting the mismatch is narrower than it first looks — when the server *does* set the flag, it rewrites a single sent `last_updated` to `last_message_at` itself, so the two agree there. The gap is only where our check fires and the server's does not.

packages/stream_chat/lib/src/core/models/channel_state.dart
xsahil03x: Checked the other SDKs as well, and the evidence says keep it — iOS curates the same filter key by hand.

| | Model field | Curated filter key | Deprecated? |
| --- | --- | --- | --- |
| **Swift** | `Channel.filterTags: Set<String>` | **Yes** — `FilterKey.filterTags` in `ChannelListQuery.swift`: `rawValue: "filter_tags"`, `isCollectionFilter: true`, documented *"Supported operators: `in`, `equal`"* | no |
| **Android** | `Channel.filterTags: List<String>` | n/a — `Filters` is generic key-based, so there is nothing to curate | no |
| **JS** | `ChannelResponse.filter_tags`, `ChannelData.filter_tags` | structural only: `ChannelFilters` maps over `keyof ChannelResponse`, so it is filterable by accident rather than by choice | no |
| **Backend** | column with a GIN index, `$eq` / `$in` handlers | — | no marker in `lib/core/mq/channel/channel.go` |
| **Spec** | `ChannelResponse.filter_tags`, `UpdateChannelRequest.add_filter_tags` / `remove_filter_tags` | — | — |

Swift is the closest comparison, because it is the only other SDK with a hand-written allowlist of filter keys rather than a structural type — and it declares this one deliberately, with the same two operators the backend accepts, which is what our declaration documents too.

Nothing in any SDK, the backend, or the spec marks it deprecated. So I have left `ChannelFilterField.filterTags` in. If it is deprecated at a product level that none of the code reflects, that is worth recording somewhere the next person will find it, and I will drop the field — `ChannelFilterField.custom('filter_tags')` would still reach it for anyone already using it.

xsahil03x: You were right, and my cross-SDK check was looking in the wrong place: the deprecation is customer guidance, not an API change. New apps are pointed at custom-data filtering or predefined filters; the field keeps working for everyone already on it.

Dropped `ChannelFilterField.filterTags` in 7e831ec99. A typed, curated field reads as a recommendation, and this is a path we tell customers not to take. What stays:

- `ChannelFilterField.custom('filter_tags')` still reaches it, so nobody already depending on it is stuck.
- `ChannelModel.filterTags` is untouched — reading a value the server sends is not the same as recommending a query, and all three other SDKs expose it on the model.

The reason is written into `core-migration/08-query-dsl.md` rather than left as a code comment, because every source this registry is derived from argues for *keeping* it: live column with a GIN index and `$eq` / `$in` handlers, in the spec both ways, and a hand-written `FilterKey` in Swift. Anyone re-deriving the registry will add it straight back — this phase did, which is how you caught it.

packages/stream_chat/lib/src/core/models/channel_model.dart
VelikovPetar: OK, sounds good to me!

packages/stream_chat/lib/src/core/models/channel_model.dart
VelikovPetar: Maybe something that we should keep track of as a follow up - updating the `muted` channel state on the relevant events.

packages/stream_chat/lib/src/core/models/message.dart
VelikovPetar: 🙈 Completely missed that...

VelikovPetar (approved):

PR #2959, opened 2026-09-12
refactor(llc)!: adopt stream_core's LocationCoordinate, sort normalizer and list extensions

xsahil03x:
Linear: FLU-786

## CLA

### 🎯 Goal

Three more forks retired: the coordinate model, the string sort normalizer, and the list extensions that `ChannelClientState` runs on every message, member and read update.

### 🛠 Implementation details

- `LocationCoordinates` becomes core's `LocationCoordinate` — singular, re-exported from this package. `copyWith` is removed; it gains `distanceTo`. Two coordinates now compare equal when they agree to about a centimetre rather than requiring identical doubles.
- `normalizeStringForSort` moves to core, dropping `diacritic` as a direct dependency here.
- The list extensions become core's `SortedListExtensions`. `mergeSorted` becomes `sortedMerge` and now collapses a repeated key instead of carrying it through; `updateIf` becomes `updateWhere`; `mergeFrom` is dropped in favour of `merge` over a projected list.

### ☑️ Verification

`melos run analyze`, `melos run format` and `melos run test:dart` pass against the pinned `stream_core` commit this stack resolves from.

The list extensions are the SDK's hottest path, so the two methods were benchmarked before adoption rather than after: parity with the chat code they replace is 0.97–1.01x, and a disjoint-key fast path makes pagination ~8% faster. Numbers in `core-migration/01-utilities.md`.


packages/stream_chat/lib/src/client/channel/channel_client_state.dart
```diff
@@ -1362,13 +1362,17 @@ class ChannelClientState {
...
       membership: updatedState.membership,
-      read: newReads.toList(),
+      read: newReads,
       draft: updatedState.draft,
-      pinnedMessages: updatedState.pinnedMessages,
+      // Sorted on the way in because every later write goes through
+      // `_mergePinnedMessagesIntoExisting`, which merges by `createdAt` and
+      // requires a receiver already in that order. The API returns these by
+      // `pinned_at` descending, so the payload order is not it.
+      pinnedMessages: updatedState.pinnedMessages?.sorted(_sortByCreatedAt),
```
VelikovPetar: Question: Is this a bug that we currently have? Wouldn't it be better if we keep the order of the pinned messages as they are returned from the server (ordered by `pinned_at`)?

xsahil03x: Yes, it is a pre-existing bug — this PR only makes it visible.

The server returns pinned messages by `pinned_at` (`ORDER BY message.pinned_at DESC NULLS LAST`, `monolith/state/messages/`), and that order was stored as-is. But every later write goes through `_mergePinnedMessagesIntoExisting` → `_mergeMessagesIntoExisting`, whose single-message fast path calls `sortedUpsertAt(..., compare: _sortByCreatedAt)` — a sorted insert that assumes the receiver is already in `createdAt` order. So on v10 the first pin after a query was already placed by the wrong key; it was just silent. On v11 core's `sortedUpsertAt` asserts its receiver is sorted, which turns the same inconsistency into a debug crash.

Sorting on the way in was the minimal fix that makes the existing merge sound.

**I agree `pinned_at` is the better order to keep** — newest-pinned-first is what a pinned list should show, and it is what the server already gives us. It is a bigger change than this PR though: every write path through `_mergePinnedMessagesIntoExisting` has to compare by `pinnedAt` instead, and `Message.pinnedAt` is `DateTime?` where `createdAt` is non-null, so it needs a null-ordering decision (and `_pinIsValid` already drops expired pins on the way out). Happy to do it as its own change if you want it in v11 — say the word and I will open it.

VelikovPetar: I would say to track and fix it separately, it is pretty unrelated to what this PR does!

VelikovPetar (approved):

packages/stream_chat/lib/src/client/channel/channel_client_state.dart
xsahil03x: Agreed — tracked in 7962bd6ff, as a row in `core-migration/DEFERRED.md` rather than folded in here.

The row records what you would need to know to pick it up: the API returns pinned messages `ORDER BY message.pinned_at DESC NULLS LAST`; every write goes through `_mergePinnedMessagesIntoExisting` → `sortedUpsertAt(compare: _sortByCreatedAt)`, so a pin lands in creation order; and ordering by `pinnedAt` properly means changing every path plus a null-ordering decision, since `Message.pinnedAt` is nullable where `createdAt` is not.

The sort-on-the-way-in stays here, since without it core's `sortedUpsertAt` assert turns the pre-existing mis-placement into a debug crash — it makes the existing merge sound without claiming to be the right order.

xsahil03x: Reverted in 5c4b28805 — the sort is out of this PR, along with the test that pinned it. `DEFERRED.md` carries the follow-up.

One thing worth knowing before that separate PR is picked up, because the revert leaves a live edge rather than restoring v10 behaviour exactly:

`_mergeMessagesIntoExisting` is shared by regular and pinned messages, and its fast path calls `existingList.sortedUpsertAt(..., compare: _sortByCreatedAt)`. This phase swaps chat's helper for core's, and core's opens with `assert(_debugAssertSorted(this, compare))`. So a pinned payload in `pinned_at` order — which is what the API returns — now trips that assert in **debug** the first time a pin arrives after a channel query. Release builds strip the assert and misplace the message, which is exactly what v10 did silently.

So the follow-up is two halves of one bug rather than a pure ordering improvement, and the `DEFERRED.md` row now says so: compare on `pinnedAt` throughout, which both fixes the order and removes the mismatch the assert is catching. It needs a null-ordering decision, since `Message.pinnedAt` is nullable where `createdAt` is not.

Happy to open that PR next if you want it before the betas — say the word.

PR #2960, opened 2026-09-12
refactor(llc, ui)!: adopt stream_core's platform detector and sealed exception family

xsahil03x:
Linear: FLU-787

## CLA

### 🎯 Goal

Closes the migration: the platform detector goes, and the old error tree is deleted outright rather than left deprecated.

### 🛠 Implementation details

- `CurrentPlatform` and `PlatformType` become core's, re-exported from this package. `CurrentPlatform.name` becomes `.operatingSystem`, reporting the same string. Deleting our copy also drops the `js_interop` branch chat needed only because its own stub threw.
- `StreamChatError`, `StreamChatNetworkError`, `StreamChatNetworkErrorType` and `StreamWebSocketError` are **removed, not deprecated**. Nothing throws them any more, so a deprecated type would leave `on StreamChatNetworkError catch` compiling while silently matching nothing — a break you ship without noticing. Removing it is a compile error instead.
- `isRetriable` becomes an extension on `StreamChatException`, so it reads the same on any failure.
- `stream_chat_flutter`'s three attachment-validation errors move to their own sealed `AttachmentValidationError` family. They are returned, never thrown, so typing a refused attachment as a transport failure invited callers to treat it as one.
- Adds a test pinning the interceptor pipeline order. The existing tests only asserted each interceptor was present, which a fully reordered pipeline satisfies too.

### ☑️ Verification

`melos run analyze`, `melos run format` and `melos run test:dart` pass against the pinned `stream_core` commit this stack resolves from.

The pipeline-order test was checked to fail on a swap rather than only to pass as written — moving `ApiErrorInterceptor` after the logging interceptor makes it red.


packages/stream_chat/lib/src/client/channel/channel.dart
```diff
@@ -608,13 +608,12 @@ class Channel {
   }) {
     final cancelToken = _cancelableAttachmentUploadRequest[attachmentId];
     if (cancelToken == null) {
-      throw const StreamChatError(
-        "Upload request for this Attachment hasn't started yet or maybe "
-        'Already completed',
+      throw StateError(
```
VelikovPetar, line 611: I am curious why is this no longer a `Stream-` scoped error? 

Is it because it indicates an invalid use of the API, and not really a client-level error?

**Edit:** _I saw the updated changelog - it answers my question! We would probably need the sweep the other API surface for similar stuff._

VelikovPetar (approved):

packages/stream_chat/lib/src/client/channel/channel.dart
xsahil03x: Right on both counts — `StateError` because it reports a caller mistake (cancelling an upload that never started, or was already cancelled), not something the server said.

On the sweep: it is already done across the stack. There is no `throw StreamChatError` left anywhere in `packages/*/lib`, and the split now reads:

| package | `StateError` | `ArgumentError` | `StreamClientException` |
| --- | --- | --- | --- |
| `stream_chat` | 15 | 11 | 3 |
| `stream_chat_flutter_core` | 1 | 0 | 0 |
| `stream_chat_flutter` | 2 | 4 | 0 |
| `stream_chat_persistence` | 1 | 2 | 0 |

The rule applied was: `ArgumentError` for a bad argument, `StateError` for a call made at the wrong time, and `StreamClientException` only where the SDK itself failed at something it promised to do — the three in `stream_chat` are all the attachment-upload aggregate (`_uploadFailureMessage`), which is a real SDK failure rather than misuse. `ERROR_LAYER.md` is the source for that boundary.

VelikovPetar: Awesome, thanks for confirming - makes sense!

PR #2970, opened 2026-09-17
refactor(llc)!: move the WebSocket onto stream_core's StreamWebSocketClient

xsahil03x:
Replaces `stream_chat`'s hand-rolled WebSocket with `StreamWebSocketClient` from `stream_core`, so connection handling, health checks and reconnection are one implementation shared with Feeds and Video rather than three.

Part of [FLU-792](https://linear.app/stream/issue/FLU-792/migrate-the-websocket-onto-stream-core). Depends on GetStream/stream-core-flutter#186.

## What changed

**`websocket.dart`, `timer_helper.dart` and `connection_id_manager.dart` are gone** — roughly 1,900 lines of socket, ping/pong, backoff and connection-id bookkeeping now come from core.

**`ConnectionManager` replaces them**, owning the socket, the recovery handler and the session a connection belongs to. `ConnectRequest` builds the per-attempt URL; because chat authenticates through the query string rather than a frame, the token is loaded inside the options builder, which is what core#186's asynchronous builder is for. A token the server refuses is dropped there and reloaded on the next attempt, and a static provider with nothing else to give ends the session rather than presenting the same token forever.

**`ConnectionStatus` stays, and is what the client reports.** `wsConnectionStatus` and `wsConnectionStatusStream` become `connectionStatus` and `connectionStatusStream`, carrying the same three values. `connecting` covers a connection being opened *and* one the socket is waiting to reopen, which is what the old socket reported and what Swift reports; `disconnected` means nothing is coming. That distinction is not in the socket state alone — the same `Disconnected` describes a connection waiting out a backoff and one nothing will reopen — so the recovery handler publishes whether it has taken the connection on. The state and its disconnection sources stay inside the SDK, the same split iOS and Android make. `StreamConnectionStatusBuilder`, `ChannelListHeaderTitleBuilder` and the header widgets keep their pre-refactor shape.

**Asking for a connection that already exists is answered rather than refused.** Reopening is asked for on events that can land mid-attempt, so `connect` completes with the connection already open or being opened for that user, and two callers signing the same user in at once share one connection. Connecting a *different* user is refused while one is signed in.

**Health checks and connection ids** come from the socket's own frames: `connectionId` is read off the `Connected` state rather than tracked separately, and `ConnectionIdInterceptor` takes a getter instead of a manager.

**`Event` and its resolvers move under `ws/events/`**, next to the WebSocket event subclasses that extend `Event`. The resolvers keep running where every event enters the client rather than where the socket delivers one, so an event replayed over `/sync` after an outage is classified the same as one that arrived live.

## Breaking

`wsConnectionStatus` and `wsConnectionStatusStream` become `connectionStatus` and `connectionStatusStream`. The socket's state — `Connected`, `Connecting`, `Authenticating`, `Disconnected` and the disconnection sources — is no longer exported. `StreamChatClient.connect`'s return and the `ws` export surface change with it.

`connectUser` for a different user now throws while one is signed in, where an app with no persistence configured switched silently and kept the previous user's state; call `disconnectUser` first. Concurrent `connectUser` calls for the same user are answered with one connection instead of the second being refused.

An attempt that never connected is not retried. The socket retries a connection that dropped after being established; a first attempt that fails is reported to whoever asked for it, and making another is theirs.

## Verified

`dart analyze --fatal-infos` clean across every package. 1,831 tests pass in `stream_chat`, 373 in `stream_chat_flutter_core`, and the `stream_chat_flutter` suite is unchanged; the only failures are macOS-variant goldens, which use platform goldens locally and CI goldens in CI. New coverage lands in `connection_manager_test.dart`, `connect_request_test.dart`, `connection_status_test.dart`, `stream_chat_ws_event_test.dart` and a fake chat server that drives the socket without a network.


packages/stream_chat/lib/src/client/client.dart
```diff
@@ -672,11 +663,6 @@ class StreamChatClient {
     PaginationParams paginationParams = const PaginationParams(),
     bool waitForConnect = true,
   }) async* {
-    if (!_connectionIdManager.hasConnectionId) {
```
VelikovPetar: Just curious this changed now?

packages/stream_chat/lib/src/ws/websocket.dart
```diff
@@ -1,586 +0,0 @@
...
-    bool refreshToken = false,
-    bool includeUserDetails = true,
-  }) async {
-    final userDetails = ConnectUserDetails.fromOwnUser(_user!);
-    if (refreshToken) tokenManager.expireToken();
-    final token = await tokenManager.getToken();
-    final params = {
-      'user_id': userDetails.id,
-      'user_details': includeUserDetails ? userDetails : {'id': userDetails.id},
-      'user_token': token.rawValue,
```
VelikovPetar, line 189: I believe the `user_token` is now dropped from the `params` - just checking if it was intentional?

packages/stream_chat/lib/src/ws/connection_manager.dart
```diff
@@ -0,0 +1,218 @@
...
+  /// Reports under [tag], and the socket and recovery handler it owns under `<tag>:Ws` and
+  /// `<tag>:Recovery`, so one prefix selects the whole family.
+  ConnectionManager({
+    required this._request,
+    required TokenManager tokenManager,
+    WebSocketProvider? wsProvider,
+    String tag = 'SCh:Connection',
+  }) : _tokens = tokenManager,
+       _logger = StreamLogger(tag) {
+    _ws = StreamWebSocketClient(
```
VelikovPetar, line 28: This now takes the default `pingInterval`/`pongInterval` values from `core`, which are different from what we use now. Is this a fine change to make?

packages/stream_chat/lib/src/ws/websocket.dart
```diff
@@ -1,586 +0,0 @@
...
-    required this.apiKey,
-    required this.baseUrl,
-    required this.tokenManager,
-    this.systemEnvironmentManager,
-    this.handler,
-    this.webSocketChannelProvider,
-    this.reconnectionMonitorInterval = 10,
-    this.healthCheckInterval = 20,
-    this.reconnectionMonitorTimeout = 40,
-    this.maxReconnectAttempts = 6,
```
VelikovPetar, line 53: I think we also lost this logic (`maxReconnectAttempts`) - are we OK with that? (this is more of a `core` question, but it just came to my mind.

packages/stream_chat/lib/src/ws/connection_status.dart
```diff
@@ -1,11 +1,40 @@
...
+  ///
+  /// A connection that has yet to be opened reads as [disconnected], the same as one that closed:
+  /// neither carries events.
+  @internal
+  static ConnectionStatus fromState(
+    WebSocketConnectionState state,
+  ) => switch (state) {
+    Connected() => connected,
+    Connecting() || Authenticating() => connecting,
+    Initialized() || Disconnecting() || Disconnected() => disconnected,
```
VelikovPetar, line 30: > Detected by Claude

This mapping inverts what the header shows during a sustained outage.

Core sits in `Disconnected` for the whole backoff gap between retry attempts — `_scheduleReconnection` just arms a timer, nothing changes the state — so that's what this line reports for most of an outage, with brief flickers of `connecting` while an attempt is actually in flight. `channel_header.dart:198` renders those as "Disconnected" and "Reconnecting" respectively.

The old socket set `connecting` *before* the delay and held it across the whole run, so the header read "Reconnecting" steadily and only said "Disconnected" once it gave up.

Worth knowing the first retry is free: `getNextRetryDelay` returns `Duration.zero` while `consecutiveFailuresCount <= 0`, so a single blip is one event-loop turn and nobody sees it. It's from the second failure on that the gaps become whole seconds and grow toward 25 — that's where the user spends most of the outage looking at "Disconnected" instead of "Reconnecting".

No test covers the status across a retry sequence, so nothing would catch this either way.

packages/stream_chat/lib/src/ws/websocket.dart
xsahil03x: Intentional. The server's connect payload is three fields, and `user_token` is not among them — `lib/chat/controller/v1/connect.go`:

```go
type ConnectRequest struct {
	UserDetails         *payload.ConnectUserDetails `json:"user_details" validate:"required,dive"`
	UserDevice          *types.DeviceFields         `json:"device" validate:"omitempty,dive"`
	MemberCustomInclude []string                    `json:"member_custom_include" validate:"omitempty,max=8,dive,max=64"`
}
```

Authentication comes from the `authorization` query parameter, which we still send. Dropping the copy in `json` also means the token appears once in the URL rather than twice, so nothing has to redact a second occurrence.

Two related things that struct shows, both pre-existing: we also send `user_id` and `server_determines_connection_id`, which it does not parse either, and we never send `device`, which it does — the server registers a push device from it. That one is tracked as a follow-up on FLU-792.

packages/stream_chat/lib/src/ws/connection_manager.dart
xsahil03x: Yes, and the change is in our favour. Core's defaults are `defaultPingInterval = 25s` and `defaultPongTimeout = 3s` (`web_socket_health_monitor.dart:59-62`), against the old `healthCheckInterval: 20` with `reconnectionMonitorInterval: 10` / `reconnectionMonitorTimeout: 40`.

So we ping slightly less often and notice a dead socket in ~28s instead of ~40s.

If we would rather keep the old cadence, it is two arguments rather than a core change: `StreamWebSocketClient` takes `pingInterval` and `pongTimeout`, added for exactly this in GetStream/stream-core-flutter#186. Happy to pin them if you prefer parity with what shipped.

packages/stream_chat/lib/src/ws/websocket.dart
xsahil03x: Deliberate, and core has no cap to lose: I grepped `Sources/` in stream-chat-swift and there is no attempt limit there either — reconnection runs until it succeeds or the source says it cannot.

The old cap of 6 meant a socket that exhausted its attempts stayed dead until something external kicked it, which on mobile is worse than continuing to poll at the 25s ceiling: a user who walks back into coverage after a lift ride would have had to background and foreground the app to get their messages.

What replaces the cap is `DisconnectionSource.isReconnectable` — an invalid API key, an invalid token signature or a non-rate-limit 4xx now stops the loop, where the old socket retried everything except token expiry.

packages/stream_chat/lib/src/ws/connection_status.dart
xsahil03x: Confirmed and fixed in b082d67a3.

Probing the real socket showed it is worse than a steady "Disconnected" — each failed attempt also passes through `Disconnecting`, which mapped to the same value, so the sequence across a retry run was:

```
Connected → Disconnected(ServerInitiated) → Connecting → Disconnecting(ServerInitiated) → Disconnected(ServerInitiated) → …
```

i.e. the header flickered Disconnected → Reconnecting → Disconnected for every attempt, not just during the gaps.

`fromState` now reads the source rather than the state alone:

```dart
Disconnecting(:final source) || Disconnected(:final source) when source.isReconnectable => connecting,
```

so anything the socket intends to reopen reads as `connecting`, and `disconnected` comes to mean nothing is coming — which is also what `maybeReconnect`'s guard keys on, so the two agree.

Worth recording that the SDKs differ here, and I went with Swift: `stream-core-swift` maps a `disconnected` whose `isAutomaticReconnectionEnabled` is true to `.connecting` and keeps `.disconnecting` as its own case. Android does the opposite — `DisconnectedEvent` sets `ConnectionState.Offline` immediately and `Connecting` only arrives when `HealthMonitor.launchDelayed` fires the next attempt, so it shows Offline through the gap the way we did. Swift matches what our own socket did before this stack, and this is a regression report against that, so I followed it.

Two tests cover it now: the mapping in `connection_status_test.dart`, and a run against the fake socket asserting the status stays `connecting` across several failed attempts while the attempt count grows.

packages/stream_chat/lib/src/ws/websocket.dart
VelikovPetar: Nice, thanks for confirming!

packages/stream_chat/lib/src/ws/connection_status.dart
VelikovPetar: Makes sense, thanks for addressing it!

packages/stream_chat/lib/src/ws/connection_manager.dart
VelikovPetar: My only concern is that `defaultPongTimeout = 3s` is slightly low, but maybe we can fine-tune it along the way.

VelikovPetar (approved):

PR #2972, opened 2026-09-17
test(repo): tighten throw assertions in tests

VelikovPetar:
# Submit a pull request
Linear: FLU-794

## CLA

`packages/stream_chat/test/` asserted thrown errors with a bare `try`/`catch` and no `fail()` on the fall-through path:

```dart
test('should throw if trying to set `name`', () {
  try {
    channel.name = 'New name';
  } catch (e) {
    expect(e, isA<StateError>());   // never runs if nothing throws
  }
});
```

When the production guard goes away the `catch` body never runs, so the test stays green. `grep -rn '\bfail(' packages/*/test/` returned **zero hits** repo-wide, so no instance of this pattern anywhere closed the hole.

This converts all **56** such blocks to the `throwsA` idiom the suite already uses ~29 times (e.g. `channel_test.dart:6153`), and adds a `TESTING.md` rule so the pattern does not come back.

### Scale

**58** `try`/`catch` blocks in `packages/stream_chat/test/`; **56 are vacuous with respect to the throw**. Two were already correct — they capture the error and assert outside the `try` — and are left alone.

| File | Blocks | Pre-fix, mutated | Post-fix, mutated |
| --- | --- | --- | --- |
| `channel_test.dart` | 25 | all PASS | all FAIL |
| `stream_http_client_test.dart` | 10 | all PASS | all FAIL |
| `client_test.dart` | 8 | all PASS | all FAIL |
| `auth_interceptor_test.dart` | 4 | all PASS | all FAIL |
| `general_api_test.dart` | 3 | all PASS | all FAIL |
| `token_manager_test.dart` | 2 | all PASS | all FAIL |
| `websocket_test.dart` | 2 | all PASS | all FAIL |
| `requests_test.dart`, `token_test.dart` | 2 | PASS | FAIL |

### How this was verified

Every block was mutation-tested in both directions: neutralise the throw source, run that single test **before** the fix (must pass — proving it was vacuous) and **after** (must fail — proving the assertion is live). 56/56 in each direction.

Tier-1 blocks were mutated by neutralising the production guard: the `channel.dart` setters, the `requests.dart` / `general_api.dart` / `token.dart` / `token_manager.dart` asserts, the `websocket.dart` re-entrancy guard, the `AuthInterceptor` reject paths, and `StreamHttpClient`'s `_parseError` wrapping. The rest were mutated at the production seam — replacing each `rethrow` in `channel.dart` with a normal return — which leaves the surrounding state-transition assertions intact and isolates the error contract exactly.

Two mutations had to be deepened to mean anything. Deleting the `general_api.dart` asserts leaves an unstubbed mock returning null, so the test fails with a `TypeError` — that is "a different throw appeared", not "the throw was removed"; a faithful *returns-normally* simulation shows all three blocks are vacuous. `token_manager_test:54` is the same story: a second, same-typed assert downstream in `loadToken` keeps it green for the wrong reason until both are removed.

Production code is untouched — every mutation was reverted, and `git diff HEAD -- packages/stream_chat/lib` is empty.

### Two real defects this surfaced

1. **`token_test.dart` had zero coverage of the guard it names.** `` `.fromRawValue` should throw if does not contain `user_id` `` passed `'bad-token-without-a-user-id'`, which is not a JWT at all — `jose` throws `ArgumentError` while parsing, and the actual missing-claim guard (the `assert` at `token.dart:41`, an `AssertionError`) was never reached. The assertion happened to match the wrong error. Split into two tests, one per path.
2. **Two synchronous-throw traps.** `WebSocket.connect` and both `pinMessage` overloads return `Future` but are not `async`, so they throw while `expectLater`'s argument is being evaluated and the matcher never sees the error. These use the closure form; the old `try`/`catch` hid the distinction entirely.

### Two deviations from the fix proposed in FLU-794

- **Plain `throwsA(isA<T>())` would have dropped assertions.** Several catch bodies assert more than the type (`networkError.code`, `err.message`, equality against `StreamChatNetworkError.fromDioException(error)`). Those are preserved with `.having(...)` / `allOf(...)`.
- **`auth_interceptor_test` keeps a `try`/`catch`.** Those four blocks unwrap a private dio type via `(e as dynamic).data`. They use capture-and-assert-outside instead — the form the `stream_chat_flutter` / `stream_chat_flutter_core` suites already use — which closes the hole without forcing a `dynamic` access into a matcher.

So six `try`/`catch` blocks remain in the package by design: the two that were already correct, plus those four.

### Scope

`repo` rather than `llc`: this is test-only, changes no package behavior, and therefore takes no `CHANGELOG.md` entry per the [changelog policy](https://github.com/GetStream/stream-chat-flutter/blob/master/STYLE_GUIDE.md#changelog-policy). The `semantic_changelog_update` job keys that requirement off the PR scope.

### Testing

- `packages/stream_chat`: **1922 tests green** (1921 before, +1 from the `token_test` split).
- `melos run analyze`: clean across all packages.
- `dart format`: clean on every changed file.

### Follow-up, deliberately not in this PR

`general_api.dart:44`, `channel.dart:1207`, `client.dart:2172` and `requests.dart:32` validate **caller input** inside `assert`s, which [`STYLE_GUIDE.md`](https://github.com/GetStream/stream-chat-flutter/blob/master/STYLE_GUIDE.md) explicitly forbids ("Do not use asserts to validate user input or network data (those must throw at runtime)"). Those "should throw" contracts therefore do not exist in release builds at all. Fixing it is a behavior change and needs its own ticket.

## Screenshots / Videos

Not applicable — test-only change.

## Summary by CodeRabbit

- **Tests**
  - Updated error-handling tests to use clearer synchronous and asynchronous exception matchers.
  - Expanded coverage for malformed tokens, connection failures, invalid arguments, and WebSocket errors.
  - Added detailed validation of error types, messages, codes, and other relevant fields.
- **Documentation**
  - Added guidance for asserting synchronous and asynchronous errors, including when to use closures, futures, and captured exceptions.


TESTING.md
```diff
@@ -86,6 +86,91 @@ are cases where multiple calls represent a single behavior (e.g. "when the WebSo
...
+
+`throwsA` covers both, but only when it can invoke the call itself. Passing a future
+directly only ever covers the second case: in the first, the call throws while Dart is
+still evaluating the argument, so `expectLater` never runs and the error escapes as a
+raw failure instead of a matcher message.
+
+You cannot tell the two apart from a signature — `Future<T> foo()` can throw
+synchronously, `Future<T> foo() async` never does. Anything before the first `await`,
+including an `assert`, runs eagerly. So wrap the call in a closure whenever it might
+throw synchronously, and whenever you are not sure:
```
xsahil03x: This paragraph argues against itself, and it's the one doing the actual teaching.

1. "You cannot tell the two apart from a signature" is immediately followed by a rule that *does* tell them apart by signature. The true statement is narrower: `async` is at the **declaration**, not in the type — a call site, an abstract method, or a doc page won't show it.
2. "Anything before the first `await` … runs eagerly" is true, but it lands where it reads as the explanation for synchronous throws. It describes an `async` body, which is exactly the case where that eager code's throw *is* still captured into the returned future. As written it undercuts the sentence before it.

```suggestion
The call site doesn't tell you which one you have: `async` is part of the
declaration, not the return type — `Future<T> foo()` can throw synchronously,
`Future<T> foo() async` never does. An `async` body's throws are captured into
the returned future even when they happen before the first `await`. So wrap the
call in a closure whenever it might throw synchronously, and whenever you are
not sure:
```

xsahil03x (commented):
Good work, and the methodology holds up under checking rather than just being claimed.

I independently verified the three synchronous-throw sites this PR singles out, and all three are right:

- `client.pinMessage` (`client.dart:2169-2178`) — non-`async`, the `assert` throws `ArgumentError` eagerly
- `channel.pinMessage` (`channel.dart:1204-1213`) — same shape
- `WebSocket.connect` (`websocket.dart:224-233`) — non-`async`, the re-entrancy guard throws before the `Completer` is created

And the counter-case too: `openConnection` (`client.dart:506`) and `channel.markUnread` (`channel.dart:1892`) *are* `async`, so passing their futures directly is correct. `dart:async` is already imported in `websocket_test.dart`, so `unawaited` resolves. The `token_test.dart` split is a real defect fix — `'bad-token-without-a-user-id'` never reached the `user_id` guard at `token.dart:41`, because `jose` rejected it as malformed first.

One change I'd want before merge, one doc fix inline, and two notes.

---

## `translations_test.dart` still has the pattern this PR bans

`packages/stream_chat_localizations/test/translations_test.dart:573-580` is the exact shape being eliminated here:

```dart
test('should throw if try to load locale which is not supported', () async {
  const locale = Locale('not-supported-locale');
  try {
    getStreamChatTranslation(locale);
  } catch (e) {
    expect(e, isA<AssertionError>());
  }
});
```

Scoping the conversion to `packages/stream_chat/test/` is defensible on its own. But this PR also adds a **repo-wide** rule to `TESTING.md`, and this is the one known violator left outside that scope. A rule that ships with a live counter-example in the tree is the one people cite when they want to skip it.

I swept every other suite to confirm it is the only one. The four remaining catches — `stream_chat_test.dart:42`, `stream_channel_test.dart:78`, `stream_chat_core_test.dart:52`, `stream_state_scope_test.dart:55` — are the builder capture-and-assert-outside form the new doc explicitly endorses, and the `error_alert_sheet_test.dart` catches are fixture code, not assertions.

`getStreamChatTranslation` (`stream_chat_localizations.dart:54-59`) is synchronous and its unsupported-locale guard is a real `assert`, so the contract exists and this is a direct conversion (the file is outside this diff, so I can't attach an applicable suggestion):

```dart
test('should throw if try to load locale which is not supported', () {
  const locale = Locale('not-supported-locale');
  expect(
    () => getStreamChatTranslation(locale),
    throwsA(isA<AssertionError>()),
  );
});
```

`async` can go too — nothing in the body awaits.

---

## Two notes, not blocking

**`stream_http_client_test.dart:128-148`.** The two logger tests use an unmocked `StreamHttpClient`, so `client.get('path')` is a real network call. Replacing `catch (_) {}` with `throwsA(isA<StreamChatNetworkError>())` is the right call and robust in practice — both a connection failure and the inevitable 401 wrap into `StreamChatNetworkError` via `_parseError`. Just flagging that these two tests are now coupled to the request failing, where before they were indifferent to it.

**`token_test.dart` now asserts `AssertionError` for the missing-`user_id` case.** Accurate today, but it pins a debug-only contract: with asserts off, `userId!` at `token.dart:46` throws a `TypeError` instead, so the test name overstates what release builds guarantee. Fine to land given the follow-up is named.

---

## PR body

Per the repo convention the body should be what changed / why / how verified. The **"Follow-up, deliberately not in this PR"** section should collapse to a one-line `Follow-up: #NNN` against a tracked issue — an author-acknowledged "these asserts validate caller input and STYLE_GUIDE forbids it" paragraph reads as in-scope to a reviewer and comes straight back as a finding. The "Two deviations from FLU-794" rationale belongs on the lines it defends; the `auth_interceptor_test` half of it is already carried by the inline comment there.

---

Style is clean otherwise: 120-col respected, `format` green, and `.having(...)` / `allOf(...)` preserve every field assertion the old catch bodies made rather than flattening to a bare type check. The three inline comments explaining the closure form each mark a genuine non-obvious constraint, which is what STYLE_GUIDE asks of a comment.

TESTING.md
VelikovPetar: Will address it!

VelikovPetar:
@xsahil03x, thanks for the review!

Fixing the tests for the `translations` in a follow-up commit!

On the topic: **stream_http_client_test.dart:128-148. The two logger tests use an unmocked StreamHttpClient, so client.get('path') is a real network call.**: How do you think we should address this? Maybe with `MockHttpClientAdapter`  , which we can setup per test scenario? 

On the topic: **token_test.dart now asserts AssertionError for the missing-user_id case.**: This should be re-evaluated in the follow-up, which I just created: https://linear.app/stream/issue/FLU-812/asserts-used-to-validate-public-api-input-dont-exist-in-release-builds It is a bit of an exploration task - as there many places that we need to double check. (listed as low-priority, as this is more of a cross-check against the style-guide, rather than a bug)

xsahil03x:
Both fixes in `b3f0ecf` look right — thanks. And FLU-812 is the right home for the assert question; agreed it's an exploration task rather than a bug.

On the logger tests: yes to the adapter, and it's cheaper than you'd think — **`StreamHttpClient` already takes an `httpClientAdapter`**, so no production change is needed:

https://github.com/GetStream/stream-chat-flutter/blob/master/packages/stream_chat/lib/src/core/http/stream_http_client.dart#L31

Worth saying why `MockDio` specifically can't serve here, since it's the reason these two tests reached for a real `Dio` in the first place: `MockDio` is a mocktail mock of `Dio`, so `get()` is stubbed wholesale and the interceptor chain never runs — `LoggingInterceptor` would never fire and `verify(() => logger.info(any()))` would fail.

But the adapter alone only fixes the network call, and the network isn't the weakest thing about these two tests. `verify(() => logger.info(any())).called(greaterThan(0))` is "something logged, don't care what" — and the `response → info` branch has **no coverage at all today**, in any form. A 200 logs request(info) + response(info); a 400 logs request(info) + error(severe). With `any()` / `greaterThan(0)` you can't tell which branch fired, so routing `response` to `severe` stays green either way.

So I'd suggest splitting this into the two things it actually is.

### 1. The mapping — what `StreamHttpClient` owns

The only thing this class contributes is the `logPrint` routing in its constructor: `request → info`, `response → info`, `error → severe`. `logPrint` is public on `LoggingInterceptor`, so that's directly testable, and it matches the three sibling tests already in `group('loggingInterceptor')`:

```dart
LoggingInterceptor loggingInterceptorOf(Logger logger) =>
    StreamHttpClient('api-key', logger: logger)
        .httpClient
        .interceptors
        .whereType<LoggingInterceptor>()
        .single;

test('should log requests through `logger.info`', () {
  final logger = MockLogger();
  loggingInterceptorOf(logger).logPrint(InterceptStep.request, 'message');
  verify(() => logger.info('message')).called(1);
});

test('should log responses through `logger.info`', () {
  final logger = MockLogger();
  loggingInterceptorOf(logger).logPrint(InterceptStep.response, 'message');
  verify(() => logger.info('message')).called(1);
});

test('should log errors through `logger.severe`', () {
  final logger = MockLogger();
  loggingInterceptorOf(logger).logPrint(InterceptStep.error, 'message');
  verify(() => logger.severe('message')).called(1);
});
```

Exact arguments, `.called(1)`, all three steps, no network. `.single` is safe — the neighbouring test already pins the count at 1. This is also the only form that covers the response branch, since it's the only one that can name a step.

### 2. The chain — that dio actually drives it

If you want to keep an end-to-end pair, the adapter is the right seam:

```dart
class FakeHttpClientAdapter extends Fake implements HttpClientAdapter {
  FakeHttpClientAdapter({this.statusCode = 200, this.body = '{}'});

  final int statusCode;
  final String body;

  @override
  Future<ResponseBody> fetch(
    RequestOptions options,
    Stream<Uint8List>? requestStream,
    Future<void>? cancelFuture,
  ) async => ResponseBody.fromString(
    body,
    statusCode,
    headers: {
      Headers.contentTypeHeader: [Headers.jsonContentType],
    },
  );

  @override
  void close({bool force = false}) {}
}
```

```dart
test('should log the request and response through `logger.info`', () async {
  final logger = MockLogger();
  final client = StreamHttpClient(
    'api-key',
    logger: logger,
    httpClientAdapter: FakeHttpClientAdapter(),
  );

  await client.get('path');

  verify(() => logger.info(any())).called(greaterThan(0));
  verifyNever(() => logger.severe(any()));
});

test('should log the error through `logger.severe`', () async {
  final logger = MockLogger();
  final client = StreamHttpClient(
    'api-key',
    logger: logger,
    httpClientAdapter: FakeHttpClientAdapter(statusCode: 400),
  );

  await expectLater(
    client.get('path'),
    throwsA(isA<StreamChatNetworkError>()),
  );

  verify(() => logger.severe(any())).called(greaterThan(0));
});
```

`verifyNever(() => logger.severe(any()))` is the load-bearing line in the first test — without it, `info` firing "more than zero times" is true under almost any wiring.

It also makes `throwsA(isA<StreamChatNetworkError>())` exact rather than incidental: 400 fails dio's default `validateStatus`, `LoggingInterceptor.onError` logs it and forwards via `super.onError`, and `_parseError` (`stream_http_client.dart:100-105`) converts it.

### One limit worth knowing

No end-to-end test here can separate `request → info` from `response → info` — both land on the same method, and the only distinguishing evidence is the formatted log text, which lives in a file marked `// coverage:ignore-file` (`logging_interceptor.dart:2`) whose formatting is free to change. That's why the two halves are complementary rather than redundant: (1) pins the mapping, (2) proves the chain is connected and the error channel is distinct.

I'd skip a loopback `HttpServer` variant. `baseUrl` is injectable via `StreamHttpClientOptions` so it's easy enough, but it only buys proof that dio's own IO adapter works, and costs a `dart:io` import that pins the file to the VM.

Entirely happy for this to be a separate PR if you'd rather keep this one to the `try`/`catch` conversion — it's a different concern, and the current state is already strictly better than `catch (_) {}`.

xsahil03x (approved):

